PR #1, opened 2020-04-26
Create infra code for runners

npalm:
Runners are now defined by a launch template. To launch a runner spot instance at least the desired `subnet` should be provided. After launching a paramater in SSM should be created with the following requirements

type: SecureString
name: `<environment>-<instance-id>`
value: config to append to the action runner config command. e.g. --token abc --url https://github --labes mylabes


npalm:
Manually tested as follow:
1. Create instance based on launch template and subnet to a private subnet as created in the example.
2. Create an ssm paramter based on envrionment name and instance ide with the config for the runner.
3. Trigger a build

main.tf
```diff
@@ -0,0 +1,44 @@
...
+}
+
+module "dsitrubtion_cache" {
+  source = "./modules/action-runner-binary-cache"
+
+  aws_region  = var.aws_region
+  environment = var.environment
+  tags        = var.tags
+
+  distribution_bucket_name = random_string.random.result
```
gertjanmaas, line 14: Don't we want to add something readable into this like `environment`?

modules/runners/README.md
```diff
@@ -0,0 +1,9 @@
+# Action runner module
+
+The module create resources to facilitate the `orchestrator labmda` to recreate action runners.
+
+- *launch template* : A launch template is creates that can crate an action runner, by default a spot instance is requested. For configuration parameters SSM is used.
```
gertjanmaas, line 5:
```suggestion
- *launch template* : A launch template is created that can create an action runner, by default a spot instance is requested. For configuration parameters SSM is used. 
```

modules/runners/README.md
```diff
@@ -0,0 +1,9 @@
+# Action runner module
+
+The module create resources to facilitate the `orchestrator labmda` to recreate action runners.
+
+- *launch template* : A launch template is creates that can crate an action runner, by default a spot instance is requested. For configuration parameters SSM is used. 
+- *security group* : Security groups attached to the action runner.
+- *s3 bucket* : To avoid the action runner distribution have to be download a cached version is expected in a S3 bucket.
```
gertjanmaas, line 7:
```suggestion
- *s3 bucket* : To avoid the action runner distribution to be downloaded from Github every time (which could be slow), a version is cached in a S3 bucket.
```

modules/runners/main.tf
```diff
@@ -0,0 +1,113 @@
...
+  }
+  tags = merge(
+    local.tags,
+    {
+      "Name" = format("%s", local.name_sg)
+    },
+  )
+}
+
+resource "aws_resourcegroups_group" "resourcegroups_group" {
```
gertjanmaas, line 95: How does this work? Nothing is tagged with `ResourceGroup` so how is this resource group filled with resources?

npalm: Work based on the environment name

npalm: moved one level up

PR #5, opened 2020-04-30
sync GitHub action distribution to S3

npalm:
To avoid we have to download the github action dist everytime on runner is created we cach the dist in S3.


modules/action-runner-binary-cache/lambdas/syncer/src/lambda.ts
```diff
@@ -0,0 +1,8 @@
+import { handle } from './syncer/handler';
+
+module.exports.handler = async (event: any, context: any, callback: any): Promise<any> => {
+  const statusCode = await handle();
```
gertjanmaas, line 4: StatusCode isn't necessary, this lambda isn't part of the API

modules/action-runner-binary-cache/lambdas/syncer/src/syncer/handler.ts
```diff
@@ -0,0 +1,95 @@
...
+
+interface ReleaseAsset {
+  name: string;
+  downloadUrl: string;
+}
+
+async function getLinuxReleaseAsset(): Promise<ReleaseAsset | undefined> {
+  const githubClient = new Octokit();
+  const linuxAssets = (
+    await githubClient.repos.getLatestRelease({
```
gertjanmaas, line 43: Why not use the API? https://developer.github.com/v3/actions/self-hosted-runners/#list-runner-applications-for-a-repository

modules/action-runner-binary-cache/variables.tf
```diff
@@ -19,3 +19,8 @@ variable "distribution_bucket_name" {
   type        = string
 }
 
+variable "lambda_schedule_expression" {
+  description = "Scheduler expression for action distributio syncer."
```
gertjanmaas, line 23:
```suggestion
  description = "Scheduler expression for action runner binary syncer."
```

modules/action-runner-binary-cache/lambdas/syncer/src/syncer/handler.ts
npalm: This API requires a repo based auth taken as far as I understand, tried to use a public API. We do not receive an app event so no option to use repo based token.

npalm:
@gertjanmaas solved review comments, also add tagging for resources and update default example. Did not change the used API. The one you suggested is not a public api.

modules/action-runner-binary-cache/lambdas/syncer/src/syncer/handler.ts
gertjanmaas: It is also possible to use the org level api: https://developer.github.com/v3/actions/self-hosted-runners/#list-runner-applications-for-an-organization
If we use the Github App in that application we can generate tokens for both org and repo level if needed.

npalm: I suggest we create an issue for this as improvement, the advantages of the current approach is that we do not require any auth.

PR #16, opened 2020-05-13
Feature/config for lambda

npalm:
- Add module to download release lambda
- Add release workflow, extending the current workflows does not work since we would like to have one release.
- Refactor and clean up terraform to support cross account deployment


npalm:
@gertjanmaas read, please can you have a look?

PR #36, opened 2020-06-06
tag created EC instances with environment

gierdo:
Created EC2 instances were not tagged with the environment.

However, the instance-number limitation and downscaling only considers EC2 instances with a correct 'Environment' tag.

-> This pull request adds the 'Environment' tag to created EC2 instances
-> Fixes the instance number limitation
-> Fixes downscaling


npalm:
@gierdo Nice to see you are working on a PR, I assume the PR is still draft mode?

The tag environment should be applied via the default setup via the launch template responsible for creating the ec2 instances. You can chehck the main.tf (https://github.com/philips-labs/terraform-aws-github-runner/blob/88bc451e5e36358666a220d9c737ad6639da03a1/main.tf#L2) where via the locals the tags with the Environent tag is passed to the launch template in the runner module.

gierdo:
I see! Thank you for clearing that up!
I "ported" the deployment to cloudformation, using only the lambdas. 
I obviously have missed this bit.

PR #41, opened 2020-06-22
Add semantic release

nielswijers:


.github/workflows/release.yml
```diff
@@ -2,73 +2,83 @@ name: Release build
...
-          draft: true
-          prerelease: true
-      - name: Extract tag name
-        id: tag_name
-        run: echo ::set-output name=TAG::${GITHUB_REF##*/}
+      - name: Add git and zip
+        run: apk update && apk add git zip
+      # - name: Create Release
+      #   id: create_release
+      #   uses: actions/create-release@latest
```
npalm, line 23: Please remove commented code

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
+## [0.2.2](https://github.com/nielswijers/terraform-aws-github-runner/compare/v0.2.1...v0.2.2) (2020-06-22)
```
npalm, line 1: Please can you revert the changelog

package.json
```diff
@@ -0,0 +1,15 @@
+{
```
npalm, line 1: Maybe it is not clear to me, but in the build spec you run explicit yarn add commands for dependecies listed below. Would it not be logical to either add the package.json and run yarn instaal or add package explicit as done in the build spec.

npalm:
@nielswijers thanks for the work, we have already a semantic release setup. Done in a slightly different way

PR #76, opened 2020-07-20
be consistent with new lines

bendavies:
### Description of the Change

Introduces consistency where newlines are used.

### Release Notes

- Introduces consistency where newlines are used.


bendavies:
Afraid not, I did it manually

PR #79, opened 2020-07-20
fix: Remove Orphan AWS runners

### LINKED ISSUE #35: If termination of EC2 instance fails, it is never cleaned up
If termination of an EC2 instance fails the code never cleans up the runner on the next pass.

We should handle this better and clean the EC2 instance up in a next attempt.

npalm:
Remove potential orphan AWS instances 

Closes #35

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
@@ -74,9 +74,11 @@ export async function scaleDown(): Promise<void> {
           repo: repo.repoName,
         });
 
+    let orphanEc2Runner = true;
     for (const ghRunner of registered.data.runners) {
       const runnerName = ghRunner.name as string;
       if (runnerName === ec2runner.instanceId && runnerMinimumTimeExceeded(ec2runner, minimumRunningTimeInMinutes)) {
+        orphanEc2Runner = false;
         try {
           const result = enableOrgLevel
             ? await githubAppClient.actions.deleteSelfHostedRunnerFromOrg({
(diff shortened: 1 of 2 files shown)


modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -74,9 +74,11 @@ export async function scaleDown(): Promise<void> {
           repo: repo.repoName,
         });
 
+    let orphanEc2Runner = true;
     for (const ghRunner of registered.data.runners) {
       const runnerName = ghRunner.name as string;
       if (runnerName === ec2runner.instanceId && runnerMinimumTimeExceeded(ec2runner, minimumRunningTimeInMinutes)) {
```
gertjanmaas, line 80: Split up this if statement to determine if a EC2 instance is orphaned. Then only try to delete if the `runnerMinimumTimeExceeded` function returns true

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -100,5 +102,15 @@ export async function scaleDown(): Promise<void> {
         }
       }
     }
+
+    // Remove orphan AWS runners.
+    if (orphanEc2Runner && runnerMinimumTimeExceeded(ec2runner, minimumRunningTimeInMinutes)) {
```
gertjanmaas, line 107: Remove the `runnerMinimumTimeExceeded` check if the if statement on line 80 is changed (see comment above)

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
npalm: @gertjanmaas created a fix

PR #83, opened 2020-07-20
RFC: allow more easily customizing the runner user data to allow non Amazon Linux AMIs

bendavies:
### Description of the Change

Hi there,

I noticed that this module is quite heavily coupled to amazon linux: the hard coded non configurable user data and runner user (`ec2-user`).

I want to run my runners on a custom AMI based on ubuntu, so I need to be able to provide custom user data and user.

I've made a very quick change here to gather some feedback on if this approach would be acceptable or not.

I have maintained the current defaults of this module making it usable on Amazon linux, but made it more configurable for my use case.

Description of changes:

1. replace `var.runner_as_root` with `var.runner_user`, for maximum flexibility.
1. add `var.runner_user_data`, maintaining current default, for maximum flexibility.
1. ~~remove `var.userdata_pre_install` and `var.userdata_post_install` as they don't really provide much flexibility and do not solve my use case. Prefer `var.runner_user_data`.~~

TODO: 

- [ ] Please can you add an extra example, based on default where you provide the setup with for example  ubunut, I assume you have all the code all ready.
- [ ] If find a good spot in the readme feel free to add a note about this option, the feature make really difference.

Thanks for the consideration!

### Alternate Designs

Not sure. Can't yet think of anything.

### Possible Drawbacks

None?

### Verification Process

Minimal at the moment  - this is an RFC rather than a finished PR.

### Release Notes

TODO

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -123,10 +123,31 @@ variable "instance_type" {
   default     = "m5.large"
 }
 
-variable "runner_as_root" {
-  description = "Run the action runner under the root user."
-  type        = bool
-  default     = false
+variable "ami_filter" {
+  description = "List of maps used to create the AMI filter for the action runner AMI."
+  type        = map(list(string))
+
+  default = {
+    name = ["amzn2-ami-hvm-2.*-x86_64-ebs"]
+  }
+}
+
+variable "ami_owners" {
+  description = "The list of owners used to select the AMI of action runner instances."
+  type        = list(string)
+  default     = ["amazon"]
+}
+
+variable "runner_user_data" {
+  description = "User data to provide when launching the instance. Defaults to a user data for setting up the runner on Amazon Linux."
+  type        = string
+  default     = null
+}
+
+variable "runner_user" {
+  description = "The user to run the action runner under when using the default provided user_data (see var.runner_user_data)."
+  type        = string
+  default     = "ec2-user"
 }
 
 variable "runners_maximum_count" {
@@ -152,13 +173,15 @@ variable "kms_key_id" {
   type        = string
   default     = null
 }
+
 variable "userdata_pre_install" {
+  description = "Script to be ran before the GitHub Actions runner is installed on the EC2 instances. Only used if the default var.runner_user_data is not set."
   type        = string
   default     = ""
-  description = "Script to be ran before the GitHub Actions runner is installed on the EC2 instances"
 }
+
 variable "userdata_post_install" {
+  description = "Script to be ran after the GitHub Actions runner is installed on the EC2 instances. Only used if the default var.runner_user_data is not set."
   type        = string
   default     = ""
-  description = "Script to be ran after the GitHub Actions runner is installed on the EC2 instances"
 }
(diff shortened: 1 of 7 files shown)


npalm:
@bendavies thanks for the PR, I will do my best to check the PR in the next days. I only did a quick scan. Pleae can you revert the pre/post install back. I really would like to give the user the option to extend the default. So add custom script to the pre / post install.

bendavies:
@npalm thanks. I've reverted that change as requested.

variables.tf
```diff
@@ -123,10 +123,15 @@ variable "instance_type" {
   default     = "m5.large"
 }
 
-variable "runner_as_root" {
-  description = "Run the action runner under the root user."
-  type        = bool
-  default     = false
+variable "runner_user_data" {
+  description = "User data to provide when launching the instance. Defaults to a user data for setting up the runner on Amazon Linux."
+  type        = string
```
npalm, line 128: Please add here the default value `null` it keeps the module backwards compatible and let the default be the provided user_data by the module

bendavies: good spot!

bendavies:
> @bendavies Really nice the work you did! Just a few comments:
> 
>     * Please can you add an extra example, based on default where you provide the setup with for example  ubunut, I assume you have all the code all ready.
> 
>     * If find a good spot in the readme feel free to add a note about this option, the feature make really difference.
> 
>     * The last comment is inline, please add a default value.

Yep, great. Can do all of this. 
Glad you think this is a good path forward.

npalm:
@bendavies Did you already had the time to create an example, looking forward to try out your extenstion.

bendavies:
> @bendavies Did you already had the time to create an example, looking forward to try out your extenstion.

Yes! Sorry, i'm going to - just a bit busy the last week :)

npalm:
> > @bendavies Did you already had the time to create an example, looking forward to try out your extenstion.
> 
> Yes! Sorry, i'm going to - just a bit busy the last week :)

No problem, I was just wondering

bendavies:
current issue, getting `var.s3_action_runner_url` into my custom user data...

bendavies:
ok, i'll add templating to the user provided `runner_user_data`

npalm:
> ok, i'll add templating to the user provided `runner_user_data`

Did you pushed your updates, I see now new files.

bendavies:
Not fully!

npalm:
@bendavies did you already find a bit of time to work out a small example?

bendavies:
Hi Niek,

I'm on holiday until the end of the month.
I'll submit it then 🤞

bendavies:
Hey - we've taken our CI in a different direction for the moment, so not using this module.
I can complete this PR with an example if you still want to get it merged, but may take a little longer

bendavies:
looks like @sjagoe has done very similar job in #142, and had a bit more time to finish it off, so closing in favour of #142. Thanks @sjagoe !

npalm:
@bendavies thanks for updating and also for you work!

PR #102, opened 2020-07-28
Add support for ARM64 runners

### LINKED ISSUE #101: Add support for ARM64 runners using AWS Graviton/Graviton2 instance-types.
## Problem to solve

The current solution is unable to launch instances compatible to use with GitHub's ARM64 self-hosted runner.

## Intended users

Developers/Teams building for ARM64 (e.g. Raspberry Pi)

## Further details

Benefit: extends support to GitHub Actions pipelines that use ARM64

## Proposal

PR in progress with changes to `runner_architecture` auto-detected from instance type, support for downloading arm64 actions-runner from GitHub, and a patch to account for lack of pre-installed ICU support in .NET Core, required by the arm64 actions-runner.

## Documentation

Will document how to enable arm64 support as well as some gotchas I ran into (some Graviton instances aren't available in all AZs)

## Availability & Testing

Not much? Might need a test case for the change to `syncer` lambda.

## What does success look like, and how can we measure that?

Setting a Graviton/Graviton2 instance type in `example/default/main.tf` and (optionally) specifying subnet AZs in `example/default/vpc.tf` results in a successful stack that can launch functional ARM64 self-hosted runners.

## Other links/references

n/a

bdruth:
### Issue Endorsed by Maintainers

See https://github.com/philips-labs/terraform-aws-github-runner/issues/101

### Description of the Change

Basically tracked down where architecture relevant things were hard-coded and introduced variables to manage them. Two main things: the AMI and the actions-runner download from GitHub that's sync'd to the S3 bucket.

### Alternate Designs

?

### Possible Drawbacks

Everything defaults to the existing behavior and only changes if it detects a Graviton/Graviton2 instance-type. Probably introduced some maintenance as new instance types are released to correctly detect them as ARM64.

### Verification Process

* Destroyed/created the `example/default` stack with a recent Graviton2 instance type (c6g.large).
* Ran tests for the `syncer`
* Verified successful runner creation & association to GitHub repo.

### Release Notes

- The module now supports ARM64 self-hosted runners on AWS Graviton/Graviton2 instance types (a1/*6g*)

Checklist:
* [x] Basic POC/functionality working
* [x] Determine if conditional for instance-type should be moved to submodule
* [x] Write test for `syncer`
* [x] Update README/documentation for auto-detect logic and AZ constraints
* [x] All repo checks are passing

Closes #101


npalm:
@bdruth nice work!, we will do our best to check the PR later this week.

compiaffe:
@bdruth Thank you for this MR, just what we needed :)

From what I understand this will not handle deciding whether to spin up a x86 machine or a arm64 machine based on the tags/labels of the run, correct?

npalm:
> @bdruth Thank you for this MR, just what we needed :)
> 
> From what I understand this will not handle deciding whether to spin up a x86 machine or a arm64 machine based on the tags/labels of the run, correct?

Spinning up based on the labels in the workflow requires that we extend the lambda scale up function to retrieve and parse the workflow. GitHub does not support any simple API to retrieve the information, and is not part of the event.

bdruth:
@compiaffe  what @npalm said :-) there's nothing stopping anyone from having multiple stacks of runners added to a repo, though? It might be inefficient as `check_run` will launch runners from multiple stacks, but the labels will then sort out which gets used and the next culling will kill off the inactive ones, I think? I haven't tried this, just thinking through it.

npalm:
I fully agree, launching on any check_run was our first approach. There seems no other good alternative. But I think we should find a way to create runners a bit more sophisticated. It seems that every job leads to a check_run, but not completely sure. And eacht workflow seems related to a check_suite.

modules/runners/templates/user-data.sh
```diff
@@ -20,6 +20,56 @@ aws s3 cp ${s3_location_runner_distribution} actions-runner.tar.gz
 tar xzf ./actions-runner.tar.gz
 rm -rf actions-runner.tar.gz
 
+# Patch for ARM64 (no ICU install by default)
```
npalm, line 23: As far I understand the patch is required for running the action runners on a ARM machine. Should this patch not be done in the GitHub action runner? In that case it is an idea to create a issue on the github action runner repo. W

Since patched is added for ARM only, can you add an if statement to apply the pathc only for ARM machines, it would helpfull if you can add in the code a bit more explanation of the patch or some references. Would help when we got in the future issues here.

npalm:
Verified deployments for both with and without ARM, all works

modules/runners/templates/user-data.sh
bdruth: Roger that - I'll see what I can do. I'll also double-check that there isn't a simpler way to do this - a lot of the Google refs for this issue seem to indicate you should "just install icu" or some similar package, but I think when I tried that, it wasn't available for ARM in the Amazon Linux distro, or something along those lines. As it stands, though, GitHub's ARM action runner just doesn't work out-of-the-box on AWS Graviton/Graviton2 instances, so I definitely feel like them fixing it upstream would be the right call. Will see if I can chase down where to file that request or push a PR.

bdruth: Filing a bug at https://github.com/actions/runner - will link here when I've got everything filled in.

bdruth: https://github.com/actions/runner/issues/629

bdruth: Push the "patch" as a PR to the `actions/runnner` repo ... https://github.com/actions/runner/pull/630

bdruth: I'll work on the cleanup you recommended now, and throw in a caveat that this is only needed until https://github.com/actions/runner/pull/630 is merged or otherwise addressed.

bdruth: Pushed https://github.com/philips-labs/terraform-aws-github-runner/pull/102/commits/139cb9c80684e8244bab3493265cfbc0c431001c to address making the ARM/ICU patch conditional @npalm.

bdruth:
> Sorry for the late notice, please can you use the pre-commit hooks. Yest it missing in the contribution guide / docs (is an issue ;) ) The pre-commit hooks will format your terraform code (reason why the build is failing https://github.com/philips-labs/terraform-aws-github-runner/pull/102/checks?check_run_id=933015675). And it will updte the readme with the terraform variables

Crap - sorry I missed this. I caught the formatting errors last time, will get this fixed. Mea culpa.

bdruth:
OK @npalm - I ran the pre-commit hook and everything looks good now. It made some changes in the README files, too? Not sure what to make of that, though? It wasn't even in files that I edited, mostly, and in the one README I edited, not in the area that I edited.

npalm:
@bdruth thanks for your contribution, changed is merged.

PR #113, opened 2020-08-06
Feature/idle runners

npalm:
Ensure we can keep idle runners alive during specified periods to speed up builds.


modules/runners/variables.tf
```diff
@@ -190,3 +190,14 @@ variable "runner_architecture" {
...
   default     = "x64"
 }
+
+
+variable "idle_config" {
+  description = "List of time period that can be defined as cron expression to keep a minimum ammount of runners active instead of scaling down to 0. By defining this ist you can ensure that in time periods that match the cron expression within 5 secons a runner is kept idle."
+  type = list(object({
+    cron       = string
+    timeZone   = number
+    indleCount = number
```
bendavies, line 200:
```suggestion
    idleCount = number
```

modules/runners/variables.tf
```diff
@@ -190,3 +190,14 @@ variable "runner_architecture" {
   type        = string
   default     = "x64"
 }
+
+
+variable "idle_config" {
+  description = "List of time period that can be defined as cron expression to keep a minimum ammount of runners active instead of scaling down to 0. By defining this ist you can ensure that in time periods that match the cron expression within 5 secons a runner is kept idle."
```
bendavies, line 196:
```suggestion
  description = "List of time period that can be defined as cron expression to keep a minimum amount of runners active instead of scaling down to 0. By defining this list you can ensure that in time periods that match the cron expression within 5 seconds a runner is kept idle."
```

bendavies:
how does the cron expression relate to the 5 second check? can you give an example?

modules/runners/variables.tf
npalm: Thx for fixing the typo's, this one I will do in my IDE :)

npalm:
Checked with the default example by adding the following config:

```hcl
  idle_config = [{
    cron      = "* * 12 * * *"
    timeZone  = "Europe/Amsterdam"
    idleCount = 1
  }]

  # trigger the lambda every 1 minute for test purpose
  scale_down_schedule_expression = "cron(* * * * ? *)"
```

Next trigger een build, after a while a runner is available to process the build Now monitor the cloudwatch logs for scale down:
```2020-08-07T10:59:14.348Z	5b2dc212-0aa7-405d-b843-17bd79040d57	INFO	Runner 'i-0a25093117733be1f' will kept idle.
2020-08-07T11:00:15.156Z	c21b2a97-7d1b-475f-8429-cd39294effb3	INFO	Runner terminated.[object Object]
2020-08-07T11:00:15.157Z	c21b2a97-7d1b-475f-8429-cd39294effb3	INFO	AWS runner instance 'i-0a25093117733be1f' is terminated and GitHub runner is de-registered.```

PR #144, opened 2020-08-18
Split user data in more parts

npalm:
experiment for now


npalm:
not relevant any more due to #144

PR #145, opened 2020-08-18
Add wildcard for log events in arbitrary streams

HenryNguyen5:
Recently tried deploying v0.4.0 version of this module. None of the lambdas can log because the resource restriction stops log events from being created on log streams.


npalm (approved):
@HenryNguyen5 thanks for fixing this bug!

PR #165, opened 2020-08-25
Feature/fetch prerelease binaries

gertjanmaas:
### Description of the Change

Due to updating issues with regards to prerelease binaries (see https://github.com/actions/runner/issues/581 ), there is now an option to allow the binaries syncer to fetch these prerelease binaries. It is disabled by default.

Hopefully Github will give more guidance on how to use prereleases soon.

This is based on the work in #141 by @sjagoe

### Possible Drawbacks

Prereleases might not be stable, hence why this option is disabled by default.

### Verification Process

As there currently are no prereleases, I have updated the unit tests mock responses to real responses from the Github API and changed the `prerelease` flag in that response. See the changes in `modules/runner-binaries-syncer/lambads/runner-binaries-syncer/test/resources`.

Also deployed it to my test environment and made sure the binaries lambda runs with both options. Can't test more because there is no prerelease yet.

### Release Notes

Already updated the `CHANGELOG.md` with:

- feat: Added option to binaries syncer to upgrade to prereleases, preventing any auto-updating on startup. Option `runner_allow_prerelease_binaries` is disabled by default.


modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/handler.test.ts
```diff
@@ -65,12 +85,63 @@ describe('Synchronize action distribution.', () => {
...
+    expect(mockS3.upload).toBeCalledTimes(1);
+    const s3JsonBody = mockS3.upload.mock.calls[0][0];
+    expect(s3JsonBody['Tagging']).toEqual('name=actions-runner-linux-x64-2.273.0.tar.gz');
+  });
+
+  it('Distribution should not update to prerelease if there is a newer release.', async () => {
+    process.env.GITHUB_RUNNER_ALLOW_PRERELEASE_BINARIES = 'true';
+    const releases = listReleases;
+    releases[0].prerelease = false;
+    releases[1].prerelease = true;
```
npalm, line 123: Would it not be more logical that the pre-relese is the first element in the list `[0]`. Do we have a test case with a list with multiple releases and pre-releases?

modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/handler.ts
```diff
@@ -70,10 +89,18 @@ async function uploadToS3(s3: S3, cacheObject: CacheObject, actionRunnerReleaseA
...
+  const re = new RegExp(/^(true|1|on)$/i);
+  return re.test(s.trim());
+}
+
 export const handle = async (): Promise<void> => {
   const s3 = new AWS.S3();
 
-  const runnerArch = process.env.GITHUB_RUNNER_ARCHITECTURE || 'x64'
+  const runnerArch = process.env.GITHUB_RUNNER_ARCHITECTURE || 'x64';
+  const fetchPrereleaseBinaries = process.env.GITHUB_RUNNER_ALLOW_PRERELEASE_BINARIES
```
npalm, line 101: In the lambda `runner` we use the lib `yn` to parse input boolean, why not here?

modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/handler.ts
```diff
@@ -31,13 +31,32 @@ interface ReleaseAsset {
...
+  const latestPrereleaseIndex = assetsList.data.findIndex((a) => a.prerelease === true);
+  const latestReleaseIndex = assetsList.data.findIndex((a) => a.prerelease === false);
+
+  let asset = undefined;
+  if (fetchPrereleaseBinaries && latestPrereleaseIndex < latestReleaseIndex) {
+    asset = assetsList.data[latestPrereleaseIndex];
+  } else if (latestReleaseIndex != -1) {
+    asset = assetsList.data[latestReleaseIndex];
+  }
+  if (!asset) {
```
npalm, line 56: this if statement does make sense, since `asset` is alredy set by default to `undefined`. I would prefer to have a else clause to the if / else if statement to handle the behavior if non of the above applies.

npalm (changes requested):
Just a few remarks / questions

modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/handler.test.ts
gertjanmaas: Yes that is the most logical, but it theoretically could happen. I just wanted a test that would cover the scenario where it would pick the newest release over a prerelease. I have not created a test for multiple releases and pre-releases, but I'm confident these tests wouldn't add much over the tests that are already there.

npalm (approved):

PR #202, opened 2020-09-11
Fix pagination with listing self-hosted runners

### LINKED ISSUE #227: If there are more than 100 runners registered, scale-down fails
```scale-down.ts:113``` and ```scale-down.ts:116``` use ```actions.listSelfHostedRunnersForOrg``` and ```actions.listSelfHostedRunnersForRepo``` directly.
The returned object contains an array ```data.runners```, containing up to *100* registered runners according to the API documentation.

If there are more than 100 runners registered, the additional runners are not considered for scaling down.

### PR DESCRIPTION
**Pull request by HenryNguyen5**
Fixes #227

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
@@ -124,11 +124,10 @@ describe('scaleDown', () => {
       },
     }));
 
-    mockOctokit.paginate.mockImplementation(() => {
+    mockOctokit.paginate.mockImplementationOnce(() => {
       return DEFAULT_REGISTERED_RUNNERS;
     });
 
-    function deRegisterRunnerGithub(id: number): any {}
     mockOctokit.actions.deleteSelfHostedRunnerFromRepo.mockImplementation((repo) => {
       if (repo.runner_id === 103) {
         throw Error();
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
@@ -1,11 +1,7 @@
-import { mocked } from 'ts-jest/utils';
-import { scaleDown } from './scale-down';
 import moment from 'moment';
-import { createAppAuth } from '@octokit/auth-app';
-import { Octokit } from '@octokit/rest';
+import { mocked } from 'ts-jest/utils';
 import { listRunners, terminateRunner } from './runners';
-import { stringify } from 'querystring';
-import { ScalingDownConfig } from './scale-down-config';
+import { scaleDown } from './scale-down';
 
 jest.mock('@octokit/auth-app', () => ({
   createAppAuth: jest.fn().mockImplementation(() => jest.fn().mockImplementation(() => ({ token: 'Blaat' }))),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by npalm** [comment]
Thanks for the PR will check asap.

File: modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts [hunk 1]
```diff
@@ -131,10 +124,7 @@ describe('scaleDown', () => {
       },
     }));
 
-    mockOctokit.actions.listSelfHostedRunnersForOrg.mockImplementation(() => {
-      return DEFAULT_REGISTERED_RUNNERS;
-    });
-    mockOctokit.actions.listSelfHostedRunnersForRepo.mockImplementation(() => {
+    mockOctokit.paginate.mockImplementation(() => {
```
npalm, line 127: Just wondering if we can find a better way to mock the specific function, this will mock all pagnate calls. For now no issue but create potential problems in the future.

npalm (commented):
All works, one comment that I would like to discuss.

modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
gertjanmaas: Maybe it is possible to add a check to tests, something like:
```javascript
expect(mockOcktokit.paginate).toBeCalledWith(mockOctokit.actions.listSelfHostedRunnersForOrg, expect.any());
```

Haven't tested if it works.

HenryNguyen5: I could scope the mocking of the implementation to each individual test, so that it is restored inbetween each test. Does that work?

HenryNguyen5:
Been testing this PR on our own org, I've discovered that the lambda ends up timing out because if a large amount of runners are returned from the call, we make a lot of requests to github API, I will need to implement a local cache to speed up the lambda execution time.

npalm:
> Been testing this PR on our own org, I've discovered that the lambda ends up timing out because if a large amount of runners are returned from the call, we make a lot of requests to github API, I will need to implement a local cache to speed up the lambda execution time.

How many runners do you have on averagen?

HenryNguyen5:
~30 or so

modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
HenryNguyen5: I noticed it's already been scoped to `scaleDown`, I think it's fine for now unless you're uncomfortable with the current test implementation

HenryNguyen5:
@npalm added function memoization, should reduce execution time by around 75% per my testing

modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
npalm: The risk is when we add pagination to more github call they all now handled by the same mock, all calls to ` mockOctokit.paginate` will gets the same mocked response. Wil works as long we not paginate any other call. If would prefer if we can come with a fix. Otherwise I am fine with creating an issue and keep it for the moment. @gertjanmaas can you have a look again?

HenryNguyen5:
@npalm @gertjanmaas anything i can do here to get this merged in?

gertjanmaas (approved):
LGTM

npalm (approved):
Sorry for the delay, reviewd and tested. All good. thanks once again 👍

PR #228, opened 2020-09-21
fix scaling down with more than 100 registered runners

gierdo:
This PR makes sure that all registered runners are considered for scaling down, not just 100.
This PR fixes https://github.com/philips-labs/terraform-aws-github-runner/issues/227


HenryNguyen5:
@gierdo I've covered this issue in https://github.com/philips-labs/terraform-aws-github-runner/pull/202 😅 , my PR makes use of caching too so you don't get lambda execution timeouts from lots of pagination calls.

gierdo:
> @gierdo I've covered this issue in #202  , my PR makes use of caching too so you don't get lambda execution timeouts from lots of pagination calls.

Nice!
This is, of course, much more elegant than my naive fix.

Issue #104, opened 2020-07-30
runners-scale-up fails with 'AuthFailure.ServiceLinkedRoleCreationNotPermitted'


compiaffe:
## Summary

When following the readme, using the example configuration and adjusting the Github app permissions as per https://github.com/philips-labs/terraform-aws-github-runner/issues/100#issuecomment-665923045 the `scale-up` lambda fails to create the EC2 instance due to `ServiceLinkedRoleCreationNotPermitted`

## Steps to reproduce

- Do step 1 of Github app setup
- Checkout terraform-aws-github-runner repo, cd into example folder
- Download lambda zips
- Create `terraform.tfvars` file with Github App credentials
- run `terraform init && terraform apply`
- Trigger a build on Github

## What is the current *bug* behavior?

Github app sends webhook, webhook lambda forwards it, scaleup-lambda throws error:
```
...
ERROR	AuthFailure.ServiceLinkedRoleCreationNotPermitted: The provided credentials do not have permission to create the service-linked role for EC2 Spot Instances.
    at Request.extractError (/var/task/index.js:41424:35)
    at Request.callListeners (/var/task/index.js:47771:20)
    at Request.emit (/var/task/index.js:47743:10)
    at Request.emit (/var/task/index.js:18467:14)
    at Request.transition (/var/task/index.js:17801:10)
    at AcceptorStateMachine.runTo (/var/task/index.js:26145:12)
    at /var/task/index.js:26157:10
    at Request.<anonymous> (/var/task/index.js:17817:9)
    at Request.<anonymous> (/var/task/index.js:18469:12)
    at Request.callListeners (/var/task/index.js:47781:18) {
  code: 'AuthFailure.ServiceLinkedRoleCreationNotPermitted',
  time: 2020-07-30T15:03:24.631Z,
  requestId: 'c7bab39e-b75c-4e7d-bc29-6622b3d4ddb1',
  statusCode: 403,
  retryable: false,
  retryDelay: 68.19342592727871
}
```

## What is the expected *correct* behavior?

Scale up lambda should create EC2 instance

## Possible fixes

I'm sure this is a IAM permissions issue. I am rather new to both AWS and terraform and am not sure in which of them this needs to be solved and how to go about it.
Would be great to get some pointers.

cmcconnell1:
Hey @compiaffe  ,

Assume that you are using the v0.2.0 tag?
And, curious if you always see this error now?  
[...]

PR #229, opened 2020-09-24
Add create role policy to scale-up.tf

toots:
This seems to have taken care of my issues with #104

### DIFF
--- a/modules/runners/scale-up.tf
+++ b/modules/runners/scale-up.tf
@@ -85,3 +85,9 @@ resource "aws_iam_role_policy" "scale_up_logging" {
     log_group_arn = aws_cloudwatch_log_group.scale_up.arn
   })
 }
+
+resource "aws_iam_role_policy" "service_linked_role" {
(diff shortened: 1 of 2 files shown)


modules/runners/scale-up.tf
```diff
@@ -85,3 +85,9 @@ resource "aws_iam_role_policy" "scale_up_logging" {
     log_group_arn = aws_cloudwatch_log_group.scale_up.arn
   })
 }
+
+resource "aws_iam_role_policy" "service_linked_role" {
+  name   = "${var.environment}-service_linked_role"
```
npalm, line 90: Please can you make this resource optional, with a default of `true`. This still give users the option to opt out and use a general service linked role in their account.

npalm (changes requested):
Looks good, please see comment. Tested as well.

npalm:
@toots did you had time to check the comment?

npalm:
Solved will merge asap

npalm:
Merged manually @toots thanks

PR #292, opened 2020-10-26
expand options for sourcing lambda to include S3

eky5006:
Submitting a small change to expand the options for sourcing the lambda functions to include S3. Using an S3 bucket would be an alternative to providing local files directly and providing the bucket itself would be up to the user.

We found that a change like this is personally better for our workflow and could be useful for others too. Thanks for the great work on this project!


npalm:
Thx will check asap

npalm (approved):
@eky5006 looks good to me, I wade some small documentation adjustments myself.

PR #325, opened 2020-11-08
Update user data script to confirm install prompts

kernel-io:
pass a -y option to amazon-linux-extras install so it doesn't hang waiting for user input

### DIFF
--- a/modules/runners/templates/user-data.sh
+++ b/modules/runners/templates/user-data.sh
@@ -6,7 +6,7 @@ ${pre_install}
 yum update -y
 
 # Install docker
-amazon-linux-extras install docker
+amazon-linux-extras install -y docker
 service docker start
 usermod -a -G docker ec2-user
 


npalm (approved):
@kernel-io thanks for the small fix. I have checked the fix and it all works fine. But what I don't understand why it for us (and guess others) is working without "-y". Also the aws docs or cli help of amazon-linux-estra does not mention anything about the -y paramater. Which is normal to use with yum or apt for example. I also checked on an instance and find out that for instally pythhon the -y is really required for an automated install. 

Do you have any idea why you require the -y? Did you override the AMI?

kernel-io:
I'm using the default AMI, I was having issues with the system booting, so I added ssh credentials to the machine and logged in (while removing the user data script). I then ran the user data script from within my shell and it wouldn't proceed without me acknowledging the docker install (either by pressing y, or passing in -y). 

There could have been something else happening with my setup though. 

I'm assuming that Amazon checks if you are running `amazon-linux-extras` as part of a user data script and then automatically confirms installation (as no one else seems to be having this issue?)

PR #359, opened 2020-11-23
Updates to support typescript 4.1

npalm:
Add explicit type info to support resolve in typsescript 4.1+


modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/handler.ts
```diff
@@ -72,7 +72,7 @@ async function uploadToS3(s3: S3, cacheObject: CacheObject, actionRunnerReleaseA
     Body: writeStream,
   }).promise();
 
-  await new Promise((resolve, reject) => {
+  await new Promise<void>((resolve, reject) => {
```
JeroenKnoops, line 75: Is this the only change necessary?

JeroenKnoops (approved):

PR #392, opened 2020-12-10
Allow configuring additional security groups

surminus:
We've started recently looking to use this module and to fit it into our current infrastructure we'd like to be able to attach additional security groups to the runners. This will enable us to SSH to them which will help us debug some small customisations we're making in the user data.

### DIFF
--- a/modules/runners/variables.tf
+++ b/modules/runners/variables.tf
@@ -278,3 +278,9 @@ variable "runner_log_files" {
     }
   ]
 }
+
+variable "runner_additional_security_group_ids" {
+  description = "(optional) List of additional security groups IDs to apply to the runner"
+  type        = list(string)
+  default     = []
+}
--- a/variables.tf
+++ b/variables.tf
@@ -296,3 +296,9 @@ variable "runner_log_files" {
     }
   ]
 }
+
+variable "runner_additional_security_group_ids" {
+  description = "(optional) List of additional security groups IDs to apply to the runner"
+  type        = list(string)
+  default     = []
+}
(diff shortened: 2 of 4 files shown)


modules/runners/variables.tf
```diff
@@ -278,3 +278,9 @@ variable "runner_log_files" {
     }
   ]
 }
+
+variable "additional_security_group_ids" {
```
npalm, line 282: Can you rename `additional_security_group_ids` to `runner_ additional_security_group_ids` whould be a bit more aligned with other variable names.

variables.tf
```diff
@@ -296,3 +296,9 @@ variable "runner_log_files" {
     }
   ]
 }
+
+variable "additional_security_group_ids" {
```
npalm, line 300: Can you rename `additional_security_group_ids` to `runner_ additional_security_group_ids` whould be a bit more aligned with other variable names.

npalm (changes requested):

jpalomaki:
For better readability, would it make sense to allow a user to pass in security group _names_ using https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/security_groups?

variables.tf
surminus: Done!

npalm (approved):

surminus:
> For better readability, would it make sense to allow a user to pass in security group names, using this data source to then look up the IDs?

I personally think using IDs is more idiomatic for Terraform resources, but I'm happy to make this change if the consensus is to use names instead.

surminus:
Just realised this got merged in a little while ago [here](https://github.com/philips-labs/terraform-aws-github-runner/commit/99f7cc931f94c4a34773bad20d507d5f4f0121e0). Thanks! :tada:

Issue #409, opened 2020-12-15
Scale up runner is using deprecated attribute in function all createAppAuth


npalm:
```2020-12-15T19:55:18.898Z	uuid	WARN	Deprecation: [@octokit/auth-app] "createAppAuth({ id })" is deprecated, use "createAppAuth({ appId })" instead    at Object.createAppAuth (/var/task/index.js:1912:14)    at /var/task/index.js:25303:27    at Generator.next (<anonymous>)    at fulfilled (/var/task/index.js:25276:58)    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

PR #410, opened 2020-12-15
Fix/scale up lambda deprecated attribute

npalm:
See #409


npalm:
Verified with runners connected to a single repo.

JeroenKnoops (approved):

PR #412, opened 2020-12-17
feat(ghes): Support for GitHub Enterprise Server

mcaulifn:
- Updates lambdas to support GHES URL
- Updates TF to support GHES and deploying lambda in VPC


modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -13,49 +10,21 @@ export interface ActionRequestMessage {
...
-
 export const scaleUp = async (eventSource: string, payload: ActionRequestMessage): Promise<void> => {
   if (eventSource !== 'aws:sqs') throw Error('Cannot handle non-SQS events!');
   const enableOrgLevel = yn(process.env.ENABLE_ORGANIZATION_RUNNERS, { default: true });
   const maximumRunners = parseInt(process.env.RUNNERS_MAXIMUM_COUNT || '3');
   const runnerExtraLabels = process.env.RUNNER_EXTRA_LABELS;
   const environment = process.env.ENVIRONMENT as string;
-  const githubAppAuth = await createGithubAppAuth(payload.installationId);
-  const githubInstallationClient = await createInstallationClient(githubAppAuth);
+  const ghesBaseUrl = process.env.GHES_URL as string;
```
npalm, line 19: Maybe we can find a bitter name of the variable, in this case it always looks like some ghes base url is used. In case empty, an empty string is passed into the auth methods and no base url is set. So it looks all good but a bit confusing due to nameing in my opinion.

modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
```diff
@@ -0,0 +1,57 @@
+import { Octokit } from '@octokit/rest'
```
npalm, line 1: Thanks for splitting this logic!

modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
```diff
@@ -0,0 +1,57 @@
...
+import { request } from '@octokit/request'
+import { createAppAuth } from '@octokit/auth-app'
+import { Authentication, StrategyOptions } from '@octokit/auth-app/dist-types/types'
+import { OctokitOptions } from '@octokit/core/dist-types/types'
+import { decrypt } from './kms'
+
+export async function createOctoClient(token: string, githubServerUrl: string = ''): Promise<Octokit> {
```
npalm, line 8: Maybe name the parameter also githubBaseUrl

npalm (changes requested):
@mcaulifn thanks for this great contribution, you did really a great job not only by adding GHES but also updating and improving the code. Just a few remarks for now. 

I only was able to verify the cloud runners, no impact noted. The only TF change is passing the GHES URL to the lambda, in the cloud case empty of source. I was not able to verify the GHES runners since I have no access to a GHES setup.

modules/runners/variables.tf
```diff
@@ -278,3 +278,27 @@ variable "runner_log_files" {
...
 }
+
+variable "ghes_url" {
+  description = "GitHub Enterprise Server URL. DO NOT SET IF USING PUBLIC GITHUB"
+  type        = string
+  default     = ""
+}
+
+variable "lambda_subnet_ids" {
+  description = "List of subnets in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`."
```
npalm, line 289: Assume you mean the lambda, not the runner.

variables.tf
```diff
@@ -296,3 +296,27 @@ variable "runner_log_files" {
...
+}
+
+variable "lambda_subnet_ids" {
+  description = "List of subnets in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`."
+  type        = list(string)
+  default     = []
+}
+
+variable "lambda_security_group_ids" {
+  description = "List of subnets in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`."
```
npalm, line 313: Assume you mean the security group for the lambda

modules/runner-binaries-syncer/runner-binaries-syncer.tf
```diff
@@ -23,6 +23,10 @@ resource "aws_lambda_function" "syncer" {
       GITHUB_RUNNER_ALLOW_PRERELEASE_BINARIES = var.runner_allow_prerelease_binaries
     }
   }
+  vpc_config {
```
deverton, line 26: Would this cause issues if you don't want the lambda to be deployed inside the VPC? You probably want something like

```hcl
  dynamic "vpc_config" {
    for_each = var.lambda_subnet_ids != null && var.lambda_security_group_ids != null ? [true] : []
    content {
      security_group_ids = var.lambda_security_group_ids
      subnet_ids         = var.lambda_subnet_ids
    }
  }
```

so that the `vpc_config` block is only generated when needed.

modules/runners/scale-down.tf
```diff
@@ -36,8 +36,13 @@ resource "aws_lambda_function" "scale_down" {
       GITHUB_APP_CLIENT_ID            = var.github_app.client_id
       GITHUB_APP_CLIENT_SECRET        = local.github_app_client_secret
       SCALE_DOWN_CONFIG               = jsonencode(var.idle_config)
+      GHES_URL                        = var.ghes_url
     }
   }
+  vpc_config {
```
deverton, line 42: See comment against other lambda definition about making the `vpc_config` block dynamic.

modules/runners/scale-up.tf
```diff
@@ -40,8 +40,13 @@ resource "aws_lambda_function" "scale_up" {
       SUBNET_IDS                  = join(",", var.subnet_ids)
       LAUNCH_TEMPLATE_NAME        = aws_launch_template.runner.name
       LAUNCH_TEMPLATE_VERSION     = aws_launch_template.runner.latest_version
+      GHES_URL                    = var.ghes_url
     }
   }
+  vpc_config {
```
deverton, line 46: See comment against other lambda definition about making the vpc_config block dynamic.

variables.tf
```diff
@@ -296,3 +296,27 @@ variable "runner_log_files" {
...
+  type        = list(string)
+  default     = []
+}
+
+variable "key_name" {
+  description = "Key pair name"
+  type        = string
+  default     = ""
```
deverton, line 321: Would `null` be a better default?

modules/runner-binaries-syncer/runner-binaries-syncer.tf
npalm: I had the same feeling, but yesterday I upgraded my test env (default example). With no change in the config. So this dynamic block is not causing issues.

deverton: Ah okay, cool. The [terraform-aws-lambda](https://github.com/terraform-aws-modules/terraform-aws-lambda/blob/252e5934fd2f2710863bca689572264c00b8ea77/main.tf#L68) does this trick so I thought it might be necessary.

npalm: That trick looks a bit more safe indeed.

modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
mcaulifn: The confusing nature of Octokit is causing confusion here (and elsewhere). For `OctokitOptions`, the URL needs to be `ghes.company.com/api/v3` or it defaults to `api.github.com`. I'm thinking `ghesApiUrl` to be consistent with what is being passed in.

modules/runners/variables.tf
mcaulifn: oops

modules/runner-binaries-syncer/runner-binaries-syncer.tf
mcaulifn: That is a bit more explicit. I have no issues adding this in.

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
mcaulifn: `Octokit` has been less than helpful in this case. When a a base URL (`ghes.company.com`) is provided, append `api/v3` to make the full URL. If nothing is set, `Octokit` defaults to `api.github.com`. The intention is to only have a value if it is GHES.

I'm relatively new to TS so not sure I did most of this right. I was tempted to re-write in Python 🐍 but used this as a learning opportunity. 😄

variables.tf
```diff
@@ -78,7 +78,7 @@ variable "runners_lambda_zip" {
 variable "runners_scale_up_lambda_timeout" {
   description = "Time out for the scale down lambda in seconds."
   type        = number
-  default     = 60
+  default     = 180
```
npalm, line 81: Just a small question, I assume you increased this default since it take a bit more time to boot a lambda in a VPC?

npalm (commented):
@mcaulifn looks all great, great all the work you did. Just one simple question

variables.tf
mcaulifn: No, lambda launches in VPC are much faster now. And I think this timer starts after it boots. When there were issues connecting to GHES, the lambda would timeout before the connection to GH would. This exposes that error. The other option would be to see if octokit can shorten the timeout.

npalm:
@gertjanmaas please can you have a look to this PR? To me it looks all good.

modules/runners/lambdas/runners/.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "printWidth": 120,
   "singleQuote": true,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "semi": false,
```
gertjanmaas, line 5: I'd prefer if this was set to `true` so all the code looks the same. Now it's a mix of both semicolons (our default) en non-semicolons.

gertjanmaas (changes requested):
Small comment regarding code formatting

modules/runners/lambdas/runners/.prettierrc
mcaulifn: I will switch it. `false` is the default over here so that is why that made it in 😄

gertjanmaas (approved):

npalm (approved):

mcaulifn:
@npalm Anything else keeping this from being merged?

npalm:
Thanks for pinging me, works quite qell merging via the GitHub app. @mcaulifn once agina, thanks for the great contribution. At this moment we haven't the option to test our selves on GHES. Expecting we have this option in the near feature. For that reason it tooks a bit longer with some discussion on how to proceed. Will make asap a release with this great new feature!

jonico:
@mcaulifn: When testing this with GitHub Enterprise 2.22, I get a 415 status code for the checks API in the scale-up lambda:

```
2021-01-16T00:50:28.129Z f7712571-de6f-571f-ba28-7d0de4e0db49 ERROR RequestError [HttpError]: If you would like to help us test the Checks API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details.
at /var/task/index.js:2756:23
at processTicksAndRejections (internal/process/task_queues.js:97:5) {
status: 415,
headers: {
'access-control-allow-origin': '*',
'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset',
connection: 'close',
'content-length': '288',
'content-security-policy': "default-src 'none'",
'content-type': 'application/json; charset=utf-8',
date: 'Sat, 16 Jan 2021 00:50:28 GMT',
'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
server: 'GitHub.com',
status: '415 Unsupported Media Type',
'strict-transport-security': 'max-age=31536000; includeSubdomains',
'x-content-type-options': 'nosniff',
'x-frame-options': 'deny',
'x-github-enterprise-version': '2.22.5',
'x-github-media-type': 'github.v3; format=json',
'x-github-request-id': '16446f5c-f402-42b9-96dc-f109ce3f7b9e',
'x-ratelimit-limit': '5000',
'x-ratelimit-remaining': '4996',
'x-ratelimit-reset': '1610761771',
'x-runtime-rack': '0.054488',
'x-xss-protection': '1; mode=block'
},
request: {
method: 'GET',
url: 'https://octodemo.com/api/v3/repos/baseline/baseline.github.io/check-runs/503',
headers: {
accept: 'application/vnd.github.v3+json',
'user-agent': 'octokit-rest.js/18.0.12 octokit-core.js/3.2.4 Node.js/12.19.0 (linux; x64)',
authorization: 'token [REDACTED]'
},
request: { hook: [Function: bound bound register] }
},
documentation_url: 'https://docs.github.com/enterprise/2.22/rest/reference/checks#get-a-check-run'
}
```

Could this be because GitHub.com [no longer needs](https://developer.github.com/changes/2020-10-01-graduate-antiope-preview/) the `application/vnd.github.antiope-preview+json`preview header and newer versions of octokit do no longer  include it?

As the header has been removed on GitHub.com pretty recently October 1st 2020 - I could imagine that you have been testing with an older version of octo-rest-js (< 10.0.12), could you point me to the exact version you have used?

jonico:
Update: When I reverted to octokit/rest-js/18.0.9 and removed the octokit/types reference in package.json, and rebuilt the lambda - the error went away 🎉

jonico:
In case anybody likes to have the runners lambda working with GitHub Enterprise Server 2.22: [runners.zip](https://github.com/philips-labs/terraform-aws-github-runner/files/5823765/runners.zip)

npalm:
@mcaulifn do you have the same issue, just checked. Octokit/rest was update early december by dependabot. (https://github.com/philips-labs/terraform-aws-github-runner/pull/386). In case remvoing types solved the problem most likey an issue is introduced in version 18.0.11 see https://github.com/octokit/rest.js/releases.

mcaulifn:
@npalm I can test on Tuesday, along with the actual patch version of 2.22 we have.

mcaulifn:
Seeing same issue against `2.22.3`.  I opened https://github.com/octokit/rest.js/issues/1983 to address. Suggest we revert https://github.com/philips-labs/terraform-aws-github-runner/pull/386 for now.

PR #433, opened 2020-12-30
Update ubuntu example with rootless docker and non privileged user

npalm:
Update example
- stream log to clouwatch
- re-use install and configure from module
- add rootless docker
- install in non privileges user account (runners)

Update module
- refactor user_data, splitting in seperate parts


npalm:
@gertjanmaas can you check once again, needed to rebase.

examples/ubuntu/main.tf
```diff
@@ -27,9 +27,9 @@ module "runners" {
     webhook_secret = random_password.random.result
   }
 
-  webhook_lambda_zip                = "lambdas-download/webhook.zip"
-  runner_binaries_syncer_lambda_zip = "lambdas-download/runner-binaries-syncer.zip"
-  runners_lambda_zip                = "lambdas-download/runners.zip"
+  # webhook_lambda_zip                = "lambdas-download/webhook.zip"
```
gertjanmaas, line 30: Can this be removeD?

gertjanmaas (approved):

examples/ubuntu/main.tf
chenrui333: just curious, why does this get commented?

PR #436, opened 2020-12-30
Upgrade runner dependecies for jest vulnerability (test only)

npalm:


npalm:
- tested with default example

JeroenKnoops (approved):

PR #444, opened 2021-01-04
copy directory *after* installing zip

masterful:
This should allow the `.ci/build` script to cache the `zip` installation layer between builds hopefully improving the build times slightly when starting fresh :smile:

switch places


npalm (approved):

npalm:
@masterful thanks

PR #468, opened 2021-01-17
fix: Swap scale up/scale down timeout description

jonico:
* swapping description of scale up/scale down time variable description


npalm (approved):
Thanks

PR #476, opened 2021-01-18
feat: log groups per type of logging

npalm:
- Sending from the runner instance can be grouped per type of logging.
- Logging is kept by default 180 days instead of 7
- Previous logging group of the runners will be destroyed. To keep the logging group ensure you remove it from the terraform state before running an apply: 

## Keeping logging group
```bashe
export RESOURCE=$(terraform state list | grep "aws_cloudwatch_log_group.runner")
terraform state rm $RESOURCE
```


modules/runners/variables.tf
```diff
@@ -258,23 +258,31 @@ variable "cloudwatch_config" {
 }
 
 variable "runner_log_files" {
-  description = "(optional) List of logfiles to send to cloudwatch."
+  description = "(optional) List of logfiles to send to cloudwatch, will onlybe usded if `enable_cloudwatch_agent` is set to true. Object description: `log_group_name`: Name of the log group, `prefix_log_group`: module will prefix the log group with `/github-self-hosted-runners/<var.environment>`, `file_path`: path to the log file, `log_stream_name`: name of the log stream."
```
gertjanmaas, line 261:
```suggestion
  description = "(optional) List of logfiles to send to cloudwatch, will only be used if `enable_cloudwatch_agent` is set to true. Object description: `log_group_name`: Name of the log group, `prefix_log_group`: If true, the log group name will be prefixed with `/github-self-hosted-runners/<var.environment>`, `file_path`: path to the log file, `log_stream_name`: name of the log stream."
```

variables.tf
```diff
@@ -276,23 +276,31 @@ variable "cloudwatch_config" {
 }
 
 variable "runner_log_files" {
-  description = "(optional) List of logfiles to send to cloudwatch."
+  description = "(optional) List of logfiles to send to cloudwatch, will onlybe usded if `enable_cloudwatch_agent` is set to true. Object description: `log_group_name`: Name of the log group, `prefix_log_group`: module will prefix the log group with `/github-self-hosted-runners/<var.environment>`, `file_path`: path to the log file, `log_stream_name`: name of the log stream."
```
gertjanmaas, line 279:
```suggestion
  description = "(optional) List of logfiles to send to cloudwatch, will only be used if `enable_cloudwatch_agent` is set to true. Object description: `log_group_name`: Name of the log group, `prefix_log_group`: If true, the log group name will be prefixed with `/github-self-hosted-runners/<var.environment>`, `file_path`: path to the log file, `log_stream_name`: name of the log stream."
```

gertjanmaas (commented):
Couple of small typo's, don't forget to regenerate the README's

gertjanmaas (approved):

Issue #478, opened 2021-01-19
Header needed for Checks API for Enterprise Server


mcaulifn:
```js
2021-01-19T16:58:27.528Z	784f32d9-df03-522b-86b6-6a58cee71a88	ERROR	RequestError [HttpError]: If you would like to help us test the Checks API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details. 
```

Likely caused by #386

mcaulifn:
Suggest we revert #386 for now.

mcaulifn:
Opened https://github.com/octokit/rest.js/issues/1983

npalm:
@mcaulifn do you know if the issue is caused by the underlying octokit types dependency?

npalm:
Solved by #481

samuelb:
I'm still getting this issue. Tried to build the lambdas by my own yesterday after #481 got merged as well as the just released 0.9.0 lambda zips.

```
2021-01-21T18:33:14.240Z	66df9451-e423-5303-ab9e-ccd22f36bbf2	ERROR	RequestError [HttpError]: If you would like to help us test the Checks API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details.
    at /var/task/index.js:2756:23
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  status: 415,
  headers: {
    'access-control-allow-origin': '*',
    'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, Deprecation, Sunset',
    'alt-svc': 'clear',
    connection: 'close',
    'content-length': '288',
    'content-security-policy': "default-src 'none'",
    'content-type': 'application/json; charset=utf-8',
    date: 'Thu, 21 Jan 2021 18:33:14 GMT',
    'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
    server: 'GitHub.com',
    status: '415 Unsupported Media Type',
    'strict-transport-security': 'max-age=31536000; includeSubdomains',
    via: '1.1 google',
    'x-content-type-options': 'nosniff',
    'x-frame-options': 'deny',
    'x-github-enterprise-version': '2.22.6',
[...]

PR #482, opened 2021-01-22
fix(ghes): Corrects preview option

mcaulifn:
- Adds `mediaType` to correct location
- Tests that `mediaType` is added for GHES


mcaulifn:
Resolves #478

npalm (approved):

npalm:
close #478

PR #496, opened 2021-01-26
feat: Support runner groups

npalm:
Added option to set runner group name via variable `runner_group_name`. If not set the Default group will be used.


modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
```diff
@@ -127,6 +127,18 @@ describe('scaleUp with GHES', () => {
         repoName: undefined,
       });
     });
+
+    it('creates a runner and ensure the group argument is ignored', async () => {
```
gertjanmaas, line 131: The name of the test does not match with the test

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
```diff
@@ -228,6 +252,18 @@ describe('scaleUp with public GH', () => {
...
```
gertjanmaas, line 256: The name of the test does not match with the test

gertjanmaas (approved):

PR #515, opened 2021-02-04
Retrieve installation id automatically if not present (triggered by ordinary webhook)

jonico:
tl;dr Make it possible to reuse one GitHub App for multiple mostly indeoendent environments in combination with ordinary web hooks

### Problem statement 

If multiple mostly independent groups like to share the same GitHub App for their organizations (different installation ids), but like to use isolated environments of the AWS GitHub Runner, they face a problem as a GitHub App can only have one single webhook and secret configured - independent on how many orgs have installed it.

### Solution and problem to solve

 As the different runner installations have different secrets and different endpoints, one solution would be to just install the App in the org but then register an org or repository webhook that listens on the `check_run` event and point it to the address of the corresponding AWS API endpoint. However - ordinary webhook payloads do not contain any installation ids in them, causing the webhook lambda to fail with an NPE.

### Fix in this PR

This PR checks whether the webhook event payload contains an installation id. If not, it will set the installation id to 0 which is a value that is guaranteed to not occur for "real" installation ids. The scale up lambda is then checking for the 0 value and if it finds it, it will lookup the correct installation id for the organization/repo in question.
In case a web hook from a GitHub App is received, the control flow does not change (no additional calls introduced).

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
@@ -1,6 +1,7 @@
 import { mocked } from 'ts-jest/utils';
 import { ActionRequestMessage, scaleUp } from './scale-up';
 import { listRunners, createRunner } from './runners';
+import * as ghAuth from './gh-auth';
 import nock from 'nock';
 
 jest.mock('@octokit/auth-app', () => ({
@@ -11,6 +12,10 @@ const mockOctokit = {
   actions: {
     createRegistrationTokenForOrg: jest.fn(),
     createRegistrationTokenForRepo: jest.fn(),
+  },
+  apps: {
+    getOrgInstallation: jest.fn(),
+    getRepoInstallation: jest.fn(),
   },
 };
 jest.mock('@octokit/rest', () => ({
@@ -25,6 +30,14 @@ const TEST_DATA: ActionRequestMessage = {
   repositoryName: 'hello-world',
   repositoryOwner: 'Codertocat',
   installationId: 2,
+};
+
+const TEST_DATA_WITHOUT_INSTALL_ID: ActionRequestMessage = {
+  id: 3,
+  eventType: 'check_run',
+  repositoryName: 'hello-world',
+  repositoryOwner: 'Codertocat',
+  installationId: 0,
 };
 
 const cleanEnv = process.env;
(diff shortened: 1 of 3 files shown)


npalm (commented):
I will test the change the coming days. In general looks good to me. Please can you have look at the scale runners test and add a test for the case of installationId=0?

jonico:
> will test the change the coming days. In general looks good to me. Please can you have look at the scale runners test and add a test for the case of installationId=0?

@npalm: Thanks for looking into it 😊 . I am not a regular Typescript developer, but I hope the tests I added is what you have been looking for.

npalm:
PR certainly looks fine by me, with this way of obtaining the installation ID the module could potentially also support having default runners available. 

Just wondering about your setup. You have a org or enterpise level webhook that sents events to the one webhook? Next the webhook delivers the message to a SQS queue. Which only can be consumed by one runner. If I understand you correctly you have multiple deployments of ther runners. So how do you route the events on the queue to the right one.

The module is indeed acting on the `check_run` would be nice if there was an event for eacht job created including the runner tags. That would give a much better way to scale in the right way.

Once I have ran a check on the default example I will merge your contribution!

npalm (approved):

jonico:
> Just wondering about your setup. You have a org or enterpise level webhook that sents events to the one webhook? Next the webhook delivers the message to a SQS queue. Which only can be consumed by one runner. If I understand you correctly you have multiple deployments of ther runners. So how do you route the events on the queue to the right one.

So far, I am just experimenting, but in my setup, I have an IssueOps based approach where you create a ticket in a central repo which contains the parameters like org, number runners, environments, images etc and then rolls out a complete installation of your runner setup, including a separate API gateway, message queue etc. Once this step is over, the ticket also contains instruction on how to setup an org wide web hook for the new org pointing to the API gateway that has been provisioned for that very org. The web hook of the GitHub App itself is currently not intercepted at all, as it would only trigger for  one central installation.

image

image


You may ask why not create a separate GitHub App for each installation, but this would require even more manual overhead (so far, the app installation and the web hook creation are still manual steps although the latter could be easily scripted with an API).

> The module is indeed acting on the `check_run` would be nice if there was an event for eacht job created including the runner tags. That would give a much better way to scale in the right way.

Absolutely - in order to have less redundancy and better insights what to actually provision, this would be super cool. I am using your solution along side https://github.com/summerwind/actions-runner-controller in some orgs and already run into situations where AWS runners are spun up even though the jobs have been targeted to my Kubernetes based runners.

jonico:
@npalm: is there any chance to get this fix in an official lambda runner release so that I can abandon my fork for my IssueOps project?

npalm:
@jonico thanks for the reminder, will create tomorrow a release. Still need to automate the release process

npalm:
v0.11.0 just released

jonico:
merci

PR #534, opened 2021-02-10
mcaulifn/lintfix1

mcaulifn:
Resolves lint errors from #529


mcaulifn:
@npalm looks like there was some lingering lint that was caught with the latest dependabot update.

npalm (commented):
@mcaulifn thanks for the quick action±

npalm (approved):

npalm:
Seems that dependabot is breaking the linter rules :(. I will re-create some of the dependabot PR's, and see what will happen

PR #654, opened 2021-03-09
Save lambdas in module's path

dimisjim:
instead of saving them in module's instantiation location


dimisjim:
@npalm https://github.com/philips-labs/terraform-aws-github-runner/pull/654/commits/f6909657b207f8b3f746214ef775aae9a5257083 looks good?

.github/workflows/release.yml
```diff
@@ -79,4 +79,3 @@ jobs:
           cp .release/* .
           yarn
           yarn release --repositoryUrl https://x-access-token:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git
-
```
kuvaldini, line 82: Usually we save last line empty

PR #657, opened 2021-03-09
Allow setting the market_options runners module value

dimisjim:
In order to be able to use on_demand runners

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -342,7 +342,7 @@ variable "runner_additional_security_group_ids" {
 }
 
 variable "market_options" {
-  description = "Market options for the action runner instances."
+  description = "Market options for the action runner instances. Setting the value to `null` let the scaler create on-demand instances instead of spot instanes."
   type        = string
   default     = "spot"
 }
--- a/variables.tf
+++ b/variables.tf
@@ -340,3 +340,9 @@ variable "runner_additional_security_group_ids" {
   type        = list(string)
   default     = []
 }
+
+variable "market_options" {
+  description = "Market options for the action runner instances."
+  type        = string
+  default     = "spot"
+}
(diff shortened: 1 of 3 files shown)


npalm:
Thanks for submitting the PR 

@dimisjim did you check what value needs to be provided or non spot (on demand)?

dimisjim:
@npalm Yeah, according to the docs, the only possible value is "spot"

- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceMarketOptionsRequest.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template

So my assumption is that setting it to `null` would do the trick easily

variables.tf
```diff
@@ -340,3 +340,9 @@ variable "runner_additional_security_group_ids" {
   type        = list(string)
   default     = []
 }
+
+variable "market_options" {
+  description = "Market options for the action runner instances."
```
npalm, line 345:
```suggestion
  description = "Market options for the action runner instances. Setting the value to `null` let the scaler create on-demand instances instead of spot instanes."
```

npalm (changes requested):
@dimisjim Thanks, just tested and it works. If you have time can you add in the REAMDE a small not about switching from spot to on-demand?

On small change suggested

dimisjim:
@npalm https://github.com/philips-labs/terraform-aws-github-runner/pull/657/commits/360acfcd0886d13ea436870a17589197264fb6b6 lgtm :+1:

npalm (approved):

whythecode:
Hi! thanks a lot for this change and for this module! 😄  any plans on when this change will be released? thanks a lot in advance!

npalm:
@whythecode will release the change in the next days. Need to make one more fix in the release automation. Sorry for the delay and thx for the reminder

whythecode:
awesome, thank you for the quick response, @npalm ! And thanks again for the module, it seriously saved my team's sanity.

PR #659, opened 2021-03-09
Improve readme

marcofranssen:
When testing the Ubuntu example I ran into a few small issues. To resolve these I have slightly updated the docs and as well added some core dependencies in this example that allows others to use the ubuntu example without having issues running workflows that utilize commonly used actions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,20 @@ Examples are provided in [the example directory](examples/). Please ensure you h
 
 The module supports two main scenarios for creating runners. On repository level a runner will be dedicated to only one repository, no other repository can use the runner. On organization level you can use the runner(s) for all the repositories within the organization. See https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners for more information. Before starting the deployment you have to choose one option.
 
-GitHub workflows fail immediately if there is no action runner available for your builds. Since this module supports scaling down to zero, builds will fail in case there is no active runner available. We recommend to create an offline runner with matching labels to the configuration. Create this runner manually by following the GitHub instructions for adding a new runner on your local machine. If you stop the process after the step of running the `config.sh` script the runner will remain offline. This offline runner ensures that builds will not fail immediately and stay queued until there is an EC2 runner to pick it up.
+GitHub workflows fail immediately if there is no action runner available for your builds. Since this module supports scaling down to zero, builds will fail in case there is no active runner available. We recommend to create an offline runner with matching labels to the configuration. Create this runner manually by following the [GitHub instructions](https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for adding a new runner on your local machine. If you stop the process after the step of running the `config.sh` script the runner will remain offline. This offline runner ensures that builds will not fail immediately and stay queued until there is an EC2 runner to pick it up.
+
+Another convenient way of deploying this temporary required runner is using following approach. This automates all the manual labor.
+
+```bash
+docker run -it --name my-runner \
+    -e RUNNER_LABELS=selfhosted,Linux,Ubuntu -e RUNNER_NAME=my-repo-docker-runner \
+    -e GITHUB_ACCESS_TOKEN=$GH_PERSONAL_ACCESS_TOKEN \
+    -e RUNNER_REPOSITORY_URL=https://github.com/my-org/my-repo \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    tcardonne/github-runner:ubuntu-20.04
+```
+
+You should stop and remove the container once the runner is registered as the builds would otherwise go to your local Docker container.
 
 The setup consists of running Terraform to create all AWS resources and manually configuring the GitHub App. The Terraform module requires configuration from the GitHub App and the GitHub app requires output from Terraform. Therefore you first create the GitHub App and configure the basics, then run Terraform, and afterwards finalize the configuration of the GitHub App.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -75,32 +75,45 @@ Examples are provided in [the example directory](examples/). Please ensure you h
 - AWS cli (optional)
 - Node and yarn (for lambda development).
 
-The module supports two main scenarios for creating runners. On repository level a runner will be dedicated to only one repository, no other repository can use the runner. On organization level you can use the runner(s) for all the repositories within the organization. See https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners for more information. Before starting the deployment you have to choose one option.
+The module supports two main scenarios for creating runners. On repository level a runner will be dedicated to only one repository, no other repository can use the runner. On organization level you can use the runner(s) for all the repositories within the organization. See <https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners> for more information. Before starting the deployment you have to choose one option.
```
npalm, line 78: Just wondering why this change is only syntax? Both way of linking are supported by github markdown according their spec.

README.md
```diff
@@ -75,32 +75,45 @@ Examples are provided in [the example directory](examples/). Please ensure you h
...
+GitHub workflows fail immediately if there is no action runner available for your builds. Since this module supports scaling down to zero, builds will fail in case there is no active runner available. We recommend to create an offline runner with matching labels to the configuration. Create this runner manually by following the [GitHub instructions](https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for adding a new runner on your local machine. If you stop the process after the step of running the `config.sh` script the runner will remain offline. This offline runner ensures that builds will not fail immediately and stay queued until there is an EC2 runner to pick it up.
+
+Another convenient way of deploying this temporary required runner is using following approach. This automates all the manual labor.
+
+```bash
+docker run -it --name my-runner \
+    -e RUNNER_LABELS=selfhosted,Linux,Ubuntu -e RUNNER_NAME=my-repo-docker-runner \
+    -e GITHUB_ACCESS_TOKEN=$GH_PERSONAL_ACCESS_TOKEN \
+    -e RUNNER_REPOSITORY_URL=https://github.com/my-org/my-repo \
+    -v /var/run/docker.sock:/var/run/docker.sock \
```
npalm, line 89: There is currently work in progress to fully automated this, I would call this stille manual but an alternative. I prefer only add a link here to the readme to this repo, instead of listing the full example. The socket is not required at all for creating an offline runner.

README.md
```diff
@@ -309,7 +322,7 @@ No requirements.
 | ami\_owners | The list of owners used to select the AMI of action runner instances. | `list(string)` | <pre>[<br>  "amazon"<br>]</pre> | no |
 | aws\_region | AWS region. | `string` | n/a | yes |
 | block\_device\_mappings | The EC2 instance block device configuration. Takes the following keys: `device_name`, `delete_on_termination`, `volume_type`, `volume_size`, `encrypted`, `iops` | `map(string)` | `{}` | no |
-| cloudwatch\_config | (optional) Replaces the module default cloudwatch log config. See https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html for details. | `string` | `null` | no |
+| cloudwatch\_config | (optional) Replaces the module default cloudwatch log config. See <https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html> for details. | `string` | `null` | no |
```
npalm, line 325: Pleease do not alter manaul those tables, the input and output table are created via terraform-docs. Via pre-commit hooks. There should be a note on the pre-commit hooks in the contribution guiide.

npalm (changes requested):

README.md
bumbummen99: Why is it a plain link in this paragraph but a pretty one in the paragraph below? I do not see the benefit of showing the full url in a text meant to be read.

README.md
marcofranssen: It is required to do Docker in Docker containers.

README.md
marcofranssen: Also changed this paragraph to have the pretty link.

README.md
marcofranssen: undid this change.

README.md
marcofranssen: Made example by default collapsed.

npalm (approved):

README.md
```diff
@@ -75,32 +74,50 @@ Examples are provided in [the example directory](examples/). Please ensure you h
...
+The module supports two main scenarios for creating runners. On repository level a runner will be dedicated to only one repository, no other repository can use the runner. On organization level you can use the runner(s) for all the repositories within the organization. See [GitHub instructions](https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for more information. Before starting the deployment you have to choose one option.
+
+GitHub workflows fail immediately if there is no action runner available for your builds. Since this module supports scaling down to zero, builds will fail in case there is no active runner available. We recommend to create an offline runner with matching labels to the configuration. Create this runner manually by following the [GitHub instructions](https://help.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) for adding a new runner on your local machine. If you stop the process after the step of running the `config.sh` script the runner will remain offline. This offline runner ensures that builds will not fail immediately and stay queued until there is an EC2 runner to pick it up.
```
ryanbrainard, line 79: What do you think about adding to link to this in Debugging section below? I ran into this because I missed it in the setup. It could help to have to called out.

PR #708, opened 2021-03-26
.ci: Improve dockerfile to include caching

seemethere:
Improves dockerfile / .ci/build.sh script to output lambda zips in a
central location as well as adds buildkit caching to the docker build so
that you don't have to re-download the node_modules of each zip directly
each time you build with .ci/build.sh

This also makes improvements in that if there are only changes in one
directory then the cache will remain and the remaining lambdas will not
be re-built.

I guess an open question as well should be if there should be an additional github actions wofklow added to this?

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

### DIFF
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -1,13 +1,26 @@
-FROM node:12
-
-RUN apt-get update \
-    && apt-get install -y zip \
-    && rm -rf /var/lib/apt/lists/*
-
+#syntax=docker/dockerfile:1.2
+FROM node:12 as build
 WORKDIR /lambda
+RUN apt-get update \
+        && apt-get install -y zip \
+        && rm -rf /var/lib/apt/lists/*
 
-COPY . /lambda
+FROM build as runner-binaries-syncer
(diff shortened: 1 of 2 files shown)


npalm (changes requested):
@seemethere thanks for you contribution! Just checked the chagen. First got all kind of erros. Seems the cause is the javascript heep space. Updated the memory allocated by docker from 2GB to 4GB and all works fine. 

Please can you add a note about the required memory?

seemethere:
> @seemethere thanks for you contribution! Just checked the chagen. First got all kind of erros. Seems the cause is the javascript heep space. Updated the memory allocated by docker from 2GB to 4GB and all works fine.
> 
> Please can you add a note about the required memory?

Done!

npalm (approved):

npalm:
@seemethere Can you da me a small favour and rebase your branch due to yesterdays release?

seemethere:
> @seemethere Can you da me a small favour and rebase your branch due to yesterdays release?

Done!

npalm (approved):

PR #737, opened 2021-04-08
Update Webhook check_run Event Subscription Docs

ryanbrainard:
The way to configure subscribing to the `check_run` events on GitHub has apparently changed. It is now under the "Permissions & Events" section:

image


npalm:
@ryanbrainard thanks for the improvement, actually it is not changed but the description was a bit minimal.

npalm (approved):

PR #785, opened 2021-04-21
fix: change Ubuntu example to log syslog instead of messages

### LINKED ISSUE #783: Ubuntu example should use /var/log/syslog instead of /var/log/messages
Hi all! Thank you for the hard work on this wonderful module!

I just noticed a minor logging issue in the [Ubuntu example](https://github.com/philips-labs/terraform-aws-github-runner/tree/e829e0eb5d7a0b863cf0a50f9527859c19c2b83e/examples/ubuntu). [Logs in `/var/log/messages` are being captured and sent to CloudWatch.](https://github.com/philips-labs/terraform-aws-github-runner/blob/e829e0eb5d7a0b863cf0a50f9527859c19c2b83e/examples/ubuntu/main.tf#L56) As far as I can tell, logs are not or no longer written to `/var/log/messages`, but get stored in `/var/log/syslog` instead. It might be worth changing that over.

Procrat:
Unlike RHEL and CentOS, Ubuntu writes logs to `/var/log/syslog` instead of to `/var/log/messages`. This might have been different in the past.

This fixes #783.

### DIFF
--- a/examples/ubuntu/main.tf
+++ b/examples/ubuntu/main.tf
@@ -51,9 +51,9 @@ module "runners" {
 
   runner_log_files = [
     {
-      "log_group_name" : "messages",
+      "log_group_name" : "syslog",
       "prefix_log_group" : true,
-      "file_path" : "/var/log/messages",
+      "file_path" : "/var/log/syslog",
       "log_stream_name" : "{instance_id}"
     },
     {


npalm (approved):
@Procrat Thank!

Procrat:
Awesome, thanks for the quick turnaround, Niek!

PR #871, opened 2021-06-04
fix(scale): Refactor Runner Type and Owner

mcaulifn:
- Consolidating "is org level runner" logic
- Defining `runnerType` and `runnerOwner`


npalm:
> @npalm Any chance you could review?

@mcaulifn thanks for the reminder, will check asap.

mcaulifn:
@npalm friendly reminder :)

npalm (approved):
@mcaulifn just checked the PR and test the code with the default example. Looks all good to me. I only see some small code formatting issues. But there is not prettier setup for the lambda (and CI). Created a ticket for this #899 So would not block the PR on that.

PR #915, opened 2021-06-21
Add support for white listing of repositories

ravenolf:
## Update
Add the ability to set a white list of repositories that are allowed to use the the runners.

## Reason
When creating the github app, github provides two choices: public or private. If an organisation wants to use the github app between multiple organisations they would have to make it public. In that case it would be more secure to have a filtering on which repositories are allowed to use it.

## Changes

Add an environment variable to the webhook lambda which contains a list of repository full names that are allowed to be processed. The filtering is done on the webhook level at the time of receiving events.

The changes are backward compatible

## Author
Signed-off-by: ravenolf <benahmed@soramitsu.co.jp>

### DIFF
--- a/modules/webhook/lambdas/webhook/src/webhook/handler.ts
+++ b/modules/webhook/lambdas/webhook/src/webhook/handler.ts
@@ -40,6 +40,18 @@ export const handle = async (headers: IncomingHttpHeaders, payload: any): Promis
 
   if (githubEvent === 'check_run') {
     const body = JSON.parse(payload) as CheckRunEvent;
+
+    const repositoryWhiteListEnv = process.env.REPOSITORY_WHITE_LIST as string || "[]";
+    const repositoryWhiteList = JSON.parse(repositoryWhiteListEnv) as Array<string>;
+
+    if (repositoryWhiteList.length > 0) {
+      const repositoryFullName = body.repository.full_name;
+      if (!repositoryWhiteList.includes(repositoryFullName)) {
+        console.error(`Received event from unauthorized repository ${repositoryFullName}`);
+        return 500;
+      }
+    }
+
     let installationId = body.installation?.id;
     if (installationId == null) {
       installationId = 0;
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ This [Terraform](https://www.terraform.io/) module creates the required infrastr
 - [Debugging](#debugging)
 - [Requirements](#requirements)
 - [Providers](#providers)
+- [Modules](#modules)
+- [Resources](#resources)
 - [Inputs](#inputs)
 - [Outputs](#outputs)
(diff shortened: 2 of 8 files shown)


npalm:
Thanks for creating the PR. I will check the PR early next week

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -40,6 +40,18 @@ export const handle = async (headers: IncomingHttpHeaders, payload: any): Promis
 
   if (githubEvent === 'check_run') {
     const body = JSON.parse(payload) as CheckRunEvent;
+
+    const repositoryWhiteListEnv = process.env.REPOSITORY_WHITE_LIST as string || "[]";
+    const repositoryWhiteList = JSON.parse(repositoryWhiteListEnv) as Array<string>;
+
+    if (repositoryWhiteList.length > 0) {
```
npalm, line 47: Can you add a unit test for this case?

npalm (changes requested):
@ravenolf thanks for proposing the change, in general the change looks good to me. Can you add a unit test to check the case.

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -40,6 +40,18 @@ export const handle = async (headers: IncomingHttpHeaders, payload: any): Promis
 
   if (githubEvent === 'check_run') {
     const body = JSON.parse(payload) as CheckRunEvent;
+
+    const repositoryWhiteListEnv = process.env.REPOSITORY_WHITE_LIST as string || "[]";
+    const repositoryWhiteList = JSON.parse(repositoryWhiteListEnv) as Array<string>;
```
npalm, line 45: Just a thought, may it is better to specify the repos's as `owner/repo` this ensure the nameas are always unique. Potential the runner can also be setup on org level. What do you think?

ravenolf: In the code I'm using the repository.full_name which looks like `owner/repo-name`

modules/webhook/lambdas/webhook/src/webhook/handler.ts
ravenolf: Sure, done!

modules/webhook/lambdas/webhook/src/webhook/handler.ts
npalm: yep

npalm (approved):
@ravenolf thanks!

dimisjim:
what's the default behavior after introducing this variable?

if one doesn't set this var, does it mean that the github app is not gonna work for any repo?

npalm:
Default is: no white list is used. Only once set

Issue #801, opened 2021-04-27
Setup Github App: Organization - Administration permission does not seem to be required


davidshin:
First off, thanks for this amazing terraform module! 

I've experimented and found that when setting up the Github App according to the [readme](https://github.com/philips-labs/terraform-aws-github-runner#setup-github-app-part-1), step 7,  the `Organization - Administration: Read & Write` permission is not required for everything to work properly.  Seems like the `Organization - Self-hosted runners: Read & Write` permission is sufficient.  

Perhaps the readme should be updated to reflect this?  Or maybe I'm missing something?

npalm:
Thanks for reporting. We need to check. But could be that the section was added over the last year by github.

kuvaldini:
Refer to this documentation page https://docs.github.com/en/rest/reference/permissions-required-for-github-apps#permission-on-self-hosted-runners

npalm:
Just verified and admin permissions are not longer required. Checked a org level setup.

davidshin:
Thanks!

PR #988, opened 2021-07-14
chore(doc): Remove org level admin permission #801

npalm:
Org level admin permissions for the app are not required any more, removed from docs.


gertjanmaas (approved):

npalm:
Close #801

Issue #980, opened 2021-07-13
Upgrade from to v0.13.0 to v0.15.0 leads to error in scale-up


elft3r:
Hey,
I wanted to upgrade the module from v.0.13.0 to v0.15.0. However, this leads to the following error message in the `scale-up` lambda:
```
2021-07-13T08:32:37.700Z	undefined	ERROR	Uncaught Exception 	
{
    "errorType": "TypeError",
    "errorMessage": "Cannot set property 'scaleUp' of undefined",
    "stack": [
        "TypeError: Cannot set property 'scaleUp' of undefined",
        "    at Module.30873 (/var/task/index.js:61201:24)",
        "    at __nccwpck_require__ (/var/task/index.js:66346:43)",
        "    at /var/task/index.js:66431:37",
        "    at Object.<anonymous> (/var/task/index.js:66434:12)",
        "    at Module._compile (internal/modules/cjs/loader.js:999:30)",
        "    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)",
        "    at Module.load (internal/modules/cjs/loader.js:863:32)",
        "    at Function.Module._load (internal/modules/cjs/loader.js:708:14)",
        "    at Module.require (internal/modules/cjs/loader.js:887:19)",
        "    at require (internal/modules/cjs/helpers.js:74:18)"
    ]
}
```

Can anyone help me to understand what I am missing? Also, let me know if you need additional information to understand the problem I'm facing better.

This is the code I used to apply the module:
```
module "github_runners" {
  source = "git::git@github.com:philips-labs/terraform-aws-github-runner.git?ref=v0.15.0"

  aws_region = var.aws_region
  vpc_id     = data.terraform_remote_state.vpc_mgmt.outputs.vpc_id
  subnet_ids = data.terraform_remote_state.vpc_mgmt.outputs.private_subnet_ids

  environment = var.environment

  github_app = {
    key_base64     = var.github_app_key_base64
    id             = var.github_app_id
    client_id      = var.github_app_client_id
    client_secret  = var.github_app_client_secret
    webhook_secret = random_password.random.result
  }

  webhook_lambda_zip                = "${path.module}/lambdas-downloads/webhook.zip"
  runner_binaries_syncer_lambda_zip = "${path.module}/lambdas-downloads/runner-binaries-syncer.zip"
  runners_lambda_zip                = "${path.module}/lambdas-downloads/runners.zip"
  enable_organization_runners       = true
[...]

PR #1005, opened 2021-07-23
bugfix: change module exports and upgrade vercel to latest release

### LINKED ISSUE #982: vercel/ncc cannot upgrade to version above 0.27
Upgrading ncc/vercel to version 0.28.x will casue an issue. most likely related to some export definitons in the lambda.  See also #980

PatrickMennen:
Upgrades @vercel/ncc to its latest release and updated the code to compile as well.

this should fix #982 and possibly #980


npalm (approved):
Thanks for the help

checked all the 3 lambda's on a deployment, all good!

PR #1065, opened 2021-08-10
feat(scale-down): Update Owner Logic

### LINKED ISSUE #1021: Scale Down Only Looks at Environment
The scale down lambda only looks at environment when making a decision. For deployments supporting multiple orgs, the lambda will leave the latest up, regardless of org.

The scale down lambda should fetch all runners and apply the idle config per org.

mcaulifn:
- Updates scale down logic to look at runner owner when evaluating which runners to terminate
- Changes tags on instance to `tag = type` and `tag = owner` instead of `tag = type` and  `value = owner`
- Isolates scale down tests

NOTE: Runners that do not contain the new tags will be terminated when the scale down lambda runs.

Resolves #1021


mcaulifn:
Would it be worthwhile to add a function to deal with runners that got pulled in with the environment tag but then do not have any other tags?

mcaulifn:
@npalm I am marking this ready for review. I am working on some very small code changes to deal with orphaned runners. That will solve the bug I identified and also address your concerns. I will update the PR description with that callout.

mcaulifn:
@npalm I created https://github.com/mcaulifn/terraform-aws-github-runner/pull/2 as a way to deal with legacy runners. If the approach is good, I'll merge.

gertjanmaas:
@mcaulifn did a review on https://github.com/mcaulifn/terraform-aws-github-runner/pull/2 only some small comments

If that's merged in here, it looks good to me. 

Nice work and thanks for improving this project!

mcaulifn:
@gertjanmaas @npalm I think this should be ready. I will be testing it out in our dev environment next week.

npalm:
@mcaulifn will check the PR early nexrt week as well. @gertjanmaas thanks for checking

modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -1,10 +1,19 @@
 import { EC2, SSM } from 'aws-sdk';
 
+export interface RunnerList {
+  instanceId: string;
+  launchTime?: Date;
+  owner?: string;
+  type?: string;
+  Repo?: string;
```
npalm, line 8: For consistency can you change Repo and Org to repo an org (lowercase / camelCase)? Similar to the other interface properties

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -100,44 +73,84 @@ function runnerMinimumTimeExceeded(runner: RunnerInfo, minimumRunningTimeInMinut
...
+    const ec2RunnersFiltered = ec2Runners.filter((runner) => runner.owner === ownerTag);
+    for (const ec2Runner of ec2RunnersFiltered) {
+      if (runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes)) {
+        const ghRunners = await listGitHubRunners(ec2Runner);
+        const ghRunner = ghRunners.find((runner) => runner.name === ec2Runner.instanceId);
+        if (ghRunner) {
+          if (idleCounter > 0) {
+            idleCounter--;
+            console.debug(`Runner '${ec2Runner.instanceId}' will kept idle.`);
+            ec2Runners.splice(getIndex(ec2Runners, ec2Runner), 1);
```
npalm, line 122: Not sure if it is safe to update the list during iteration. The splice operation is also duplicated 3 times. Should not be necessary

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -100,44 +73,84 @@ function runnerMinimumTimeExceeded(runner: RunnerInfo, minimumRunningTimeInMinut
...
+          }
+        } else {
+          console.debug(`Runner '${ec2Runner.instanceId}' is orphaned and will be removed.`);
+          terminateOrphan(ec2Runner.instanceId);
+          ec2Runners.splice(getIndex(ec2Runners, ec2Runner), 1);
+        }
+      }
+    }
+  }
+  if (!(ec2Runners.length === 0)) {
```
npalm, line 135: Those runners are not orphan, since the remaining list are the runners that are not exeeding the minimum booting time, see line 115. We have used this minimum time to avoid runners will be killed before they are ready and registered.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -147,46 +160,33 @@ export async function scaleDown(): Promise<void> {
...
-    console.debug(`No active runners found for environment: '${environment}'`);
-    return;
-  }
+function filterLegacyRunners(ec2runners: RunnerList[]): RunnerList[] {
+  return ec2runners.filter((ec2Runner) => ec2Runner.Repo || ec2Runner.Org) as RunnerList[];
+}
 
-  const createGitHubClientForRunner = createGitHubClientForRunnerFactory();
-  const listGithubRunners = listGithubRunnersFactory();
+function filterNewRunners(ec2runners: RunnerList[]): RunnerInfo[] {
```
npalm, line 169: Would call this function simply filterRunners

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -147,46 +160,33 @@ export async function scaleDown(): Promise<void> {
     if (a.launchTime > b.launchTime) return -1;
     return 0;
   });
+}
```
npalm, line 164:
```suggestion

/**
 * We are moving to a new strategy to find and remove runners, this function will ensure
 * during migration runners tagged in the old way are removed.
 */
```

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -147,46 +160,33 @@ export async function scaleDown(): Promise<void> {
...
-    }
+  if (ec2Runners.length === 0) {
+    console.debug(`No active runners found for environment: '${environment}'`);
+    return;
   }
+  const legacyRunners = filterLegacyRunners(ec2Runners);
+  const newRunners = filterNewRunners(ec2Runners);
+
+  await evaluateAndRemoveRunners(newRunners, scaleDownConfigs, minimumRunningTimeInMinutes);
+  legacyRunners.forEach((runner) => {
```
npalm, line 189: This will force kill any old tagged runner, and also stop any build running. It also will not de-register the runner from github. Can you add a check the runner is not active and de-register ti as well?

npalm (changes requested):

modules/runners/lambdas/runners/src/scale-runners/runners.ts
mcaulifn: Fixed.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: I was able to move the splice operation to encompass all three instances.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: Good catch!

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: Done

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: This should be dealing with the correct set of orphaned runners now.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: After re-working some logic, I no longer needed to update the original list.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
mcaulifn: The legacy list is now passed through the same logic as the new list.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -100,44 +73,71 @@ function runnerMinimumTimeExceeded(runner: RunnerInfo, minimumRunningTimeInMinut
...
+          if (idleCounter > 0) {
+            idleCounter--;
+            console.debug(`Runner '${ec2Runner.instanceId}' will kept idle.`);
+          } else {
+            await removeRunner(ec2Runner, ghRunner.id);
+          }
+        }
+      } else {
+        console.debug(`Runner '${ec2Runner.instanceId}' is orphaned and will be removed.`);
+        terminateOrphan(ec2Runner.instanceId);
```
npalm, line 124: If the diffs shows it correctly, it seems orphan runners are removed if not `(runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes))`. This will remove all runners that are NOT exceeding the minimal time. So potential still executing the user_data script. Orphan runners are all runners that not registred in github but running in aws as ec instance longer then the minimal time. 

So removal should be done if NOT `ghRunner` and minimal time exeeded. So suggest you swap if clauses on line 113 and 114. Next move the terminateOrphan to the else of the most inner if

mcaulifn: I don't think the diff is rendering correctly. I copied that section below and added two notes.

```typescript
      if (ghRunner) {
        if (runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes)) {
          if (idleCounter > 0) {
            idleCounter--;
            console.debug(`Runner '${ec2Runner.instanceId}' will kept idle.`);
          } else {  //If the idle counter is not greater than 0
            await removeRunner(ec2Runner, ghRunner.id);
          }
        // implied else - If the runner min time not exceeded, continue
        }
      } else {
        console.debug(`Runner '${ec2Runner.instanceId}' is orphaned and will be removed.`);
        terminateOrphan(ec2Runner.instanceId);
      }
```

mcaulifn:
@npalm I deployed locally and I see a legacy runner and new runners being handled correctly.

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
npalm: I still think the code is not correct see below the part int he PR, second part my proposal
```
      if (ghRunner) { // runner registred in github
        if (runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes)) { // runner is not running minimum time 
          if (idleCounter > 0) {
            idleCounter--;
            console.debug(`Runner '${ec2Runner.instanceId}' will kept idle.`);
          } else {  // enough idle runners so remove
            await removeRunner(ec2Runner, ghRunner.id);
          }
        // implied else - If the runner min time not exceeded, continue
        }
      } else { // runner not registred in gihub, potential not exceedining minimal time, current code (dev) has check.
        console.debug(`Runner '${ec2Runner.instanceId}' is orphaned and will be removed.`);
        terminateOrphan(ec2Runner.instanceId);
      }
```

Proposed change, we swam the check for minimal time and ghRunner, so if the runner is not a ghRunner we can safely remove the runner still it execeeded the minimal required time to boot. 
```
    if (runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes)) { 
      if (ghRunnr) { 
        if (idleCounter > 0) {
            idleCounter--;
            console.debug(`Runner '${ec2Runner.instanceId}' will kept idle.`);
        } else {  // enough idle runners so remove
            await removeRunner(ec2Runner, ghRunner.id);
        }
      } else { // runner not registered
          console.debug(`Runner '${ec2Runner.instanceId}' is orphaned and will be removed.`);
          terminateOrphan(ec2Runner.instanceId);
        }
      }
    }
```

mcaulifn: > if (runnerMinimumTimeExceeded(ec2Runner, minimumRunningTimeInMinutes)) {  // runner is **not** running minimum time 
Is the "not" intended here? I removed the `!` so it is now checking if the minimum time has been met.

mcaulifn: I realized I missed the comment "So potential still executing the user_data script". That shouldn't be the case as the default is 5 minutes and should be enough time to register. It might be worth calling out that this value should be the the user data execution time plus an additional allowance. In our case, with a 60 minute minimum run time, we don't want an orphaned runner hanging around for that long.

npalm: Actually the minimum time for runners had initial 2 functions.
- ensure the runner gets enough time to boot
- the minimal time we keep the runner for re-use. Over time we would lik to move to a aystem that be default uses the runner once. But we are not here yet. 

In the past we had some times the problem a runner is not booting correct at all. This caused ec2 instances that are running but not registered to github. We called them orphan. But to avoid we clean up to early. Since boot time can be involved we only clear them after the mininum time. You are right we could differ between a minimum time and a minimum time for the user data to get executed.

mcaulifn: I updated the variable description to call out the necessity for this value to allow the runner enough time to boot and register.

mcaulifn:
@gertjanmaas Anything else on this one?

gertjanmaas (approved):

gertjanmaas:
Looks good to me, tested the default example and works as expected. @npalm do you want to take a look at it?

npalm:
@gertjanmaas thanks for checking, please feel free to merge

Issue #899, opened 2021-06-15
Add prettier (format check) for lambda to CI


npalm:
Currently code formatting of the lambd'as is not checked. Add the format check to the CI for every lambda. 

- Add prettier
- Run prettier via `yarn run format-check` in CI

Patil2099:
I would love to work on this issue.

npalm:
Feel free to propose a PR, let me know if you need some direction

aadrijnberg:
I will pick this one up

PR #1080, opened 2021-08-12
feat: add format checking for lambdas in CI (#899)

aadrijnberg:
I added a question in the ticket, if the tests should also be formatted and format-checked. If so, please indicate, and I will update this pull request.


gertjanmaas (approved):
Looks good to me!

Issue #904, opened 2021-06-17
Upgrade auth-app in lambda runners breaks the tests


npalm:
Upgrade in module / lambda runners for auth-app to version 3.5 is failing the test.

PR #1081, opened 2021-08-12
chore: Bump oktokit/auth-app (#904)

aadrijnberg:
The newer version of the library requires different input data for yyy in the createAppAuth(xxx)(yyy) function, and also returns a different data structure for the 'app' and 'installation' scenarios.

- Set dependency to version 3.6.0
- Fix some (misleading) test descriptions


npalm (approved):
Thanks, LGTM. Verified end-to-end

Issue #903, opened 2021-06-17
Replace request in download binary lambda


npalm:
Lambda download binary is using depcrecated package `request`. Shoul be replaced by an alternative.

PR #1082, opened 2021-08-13
chore: replace 'request' dependency by 'node-fetch' (#903)

aadrijnberg:
Checked [alternatives](https://nodesource.com/blog/express-going-into-maintenance-mode) and decided that node-fetch is suitable for our purpose; lowest footprint, no dependencies, supports streaming, heavily used, low number of bugs.

Implementation note: await the success of **_both_** promises before reporting that it was successfully uploaded to S3 (previous implementation did not do that).


npalm (commented):
Thanks, LGTM

npalm (approved):

Issue #748, opened 2021-04-12
Security: Make optional the hard-coded Security Group `runner_sg`?


boldandbusted:
Howdy. I have a restriction in an environment that prevents wide open egress AWS Security Group rules from being applied (automation actually reverses what this module sets). Each time I apply this rule the automation undoes this, so we get repeating change-in-place loops from TF apply to apply. I'd rather manage this via other means - like using the default SG for the VPC. Is this possible? Or am I missing a path to this functionality that already is available via the module input variables?

https://github.com/philips-labs/terraform-aws-github-runner/blob/65aab803262e0616a8895d2fdbc0bc112e2572f9/modules/runners/main.tf#L117

What would be great would be a method to either:
* Customize this SG
* Disable applying this SG (and then rely on the VPC default SG)

Note that I'm already using `runner_additional_security_group_ids` input variable to specify the default VPC SG, but I don't think I can use that to undo what this code is doing?

Thanks in advance! :) Happy to provide more info.

npalm:
The security group should be better configurable, what do you think of making the egress fully configureable with the current defaults. In that case we avoid breaking changes. I have done the same for another module, check 

https://github.com/npalm/terraform-aws-gitlab-runner/blob/8b15241b34c87d7b21c7a14fddf7d0b84f750f1b/security_groups.tf#L5

Feel free to suggest the change in a PR

new23d:
@boldandbusted does the automation in your environment not like the destination range of IP addresses (that is, 0.0.0.0/0) or the list of ports (0 to 65535)?

I'm wondering if keeping 0.0.0.0/0 but restricting the ports will resolve this issue.

boldandbusted:
> @boldandbusted does the automation in your environment not like the destination range of IP addresses (that is, 0.0.0.0/0) or the list of ports (0 to 65535)?
> 
> I'm wondering if keeping 0.0.0.0/0 but restricting the ports will resolve this issue.

Sorry for the late reply. It's hard for me to tell because there are other groups that have automation that slams closed security groups that open ports and IPs to the public Internet. It would probably be useful to have what @npalm suggests for a PR, if that gives us flexibility to specify address and ports, together.

new23d:
FWIW, we analysed the egress traffic from the runner through a firewall. As of today, it was TLS port 443 (so HTTPS I guess) connections to:

```
logs.<region>.amazonaws.com
ssm.<region>.amazonaws.com

[...]

PR #1112, opened 2021-08-18
current value turned to default under new variable, for issue #748

new23d:
Proposed code changes to accommodate custom default egress rules, as discussed under issue #748.

Inspired by the GitLab runners code as mentioned under that issue.

Early feedback on these changes is most welcome!

Also, terraform-docs on my end seems to change a lot of the formatting in the README.md files. Am I missing or not doing something obvious? It adds a lot of HTML tags in each line in the tables, hyperlinking to the resources I guess. I'd have thought such formatting should already be present in the current files.

### DIFF
--- a/modules/runners/variables.tf
+++ b/modules/runners/variables.tf
@@ -346,4 +346,15 @@ variable "egress_rules" {
     to_port          = number
     description      = string
   }))
+  default = [{
+    cidr_blocks      = ["0.0.0.0/0"]
+    ipv6_cidr_blocks = ["::/0"]
+    prefix_list_ids  = null
+    from_port        = 0
+    protocol         = "-1"
+    security_groups  = null
+    self             = null
+    to_port          = 0
+    description      = null
+  }]
 }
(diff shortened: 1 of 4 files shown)


modules/runners/variables.tf
```diff
@@ -332,3 +332,18 @@ variable "kms_key_arn" {
...
+    cidr_blocks      = list(string)
+    ipv6_cidr_blocks = list(string)
+    prefix_list_ids  = list(string)
+    from_port        = number
+    protocol         = string
+    security_groups  = list(string)
+    self             = bool
+    to_port          = number
+    description      = string
+  }))
```
npalm, line 348: Can you add the same defaults here as the top level variables.tf? This will help the consumers of the submodule.

npalm (changes requested):
Small change requested.

modules/runners/variables.tf
new23d: Done.

npalm (approved):
@new23d thanks! LGTM

PR #1135, opened 2021-08-27
fix(logging): Additional Logging

mcaulifn:
- Fixes output of events
- Adds additional debug logging at possible failure point during scale up


mcaulifn:
@gertjanmaas Could you take a look? I've seen some scale up runs that stop short of launching an instance but with no explanation. I think it is this step but would like some debug logging to find out.

gertjanmaas (approved):

PR #1154, opened 2021-09-02
Changes `instance_types` from a Set to a List, so instance order preference is preserved

### LINKED ISSUE #1152: `instance_types` are made into a Set, which breaks the ordering
I'm unsure exactly how templates are picked and in what order in this module, but when specifying a list of instances they're converted into a Set, which alphabetises them and means that the order is broken. Eg, if passed an array: 

```
instance_types = ["z1d.large", "r5a.large", "r5.large", "r5ad.large", "r5n.large", "t3.xlarge", "m5a.xlarge", "t2.xlarge", "m5.xlarge", "m4.xlarge"]
```

It will be converted into a Set, which becomes:

```
instance_types = [ "m4.xlarge", "m5.xlarge", "m5a.xlarge", "r5.large", "r5a.large", "r5ad.large", "r5n.large", "t2.xlarge", "t3.xlarge", "z1d.large" ]
```

This means that the spot instances that are spawning are preferring m4.xlarge rather than the z1d.large that I'd prefer.

Makeshift:
Fixes #1152 

Note that pre-commit updated the readme's and appears to have deleted some sections. Am I running an out of date `terraform_docs` or is there a misconfiguration somewhere?

Cheers!

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -38,7 +38,7 @@ data "aws_ami" "runner" {
 }
 
 locals {
-  runner_launch_template_prefix = "${var.environment}-action-runner-"
+  runner_launch_template_prefix = "${var.environment}-action-runner"
 }
 
 resource "aws_launch_template" "runner" {
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -68,7 +68,7 @@ resource "aws_launch_template" "runner" {
   }
 
   image_id      = data.aws_ami.runner.id
-  instance_type = each.value
+  instance_type = local.instance_types[count.index]
   key_name      = var.key_name
 
   vpc_security_group_ids = compact(concat(
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -18,7 +18,7 @@ locals {
   userdata_arm_patch             = "${path.module}/templates/arm-runner-patch.tpl"
   userdata_install_config_runner = "${path.module}/templates/install-config-runner.sh"
 
-  instance_types = var.instance_types == null ? [var.instance_type] : var.instance_types
+  instance_types = distinct(var.instance_types == null ? [var.instance_type] : var.instance_types)
 
   kms_key_arn = var.kms_key_arn != null ? var.kms_key_arn : ""
 }
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -37,10 +37,14 @@ data "aws_ami" "runner" {
   owners = var.ami_owners
 }
 
+locals {
+  runner_launch_template_prefix = "${var.environment}-action-runner-"
+}
+
 resource "aws_launch_template" "runner" {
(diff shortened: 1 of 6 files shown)


modules/runners/main.tf
```diff
@@ -38,9 +38,9 @@ data "aws_ami" "runner" {
 }
 
 resource "aws_launch_template" "runner" {
-  for_each = local.instance_types
+  count = length(local.instance_types)
```
mcaulifn, line 41: Why the change from loop to count?

Makeshift: You can't `for_each` on an array without converting it into a Set, which breaks the point of this PR ;).

I could have converted it into a map, but this seemed cleaner.

```
The given "for_each" argument value is unsuitable: the "for_each" argument must be a map, or set of strings, and you have provided a value of type tuple.
```

lorengordon: Won't this introduce a different type of problem, where if an item is added or removed from the list, the indexes will change and terraform will destroy and recreate them all?

lorengordon:
It looks like the template order is picked up here:

https://github.com/philips-labs/terraform-aws-github-runner/blob/e72227bd8c5d76f14c42119e17eae5762c247f85/modules/runners/lambdas/runners/src/scale-runners/scale-up.ts#L113

Which is set here:

https://github.com/philips-labs/terraform-aws-github-runner/blob/e72227bd8c5d76f14c42119e17eae5762c247f85/modules/runners/scale-up.tf#L23

So I'd suggest an alternate fix, modifying the variable to be a list, but leaving this as a `for_each` using `toset(local.instance_types)`, and adjusting the way the lambda env is set:

```
      LAUNCH_TEMPLATE_NAME                    = join(",", [for instance_type in local.instance_types : aws_launch_template.runner[instance_type]])
```

That should preserve the order of the list, and also establish the dependency on the resource.

npalm:
Will check the PR this week, sorry for the bit late response.

modules/runners/main.tf
npalm: No the instance type list is only configuration to start instances, so it will never impact current running onces.

npalm (approved):
LGTM

modules/runners/main.tf
lorengordon: That's not quite what I was trying to get at though, I don't think... This is for the `aws_launch_template` resource, not the launched instances... If the user modifies the list of input instances types for an already applied state, and runs the apply again, won't terraform destroy and recreate all the launch templates? Or no because the `instance_type` field can be modified without recreating the resource?

lorengordon: Ahh, however this is mapped to the `name` field also, and that _does_ require replacement, so yes, this will destroy and recreate the launch template when the list changes.

lorengordon: Changing from for_each to count is also a backwards-incompatible change for any existing state.

PR #1164, opened 2021-09-07
fix(scale-down): Clearing cache between runs

mcaulifn:
The class is not being re-created between runs. Clearing each map appears to be necessary to avoid stale data.


mcaulifn:
@npalm @gertjanmaas This is to resolve a bug identified with the new cache class. It has been tested.

gertjanmaas (approved):

PR #1202, opened 2021-09-29
fix(runner): use dynamic block to ignore null market opts

taharah:
Currently, if `market_options` is set to `null`, then Terraform will continuously detect a diff for the launch template configuration and recreate it. This change will use a `dynamic` block to exclude the  `instance_market_options` block entirely if `market_options` is set to `null`.

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>


npalm (approved):
@taharah thanks! Tested and reviewed. All Good!

PR #1203, opened 2021-09-29
Bump nodejs to Node 14

marcofranssen:
This PR bumps the nodejs runtime for lambdas to Node 14. Therefore we also require a minimum version for the aws terraform provider.


npalm:
The module is depend on the node runtime support by AWS. Today AWS supports 10, 12 and 14. (https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html).

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
 # Changelog
```
npalm, line 2: Please can you revert this change?

npalm (changes requested):

examples/default/versions.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = ">= 3.20"
+      version = ">= 3.27"
```
npalm, line 5: Changing the required versions also requires an update in the lock file. Same for other examples. See file `.terraform.lock.hcl`

npalm (changes requested):

npalm:
Tested a deployment, works all fine. @marcofranssen please can you check the review comments. Thanks for your contribution!

npalm:
@marcofranssen did you had time to have a look on the comments?

PR #1222, opened 2021-09-30
fix: added more detailed logging for scaling up and down

npalm:
Improved logging for scaling up and down behavior.

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
@@ -117,6 +117,7 @@ async function evaluateAndRemoveRunners(
 
   for (const ownerTag of ownerTags) {
     const ec2RunnersFiltered = ec2Runners.filter((runner) => runner.owner === ownerTag);
+    console.debug(`Found: '${ec2RunnersFiltered.length}' active GitHub runners with owner tag: '${ownerTag}'`);
     for (const ec2Runner of ec2RunnersFiltered) {
       const ghRunners = await listGitHubRunners(ec2Runner);
       const ghRunner = ghRunners.find((runner) => runner.name === ec2Runner.instanceId);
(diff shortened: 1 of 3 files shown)


modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -190,15 +191,24 @@ export async function scaleDown(): Promise<void> {
...
+  console.info(`Found: '${activeEc2RunnersCount}' active GitHub EC2 runner instances before clean-up.`);
 
-  if (ec2Runners.length === 0) {
+  if (activeEc2RunnersCount === 0) {
     console.debug(`No active runners found for environment: '${environment}'`);
     return;
   }
   const legacyRunners = filterLegacyRunners(ec2Runners);
-  console.log(JSON.stringify(legacyRunners));
+  console.debug(JSON.stringify(legacyRunners));
```
gertjanmaas, line 202: is this debug logging intentional?

npalm: I only changed it to debug. I assume is wad added with the intentio to see if still legacy runners needs to be termimnated by @mcaulifn

modules/runners/lambdas/runners/src/scale-runners/scale-down.ts
```diff
@@ -190,15 +191,24 @@ export async function scaleDown(): Promise<void> {
...
   const runners = filterRunners(ec2Runners);
 
   await evaluateAndRemoveRunners(runners, scaleDownConfigs);
   await evaluateAndRemoveRunners(legacyRunners, scaleDownConfigs);
+
+  const activeEc2RunnersCountAfter = (await listAndSortRunners(environment)).length;
+  console.info(
+    `Removed: '${
```
npalm, line 210: I just realised that the stetement is not correct. Since the number of active runner can be interfered by the scale up lambda

gertjanmaas (approved):

PR #1223, opened 2021-09-30
fix: remove unused app client id and secret since SSH key is used for authorization

npalm:


npalm:
@mcaulifn can you do me a favour and have a look / do a review for this PR.

mcaulifn:
Looking at the docs, all examples include both vars. Neither is listed as **Required** [here](https://github.com/octokit/auth-app.js/#createappauthoptions-or-new-octokit-auth-). Is there some other documentation indicating they are not used?

npalm:
> Looking at the docs, all examples include both vars. Neither is listed as **Required** [here](https://github.com/octokit/auth-app.js/#createappauthoptions-or-new-octokit-auth-). Is there some other documentation indicating they are not used?

For the lambda we use the app installation, as far as I understood this requires the app id and SSH private key only. The client is for an oauth auth, which we don't use.

mcaulifn:
[This](https://github.com/octokit/auth-app.js/#authenticate-as-installation) was what I was looking at originally. But then the second example under [here](https://github.com/octokit/auth-app.js/#authenticate-as-installation) makes it seem like it is not required. Lovely docs 🙄 

I'm testing another PR right now but will queue this one up after that.

npalm:
> [This](https://github.com/octokit/auth-app.js/#authenticate-as-installation) was what I was looking at originally. But then the second example under [here](https://github.com/octokit/auth-app.js/#authenticate-as-installation) makes it seem like it is not required. Lovely docs 🙄
> 
> I'm testing another PR right now but will queue this one up after that.

I have also deployed the lambda with an invalid client secret, and guess what. It just worked. Using a wrong SSH private key generates errors.

mcaulifn:
I added bogus values for client secret and it worked. Seems to be good for GHES 3.1.

mcaulifn (approved):

JeroenKnoops (approved):

PR #1227, opened 2021-10-04
Add ingress rules for runner SG

sshishov:
We are using bastion in our setup to access all instances. SSM is not convenient for us.
As now we are using `key_name` to have the SSH key propagated, we would also like to have SSH access to the node without additional manual work.

To achieve this we should be able to add SSH access from special SG, instance_ip, or subnet. This PR is adding the ability to specify `ingress` for Runner SG.

P.S. Currently you can add any rules to `ingress` manually and Terraform will not show changes. But it is not good as the infrastructure is not in the code now...

Another solution would be to add SG id to the output and let a user add SG rules manually and attach them to the group. But in this case, we will have 2 different approaches for `egress` and `ingress`, which is not good, IMHO.

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -133,6 +133,23 @@ resource "aws_security_group" "runner_sg" {
 
   vpc_id = var.vpc_id
 
+  dynamic "ingress" {
+    for_each = var.ingress_rules
+    iterator = each
+
+    content {
+      cidr_blocks      = each.value.cidr_blocks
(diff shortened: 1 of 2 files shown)


ScottGuymer:
Would it not be possible to have the module output the right ids and then add this type of functionality outside of this module? I'm not sure if this is possible with the resources her.

I'm thinking a more "open for extension" kind of thing rather than to keep having to add more and more options to this module itself.

ScottGuymer:
This looks like a stale PR. Closing but feel free to re-open if you would like to continue to get this merged.

PR #1228, opened 2021-10-04
JSON Logging for Lambdas

mcaulifn:
Using the `tslog` library, this PR adds support for JSON formatting of logs for Lambdas. This assists with readability and ingesting logs in to a tool like Splunk. The default value is `pretty`, keep the module backward compatible with current logging config.


npalm:
@mcaulifn terraform checks are failing the module is still on 0.13.x I planning to bump tf versions asap. The funstion you are using requiring terraform 0.14.

mcaulifn:
@npalm I need to resolve the default formatting as well. Thanks for checking!

npalm:
I also need a bit more time to look on the full change set. Can we now also configure the logger to only log for example info level?

mcaulifn:
I don't have that configurable right now but I could add that using environment variables if you think it would be useful.

npalm:
You can just update the version in the terraform workflow that verify the examples. Would indeed start with the latest 0.14

mcaulifn:
@npalm Changed per request. 

There is state file compatibility across `0.14`, `0.15`, and `1.0`. I would suggest that these versions should be kept up-to-date as best practice. There should be no obstacles to users of this module using a newer version with the stability announcement of `0.14` https://www.hashicorp.com/blog/announcing-hashicorp-terraform-0-14-general-availability and confirmed in `0.15` https://www.hashicorp.com/blog/announcing-hashicorp-terraform-0-15-general-availability.

PatrickMennen (dismissed):

npalm:
> I wasn't familiar with any library. I happen to find `tslog` with a quick search "typescript json logger". I can look at `log4js` or `winston` if you think they will be a better fit. I have no preference as of right now.
> 
> WRT using the `logger` object or `console`: I was having issues passing around the configuration. Creating a new object in each file would not carry over the settings from the root logger like Python does. Trying to import the object led to testing issues that I had not yet worked out. I didn't want to delay this PR so chose to submit this and continue work on mocking with the next PR.

I also prefer to use the logger for all call. So please update all the calls. Currently you named the variabl rootLogger. Also common practice to name the logger just `logger` (the variable).

npalm (changes requested):
@mcaulifn thanks great job! Just tested and checked the logs. LGTM. Unfortunately some other merges creates conflicts. Can you rebase once again please?

mcaulifn:
@npalm Resolved conflicts, should be good?

npalm (approved):

npalm:
> @npalm Resolved conflicts, should be good?

Running out of time for today, will go over the PR once again tomrrow. It is a big, and very welcome improvment. So therefore taking a bit more time. Once again thanks!

npalm (approved):

Issue #1218, opened 2021-09-30
Support HTTP proxy between Lambda and AWS API (SSM & EC2)


axel3rd:
### Overview of the Feature / Motivation for or Use Case

When GitHub Enterprise Server is deployed on-premise on an intranet network with security network constraints (like DMZ from internet access ~ingress, http proxy for internet access ~egress), the VPC where deploying Lambda should have direct access to this on-premise GitHub. 

This VPC should be routed with intranet network and `lambda_subnet_ids` & `lambda_security_group_ids` parameters should be used.

But in this case, the VPC has same network constraints than intranet network, so AWS API cannot be accessed directly (not know in DNS) and http proxy should be used ; because AWS API is an internet access.

=> Supporting http proxy for Lambda interactions with AWS API (SSM & EC2) could be useful. 

### Reproduce the Error

Detailled *scale-up* Lambda (for *ssm* access):

```
ERROR	Error: getaddrinfo ENOTFOUND ssm.eu-west-3.amazonaws.com
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:66:26) {
  errno: 'ENOTFOUND',
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'ssm.eu-west-3.amazonaws.com',
  '$metadata': { attempts: 1, totalRetryDelay: 0 }
}
```

### Suggest a Fix

Implementing a proxy support for [aws-sdk v3](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html) ([ssm.ts](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/lambdas/runners/src/scale-runners/ssm.ts)) & [aws-sdk v2](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-configuring-proxies.html) (mainly [runners.ts](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/lambdas/runners/src/scale-runners/runners.ts)), when "classic" `https_proxy` environment variable is filled (on Lambda), from **lambda_https_proxy** Terraform parameter (sample value: `http://foo:bar@proxy.company.com:8080`).

Pull-Request will be proposed with this suggestion.

axel3rd:
> This issue has been automatically marked as stale because it has not had activity in the last 30 days

[...]

PR #1256, opened 2021-10-11
fix: #1218 Proxy implementation for Lambdas (basic UT for aws-sdk v2) 

axel3rd:
PR proposal for fixing #1218.

----

As first minor review, perhaps `https_proxy` could be renamed `HTTPS_PROXY` for coherence in environment variables naming (even if proxy variables is generally named in lowercase):

image

---- 

Tested in "real condition", working fine about EC2 launch:

image

PS: End2end use case is always in progress, `templates/user-data.sh` should be completely customized to manage proxy usage.


axel3rd:
> why did you only add the proxy to the runner lambda? And not to the other two?

If by "other two" you mean *scale-up* & *scale-down*: the `lambda.ts` becomes the `index.js` and the AWS lambdas deployment binds `index.scaleUp` & `index.scaleDown` for each. 
For "aws-sdk v2" we can define proxy only once with `config.update` (contrary to "aws-sdk v3") => `lambda.ts` seems relevant.

Otherwise I miss something (my apologies), my real tests was only on scale-up & scale-down.

axel3rd:
@npalm : Ready for review (basic unit tests added for AWS SDK v2 ; all AWS request API are not mocked).

npalm:
Thx, will check this week. I have no option to test the proxy.

npalm:
@ScottGuymer would you have a bit of time to check the PR?

axel3rd:
For understanding why a proxy, and linked with #1303, a WIP/draft of a future [template example](https://github.com/GroupePSA/terraform-aws-github-runner/tree/exemple-ubuntu-vpc-routed-proxy/examples/ubuntu-ghes-vpc-routed-proxy) of this use case.

modules/runners/lambdas/runners/package.json
```diff
@@ -10,7 +10,7 @@
     "lint": "yarn eslint src",
     "watch": "ts-node-dev --respawn --exit-child src/local.ts",
     "build": "ncc build src/lambda.ts -o dist",
-    "dist": "yarn build && cd dist && zip ../runners.zip index.js",
+    "dist": "yarn build && cd dist && zip ../runners.zip index.js contextify.js",
```
ScottGuymer, line 13: What does this change do?

modules/runners/lambdas/runners/src/scale-runners/ssm.ts
```diff
@@ -1,6 +1,19 @@
...
+
+  // Proxy with aws-sdk v3
+  // Configured by client (global configuration like v2 doesn't work)
+  // https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html
+  const httpsProxy = process.env.HTTPS_PROXY;
+  if (httpsProxy != null && httpsProxy.startsWith('http')) {
+    const rh = new NodeHttpHandler({
+      httpsAgent: new proxy(httpsProxy),
+    });
+    client = new SSM({ region: process.env.AWS_REGION, requestHandler: rh });
```
ScottGuymer, line 16: Structuring this without an else means that in the case that the proxy value is set then SSM is newed up twice.

It would be better to only do this once in either case.

modules/runners/lambdas/runners/src/scale-runners/ssm.ts
```diff
@@ -1,6 +1,19 @@
...
+  }
```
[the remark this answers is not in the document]
ScottGuymer: You can see some more extensive examples of mocking dependencies in https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runner-binaries-syncer/lambdas/runner-binaries-syncer/src/syncer/syncer.test.ts

Specifically the `mockedAxios` might help out

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -22,10 +25,25 @@ export async function scaleUpHandler(event: SQSEvent, context: Context, callback
...
 }
+
+/**
+ * Proxy with aws-sdk v2
+ * configured globally for all aws client(s) => SSM & EC2 in 'runners.ts' file ('ssm.ts' is already in aws-sdk v3)
+ * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-configuring-proxies.html
+ */
+export function configureProxyAwsSdkV2Only() {
+  const httpsProxy = process.env.HTTPS_PROXY;
+  if (httpsProxy != null && httpsProxy.startsWith('http')) {
```
ScottGuymer, line 44:
```suggestion
  if (httpsProxy != null && httpsProxy.startsWith('https')) {
```

modules/runners/lambdas/runners/src/lambda.test.ts
```diff
@@ -0,0 +1,38 @@
...
+  afterEach(() => {
+    process.env = { ...cleanEnv };
+
+    // Reset potential agent in AWS config
+    delete config.httpOptions?.agent;
+  });
+
+  it('Proxy configured', async () => {
+    // No 'https_proxy' environment variable should exist, not really testable otherwise
+    if (process.env.https_proxy != null) {
```
ScottGuymer, line 20: Why is this? Since we are setting it to the desired value in L23.

modules/runners/variables.tf
```diff
@@ -320,6 +320,12 @@ variable "lambda_security_group_ids" {
   default     = []
 }
 
+variable "lambda_https_proxy" {
```
ScottGuymer, line 323: Maybe make this more generic naming as it is not just about the lambda that need proxying, its anything in the VPC that needs internet access.

variables.tf
```diff
@@ -327,6 +327,12 @@ variable "lambda_security_group_ids" {
   default     = []
 }
 
+variable "lambda_https_proxy" {
```
ScottGuymer, line 330: I think this needs to eb clear its only used in the runners at the moment and is not implemented in the other lambdas. Just add to the description.

ScottGuymer (changes requested):
Hi

Thanks for the contribution.

I would suggest adding in the configuration of the https_proxy to the user-data scripting by default so it does not need to be customised directly for all usages of this feature.

This adds a little extra complexity to the overall setup, which is fine if this is a widely used way of working. Is it not possible to configure your VPC in such a way that it is able to speak to the AWS API without the need for a proxy? I can understand why you may want to block it generally but in this case it really is a key part of the solution and having to proxy all calls could introduce new failure modes on the instances in the future..

modules/runners/lambdas/runners/package.json
axel3rd: With the add of [proxy-agent](https://www.npmjs.com/package/proxy-agent), the lambda build changes with:

```sh
$ yarn build
yarn run v1.22.15
$ ncc build src/lambda.ts -o dist
ncc: Version 0.32.0
ncc: Compiling file index.js into CJS
ncc: Using typescript@4.5.2 (local user-provided)
    3kB  dist/fixasync.js
   17kB  dist/sandbox.js
   37kB  dist/contextify.js
12229kB  dist/index.js
12286kB  [XXms] - ncc 0.32.0
```

If `contextify.js` is not in package (near `index.js`), the lambdas (scale up / down) fails with:

```sh
2021-12-17 15:25:23.525  ERROR [runtime/index.js:30 process.<anonymous>] Uncaught Exception 	{
    "stack": "Error: ENOENT: no such file or directory, open '/var/task/contextify.js'\n    at Object.openSync (fs.js:497:3)\n    at Object.readFileSync (fs.js:393:35)\n    at loadAndCompileScript (/var/task/index.js:144447:18)\n    at Object.74357 (/var/task/index.js:144474:20)\n    at __nccwpck_require__ (/var/task/index.js:158489:43)\n    at Object.69440 (/var/task/index.js:144371:18)\n    at __nccwpck_require__ (/var/task/index.js:158489:43)\n    at Object.64033 (/var/task/index.js:82305:15)\n    at __nccwpck_require__ (/var/task/index.js:158489:43)\n    at Object.37055 (/var/task/index.js:129542:23)",
    "message": "ENOENT: no such file or directory, open '/var/task/contextify.js'",
    "errno": -2,
    "syscall": "open",
    "code": "ENOENT",
    "path": "/var/task/contextify.js"
} 
```

To be honest, I'm not enough aware in node/typescript to understood finely how it works. But it seems mandatory 😁.

modules/runners/lambdas/runners/src/scale-runners/ssm.ts
```diff
@@ -1,6 +1,19 @@
...
 
 export async function getParameterValue(parameter_name: string): Promise<string> {
-  const client = new SSM({ region: process.env.AWS_REGION });
+  let client = new SSM({ region: process.env.AWS_REGION });
+
+  // Proxy with aws-sdk v3
+  // Configured by client (global configuration like v2 doesn't work)
+  // https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html
+  const httpsProxy = process.env.HTTPS_PROXY;
+  if (httpsProxy != null && httpsProxy.startsWith('http')) {
```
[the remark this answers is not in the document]
axel3rd: **Not agree with that**.

Even if proxy is used for http**s** endpoints, the proxy url could remain htt**p**://proxy.company.com. For sample, it is the case in my company.
(I don't know if a http proxy could be an https intermediate 🤔).

modules/runners/lambdas/runners/src/scale-runners/ssm.ts
axel3rd: Agree, I will have a look on that.

modules/runners/lambdas/runners/src/lambda.ts
axel3rd: (Not agree, see previous comment)

modules/runners/main.tf
```diff
@@ -138,6 +138,7 @@ resource "aws_launch_template" "runner" {
     start_runner    = templatefile(local.userdata_start_runner[var.runner_os], {})
     ghes_url        = var.ghes_url
     ghes_ssl_verify = var.ghes_ssl_verify
+    https_proxy     = var.lambda_https_proxy
```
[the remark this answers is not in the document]
axel3rd: I should learn #1444 to better understand the impact with this PR.

One item to have in mind when EC2 VM is in a "routed VPC" ; the AWS role permissions doesn't work => to access SSM or S3 (to download runner, if not packed in image) some AWS credentials should be configured.
It could be nice if it could come from `TF_VAR_` environment variable.

modules/runners/lambdas/runners/src/lambda.test.ts
axel3rd: On *Proxy not configured* unit test (L30), when you have an `https_proxy` configured on shell, it causes problem to the test because a proxy is already configured.
I have put here too to be sure to have a "clean" environment variable before running test.

But I take the point, I will check if when a proxy is configured, the L23 overrides correctly the configuration 👍.

variables.tf
axel3rd: Argee, I will take the point.

modules/runners/variables.tf
axel3rd: > anything in the VPC that needs internet access

Not fully agree with that. Having other lambdas (webhook, runner-syncer) could complexify the behavior of this template.

> Maybe make this more generic naming

Perhaps naming it **http_proxy** could do the job ? And the detail of usage in description (as request in below comment) ?
I choose this name initially because only used by lambdas for AWS API (https) endpoints.

I wait your advice for final choice.

axel3rd:
@ScottGuymer: Sorry for my response delay, full time on a "logger problem" since ~10 days.

> I would suggest adding in the configuration of the https_proxy to the user-data scripting by default so it does not need to be customised directly for all usages of this feature.

Not agree with that ; when EC2 VM are in "routed VPC" and GitHub GHES in this same VPC or on corporate network, the runner (on VM) should be in capacity to request GitHub directly without proxy.

If runner communicate to GitHub with proxy, some reverse proxy (to secure internet access) could be in game and complexify the process (for sample: authentication required, which doesn't support GitHub OAuth currently).

=> "Proxy by default" introduces more problem IMO (vs configured for requests which need it).

> Is it not possible to configure your VPC in such a way that it is able to speak to the AWS API without the need for a proxy? 

I would like ... but no way to deal on that with security team currently.
For any resource on corporate network (on-premises or routed VPC):
- (egress) all resource which would access to internet should use proxy
- (ingress) all internet access to these resources should pass by reverse-proxy (with authentication in some case)

modules/runners/lambdas/runners/src/lambda.test.ts
axel3rd: *TODO for me: "if test" about https_proxy existence can be removed, useless, correctly overridden by* `process.env.HTTPS_PROXY`.

modules/runners/lambdas/runners/src/scale-runners/ssm.ts
axel3rd: I learn the usage of `undefined` in typescript (#newbie 😁).
UT are passing, I will do a real global proxy test after all changes requested in PR (some decisions/remarks are pending).

modules/runners/variables.tf
ScottGuymer: I agree with your point about the other lambdas not being in the VPC. But adding the https_proxy env var to the runners seems sensible.

So maybe `http_proxy` or `https_proxy`, with a description indicating it will be used anywhere a proxy could be needed. And then be more specific with the description on any modules that it might be passed to.

axel3rd: Ok for `http_proxy` (more generic than `https_proxy`, which not pre-suppose the protocol usage), with description depending where it is used.

axel3rd: Done with [859c1db](https://github.com/philips-labs/terraform-aws-github-runner/pull/1256/commits/859c1dba7b5788e8964222dace9501e5209e31e4).

modules/runners/main.tf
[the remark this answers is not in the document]
axel3rd: I am scared than on a routed VPC, the problem with proxy will be the same in [start-runner.sh](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/start-runner.sh) than user-data.sh.
Prebuild image with packer would be hard to use in this case 😢.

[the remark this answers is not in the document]
ScottGuymer: There is already a requirement for the instance to communicate with instance metadata endpoints and SSM for example. How do you currently handle that?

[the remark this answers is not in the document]
axel3rd: @ScottGuymer : Relate PR #1539 for global sample.

When VM is on a routed VPC (no direct internet access without proxy), VM can communicate with [metadata without proxy](https://github.com/GroupePSA/terraform-aws-github-runner/blob/8de24b0ae9f5186eca14bd0e8faf2f6b5456d747/examples/ubuntu-ghes-vpc-routed-proxy/templates/user-data.sh#L14-L17), but any [AWS API endpoint like SSM should use proxy](https://github.com/GroupePSA/terraform-aws-github-runner/blob/8de24b0ae9f5186eca14bd0e8faf2f6b5456d747/examples/ubuntu-ghes-vpc-routed-proxy/templates/user-data.sh#L38-L55), because the AWS "*env*-runner-role" doesn't work in this case (or I miss something in AWS usage ... I'm not an expert).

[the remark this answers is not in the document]
ScottGuymer: Do you have an example of your VPC setup so that we can see why these things won't work? 

I'm not sure what setup would mean that an instance would not be able to use its own identity to use the AWS CLI to retrieve the correct information. I can see that the setup might need `https_proxy` set so that the calls sent by AWS CLI can be proxied.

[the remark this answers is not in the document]
axel3rd: **Synthesis**: I will remove the `https_proxy` in user-data template, so useless.

------

> Do you have an example of your VPC setup so that we can see why these things won't work?

To be honest, this kind of network configuration is too complex for me. I understand the network architecture (fw, palo, proxy, reverse-proxy, ...) and usage ; but I don't know how to configure/understand it in AWS VPC configuration.
But now we don't care, see below.

> what setup would mean that an instance would not be able to use its own identity to use the AWS CLI to retrieve the correct information

By these sentence, you gave me a wonderful Christmas present 🎁 !

Reading carefully AWS [HTTP proxy configuration](https://docs.aws.amazon.com/en_gb/AmazonECS/latest/developerguide/http_proxy_config.html), the usage of role seems possible when proxy is used ; but `169.254.169.254` is important.

From my [incoming sample](https://github.com/GroupePSA/terraform-aws-github-runner/blob/8de24b0ae9f5186eca14bd0e8faf2f6b5456d747/examples/ubuntu-ghes-vpc-routed-proxy/templates/user-data.sh#L15), the access to `169.254.169.254` (metadata) is possible when proxy is not configured.

=> The (trivial, after understanding) solution is: `export no_proxy=169.254.169.254` ; and *aws cli* is working like a charm.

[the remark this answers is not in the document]
axel3rd: > Do you have an example of your VPC setup so that we can see why these things won't work?

To be honest, this kind of network configuration is too complex for me. I understand the network architecture (fw, palo, proxy, reverse-proxy, ...) and usage ; but I don't know how to configure/understand it in AWS VPC configuration.
But now we don't care, see below.

> what setup would mean that an instance would not be able to use its own identity to use the AWS CLI to retrieve the correct information

By these sentence, you gave me a wonderful Christmas present 🎁 !

Reading carefully AWS [HTTP proxy configuration](https://docs.aws.amazon.com/en_gb/AmazonECS/latest/developerguide/http_proxy_config.html), the usage of role seems possible when proxy is used ; but `169.254.169.254` is important.

From my [incoming sample](https://github.com/GroupePSA/terraform-aws-github-runner/blob/8de24b0ae9f5186eca14bd0e8faf2f6b5456d747/examples/ubuntu-ghes-vpc-routed-proxy/templates/user-data.sh#L15), the access to `169.254.169.254` (metadata) is possible when proxy is not configured.

=> The (trivial, after understanding) solution is: `export no_proxy=169.254.169.254` ; and *aws cli* is working like a charm.

**In synthesis**: The proxy is required in user-data template, but not AWS credentials.

And it is really a 🎁, because this limitation (AWS API usage via cli/sdk/etc from routed VPN) was a pain points for many other items.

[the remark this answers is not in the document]
ScottGuymer: OK so the combination of `https_proxy` and `no_proxy` enables the use of both the meta data endpoint and the CLI from the instance itself. Meaning there should be no specific need to customise the user-data for this use case, there may be other customisations needed using the pre or post install hooks and this is possible (for instance we set some specific network routing config on the instance).

If we add the proxy settings into the instance with the standard user-data does this make #1539 no longer needed?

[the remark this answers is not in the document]
axel3rd: > If we add the proxy settings into the instance with the standard user-data does this make #1539 no longer needed?

This sample is for detail the "particular" usage of GHES + VPC routed + runners at enterprise level. It could be always useful IMO.

axel3rd:
Set to draft even if all checks pass, because when tested really after last `develop` merge, problem on `webhook` runner (even if not part of these changes):

image

axel3rd:
> Set to draft even if all checks pass, because when tested really after last develop merge, problem on webhook runner (even if not part of these changes)

Due to previous release `webhook.zip` lambda usage. Works fine with rebuild from `develop`.

axel3rd:
@ScottGuymer : Happy new year ! I think have answered to all request/question. Don't hesitate for any complements if it is not the case.

modules/runners/lambdas/runners/package.json
```diff
@@ -42,6 +42,7 @@
     "@types/express": "^4.17.11",
     "@types/node": "^17.0.7",
     "aws-sdk": "^2.1048.0",
+    "proxy-agent": "^5.0.0",
```
npalm, line 45: Checked the library, tiny but last half year not maintained https://www.npmjs.com/package/proxy-agent

axel3rd: > last half year not maintained

Agree, but [wide used](https://github.com/TooTallNate/node-proxy-agent/network/dependents?package_id=UGFja2FnZS0xNjEwMzAxMQ%3D%3D). Not sure to have another better solution 😢.

axel3rd:
@npalm / @ScottGuymer : Sorry to bother you or seems to be insistent, but what could be the following of this PR ? Currently its status is *Changes requested* but I'm not sure to have concrete updates todo on it.
Let me know, many thanks.

modules/runners/lambdas/runners/package.json
ScottGuymer: I guess its a new bundle dependency based on new things added to the imports.

modules/runners/lambdas/runners/src/aws/ssm.ts
```diff
@@ -1,6 +1,18 @@
 import { SSM } from '@aws-sdk/client-ssm';
+import { NodeHttpHandler } from '@aws-sdk/node-http-handler';
+import proxy from 'proxy-agent';
 
 export async function getParameterValue(parameter_name: string): Promise<string> {
-  const client = new SSM({ region: process.env.AWS_REGION });
+  // Proxy with aws-sdk v3
+  // Configured by client (global configuration like v2 doesn't work)
+  // https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-proxies.html
+  let rh = undefined;
```
ScottGuymer, line 9:
```suggestion
  let rh;
```

modules/runners/lambdas/runners/src/aws/ssm.ts
```diff
@@ -1,6 +1,18 @@
...
+  const httpsProxy = process.env.HTTPS_PROXY;
+  if (httpsProxy != null && httpsProxy.startsWith('http')) {
```
ScottGuymer, line 11:
```suggestion
  if (httpsProxy !== null && httpsProxy.startsWith('http')) {
```

modules/runners/lambdas/runners/src/aws/ssm.test.ts
```diff
@@ -56,4 +58,23 @@ describe('Test getParameterValue', () => {
     // Assert
     expect(result).toBe(undefined);
   });
+
+  test('Check that proxy is used', async () => {
```
ScottGuymer, line 62: Id like a test case to show the mocked proxy isnt used when the env var is not configured.

Also some other cases like
- when the proxy env var is malformed (ie without http)
- what if the proxy library returns an error for some reason

modules/runners/lambdas/runners/src/lambda.test.ts
```diff
@@ -157,3 +164,55 @@ describe('Adjust pool.', () => {
...
+      throw new Error('Please remove "https_proxy" environment variable, not testable otherwise');
+    }
+    configureProxyAwsSdkV2Only();
+    expect(config.httpOptions?.agent).toBeUndefined();
+  });
+});
+
+describe('Test proxy configuration for AWS SDK v2', () => {
```
ScottGuymer, line 194: These seem to be duplicates.

Also see my comment on the other tests above. I think there are more cases we need to add here.

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -48,3 +52,17 @@ export async function adjustPool(event: PoolEvent, context: Context): Promise<vo
     logger.error(e);
   }
 }
+
+/**
+ * Proxy with aws-sdk v2
+ * configured globally for all aws client(s) => SSM & EC2 in 'runners.ts' file ('ssm.ts' is already in aws-sdk v3)
+ * https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/node-configuring-proxies.html
+ */
+export function configureProxyAwsSdkV2Only() {
```
ScottGuymer, line 61: Is there any way to reduce the side effects of this function? It operates on the config of something that is not passed to it and it does not return anything.

Maybe pass the config in as a parameter

modules/runners/lambdas/runners/src/aws/ssm.ts
npalm: The check can also be re-written to `if (httpsProxy?.startsWith('http'))`

modules/runners/lambdas/runners/src/lambda.test.ts
axel3rd: > These seem to be duplicates.

Oups good catch, bad last rebase 😬.

modules/runners/lambdas/runners/src/lambda.ts
axel3rd: > Maybe pass the config in as a parameter

Previously I didn't found how to pass proxy configuration as parameter in SDKv2, contrary to SDKv3.
I will double check.

modules/runners/lambdas/runners/src/aws/ssm.test.ts
axel3rd: 3 use cases added.

modules/runners/lambdas/runners/src/lambda.ts
axel3rd: Re-check, proxy can only be configured globally with AWS SDKv2.

But `configureProxyAwsSdkV2Only`was updated with:
- AWS config as parameter
- Return (if update), used in UT

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -48,3 +52,17 @@ export async function adjustPool(event: PoolEvent, context: Context): Promise<vo
...
+  const httpsProxy = process.env.HTTPS_PROXY;
+  if (httpsProxy != null && httpsProxy.startsWith('http')) {
+    config.update({
```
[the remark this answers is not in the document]
axel3rd: Added in `debug` level, for SDKv2 & SDKv3 (=> log could be repeated due to multiple SSM usage in some process).
Files `src/utils/url.ts` & `src/utils/url.test.ts` have been added for removing potential URL credentials in logs.

axel3rd:
> I have done a more in depth review with a number more comments.
> [...]. Just need to add a little more testing and completeness to it.

@ScottGuymer : Updates done. Let me now if it is acceptable 🤞.

modules/runners/lambdas/runners/src/lambda.ts
ScottGuymer: Great, this was what I meant.

ScottGuymer:
I am happy with the changes here. 

I don't have any environment in which I can test this, and I guess you don't have one where there isn't a proxy either. 

So, it's quite difficult to be totally sure there are no side effects.

@npalm what do you think?

axel3rd:
> @npalm what do you think?

@npalm : I allow me to bump this request 😁.

npalm:
@axel3rd sorry for keep you waiting, thx for the notice. Put it on my list for checking

axel3rd:
> SonarCloud Code Analysis Failing after 39s — Quality Gate failed 

About: 
- [1 Bugs](https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&pullRequest=1256&id=philips-labs_terraform-aws-github-runner): False positive IMO, waiting SonarSource feedback (follow [typescript:S3001 false-positive when “delete” operator used on optional properties object](https://community.sonarsource.com/t/typescript-s3001-false-positive-when-delete-operator-used-on-optional-properties-object/59200))
- [8 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=philips-labs_terraform-aws-github-runner&pullRequest=1256&sinceLeakPeriod=true): Could be set *Resolved as Reviewed* (mainly `http://` proxy prefix in unit-tests)

npalm:
> > SonarCloud Code Analysis Failing after 39s — Quality Gate failed
> 
> About:
> 
> * [1 Bugs](https://sonarcloud.io/project/issues?resolved=false&types=BUG&sinceLeakPeriod=true&pullRequest=1256&id=philips-labs_terraform-aws-github-runner): False positive IMO, waiting SonarSource feedback (follow [typescript:S3001 false-positive when “delete” operator used on optional properties object](https://community.sonarsource.com/t/typescript-s3001-false-positive-when-delete-operator-used-on-optional-properties-object/59200))
> * [8 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=philips-labs_terraform-aws-github-runner&pullRequest=1256&sinceLeakPeriod=true): Could be set _Resolved as Reviewed_ (mainly `http://` proxy prefix in unit-tests)

SonarQube was an experiment, it is disabled. I still hope to disable it later. So sonar can be ignored

npalm:
Tried the current PR with the default example, got the following error in scale up

`    "errorMessage": "ENOENT: no such file or directory, open '/var/task/bridge.js'",`

BinHavin0323 (approved):

axel3rd:
> "zip ../runners.zip *.js" / Why some other file than index.js

Using `ncc build -d` option:

```
Emitting /git/terraform-aws-github-runner/modules/runners/lambdas/runners/node_modules/vm2/lib/setup-node-sandbox.js for static use in module /git/terraform-aws-github-runner/modules/runners/lambdas/runners/node_modules/vm2/lib/nodevm.js
```

The `vm2` dependency comes from `pac-proxy-agent` dependency:

```
$ npm list vm2
github-runner-lambda-scale-runners@1.0.0 /git/terraform-aws-github-runner/modules/runners/lambdas/runners
└─┬ proxy-agent@5.0.0
  └─┬ pac-proxy-agent@5.0.0
    └─┬ pac-resolver@5.0.0
      └─┬ degenerator@3.0.2
        └── vm2@3.9.9
```

... probably not used because no proxy PAC is used here 😢; but not sure this transitive dependency can be removed with NPM.

axel3rd:
> but not sure this transitive dependency can be removed with NPM.

Trying with *ncc*:
```
    "build": "ncc build src/lambda.ts -o dist -e pac-proxy-agent -e socks-proxy-agent",
    "dist": "yarn build && cd dist && zip ../runners.zip index.js",
```

The additional JS are not in ZIP, but error at lambda usage 😢:
```
"Runtime.ImportModuleError: Error: Cannot find module 'pac-proxy-agent'",
```

axel3rd:
> [axel3rd](https://github.com/axel3rd) dismissed [BinHavin0323](https://github.com/BinHavin0323)’s stale review via 9fc0da3 16 seconds ago

😱 (it was just to update `package.json` for conflict resolution)

axel3rd:
Is there a chance to merge this PR ? I would start to work on #1303 and it is a prerequisite (not easy to maintain correctly both PR correctly separated)

axel3rd:
>[github-actions](https://github.com/apps/github-actions) bot added the[ Stale ](https://github.com/philips-labs/terraform-aws-github-runner/labels/Stale)label [yesterday](https://github.com/philips-labs/terraform-aws-github-runner/pull/1256#event-6834174626)

Please let me know if any changes are required. It would be nice if this proxy support could be supported IMO.

axel3rd:
> This pull request has been automatically marked as stale because it has not had activity in the last 30 days. It will be closed if no further activity occurs. Thank you for your contributions.

This PR is always relevant IMO.

mcaulifn:
down stale bot, down

PR #1343, opened 2021-10-26
feat: Extrct webhook_secret from github_app object

npalm:
However the webhook and app can be combinded they fullfill different purposes. Therefor a clear split betwoon variables.

See also #1342

### DIFF
--- a/modules/ssm/ssm.tf
+++ b/modules/ssm/ssm.tf
@@ -14,10 +14,10 @@ resource "aws_ssm_parameter" "github_app_key_base64" {
   tags   = var.tags
 }
 
-resource "aws_ssm_parameter" "github_app_webhook_secret" {
-  name   = "/actions_runner/${var.environment}/github_app_webhook_secret"
+resource "aws_ssm_parameter" "github_webhook_secret" {
+  name   = "/actions_runner/${var.environment}/github_webhook_secret"
   type   = "SecureString"
-  value  = var.github_app.webhook_secret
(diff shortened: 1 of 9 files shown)


modules/ssm/ssm.tf
```diff
@@ -14,10 +14,10 @@ resource "aws_ssm_parameter" "github_app_key_base64" {
...
-resource "aws_ssm_parameter" "github_app_webhook_secret" {
-  name   = "/actions_runner/${var.environment}/github_app_webhook_secret"
+resource "aws_ssm_parameter" "github_webhook_secret" {
+  name   = "/actions_runner/${var.environment}/github_webhook_secret"
   type   = "SecureString"
-  value  = var.github_app.webhook_secret
+  value  = var.webhook_secret
   key_id = local.kms_key_arn
   tags   = var.tags
 }
```
toast-gear, line 23: Perhaps should be done in a seperate ticket but the webhook secret is optional and so this resource should be optional too?

toast-gear:
After learning that the `workflow_job` event doesn't include the runner group information whilst the status is `queued` I don't think this PR is right thing to do. I think instead the documentation should be updated to highlight that the webhook could be created within the app or externally, however, due to the lack of runner group information in the event whilst queued it's best to manage the hook from within the app itself so the end user can support multiple runner groups.

ScottGuymer:
Is this still relevant?

npalm:
Let's close it

Issue #1280, opened 2021-10-12
Change `Environment` tag to something else?


toast-gear:
Hello again,

Great module, it's working really well so far from our testing.

One thing that confuses me slightly is that as far as I can tell, you are expected to deploy many stacks of this module if you are making use of runner groups, 1 stack per runner group. The GitHub Actions platform offers few server-side security features currently and so it's difficult to not make use of runner groups if security is important to your organisation. As a result of this, you often end up with multiple runner groups being implemented for any given environment (production being the most common). The `Environment` input causes problems because it not only gets used to namespace the stack, but also to create / set a tag called `Environment`. If you have to deploy multiple instances of this stack the set value for this tag can't be what someone would traditionally call an "environment" e.g. "production", as it needs to be unique to namespace the instance from other instances of the stack. Furthermore, the tag `Environment` is extremely common and so you end up with a clash between your internal tagging standard, and this modules use of the tag, both in terms of what this tag ends up representing as well as a duplicate key clash.

Would it be better to rename this key to `Namespace` and have the tag it creates also have the key `Namespace` so that people can still apply an `Environment` tag as well as having its value make a bit more sense. Perhaps it seems like semantics but I would think `Environment` is an extremely common key resulting in a clash in lots of organisations. `Namespace` might also cause clashing issues but I would think for fewer. You could use something a bit more utilitarian instead like `ActionsStackPrefix` or even continue with using `Environment` but have the tag creation as a configurable option if it isn't used by the module for something? If none of those work perhaps add a namespace to this tags key to differentiate it from an internal `Environment` tag like `philips-labs/Environment`? You could argue that you should namespace your internal tags however that is much harder to implement once a strategy is in place already.

Interested to hear your thoughts if this is a problem from your perspective. This modules concept of an `Environment` seems very blurry to me if you need to deploy multiple copies of the stack to support multiple runner groups. Perhaps in isolation it makes sense but when deploying this solution into an AWS organisation or running multiple instances of this stack due to the need for multiple runner groups to target the same "environment" e.g. production, this modules concept of an Environment breaks down.

ScottGuymer:
In #1444 I have started to enable the transition to tags prefixed with `ghr:` to prevent this sort of clashing.

Currently it has both option to preserve any existing functionality and allow a smoother transition and removal of the `Environment` tag (and others)

npalm:
solved in PR #1858

PR #1346, opened 2021-10-27
Remove Environment tag from IAM Roles

sammort:
Following discussion on #1328 which would break the module's functionality, this PR takes a different approach to the issue I described in #1280.

## Problem Summary
Using `var.environment` to tag all resources with a key named `Environment` is incompatible with some organization's
tagging policies. This is due to IAM Roles being case insensitive to tags, where an orgnaization is using an `environment` tag for all resources, the Roles cannot be created.

## Behaviour
When deploying to our stack, where we include an `environment` tag on all resources, we encounter this problem on the IAM Roles created for each Lambda:

```
Error: Error creating IAM Role test-runner-action-webhook-lambda-role: InvalidInput: Duplicate tag keys found. Please note that Tag keys are case insensitive.
│ 	status code: 400, request id: 3ac70edf-d44a-4410-ada8-41c786a1593b
│
│   with module.test-github-runner.module.webhook.aws_iam_role.webhook_lambda,
│   on .terraform/modules/test-github-runner/modules/webhook/webhook.tf line 51, in resource "aws_iam_role" "webhook_lambda":
│   51: resource "aws_iam_role" "webhook_lambda" {
```

## Proposed solution 
Stop tagging IAM Roles with `Environment` value, this tag isn't required on IAM Roles to maintain the functionality of this module.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -45,7 +45,7 @@ module "webhook" {
 
   aws_region  = var.aws_region
   environment = var.environment
-  tags        = local.tags
+  tags        = var.tags
   kms_key_arn = var.kms_key_arn
 
   sqs_build_queue               = aws_sqs_queue.queued_builds
@@ -72,7 +72,7 @@ module "runners" {
   vpc_id      = var.vpc_id
   subnet_ids  = var.subnet_ids
   environment = var.environment
-  tags        = local.tags
+  tags        = var.tags
 
   s3_bucket_runner_binaries   = module.runner_binaries.bucket
   s3_location_runner_binaries = local.s3_action_runner_url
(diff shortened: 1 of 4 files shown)


npalm:
Sorry for keeping the PR open so long, but need a bit more time to see the impact.

npalm:
Just a quick update, we are rethinking our model of tagging. And only enforce the few manual tags (mainly on the instances). Sorry for keep you waiting.

ScottGuymer:
We have moved to start using namespaced tags with a prefix of `ghr:` and only use them where strictly needed for the module. This is still in progress and will need ongoing work.

Other tags can be added automatically to all resources by the AWS provider using [Default Tags](https://www.hashicorp.com/blog/default-tags-in-the-terraform-aws-provider) functionality. 

This should mean that we no longer need to pass tags into the module and try and merge them with internal tags etc.

sammort:
Thanks both, really good to see the continued development of this module.

I'll close this PR

PR #1354, opened 2021-10-29
feat: adding message retention seconds

### LINKED ISSUE #953: Add `message_retention_seconds` variable
The default of 4 days is too long to have a bad message stuck in queue. A variable should be added to allow this value to be customized. https://github.com/philips-labs/terraform-aws-github-runner/blob/fd9524c6623924922e446bf312fd91873a7ccdd3/main.tf#L18

rossrollin:
Fixes #953

Set the default to 1 day because jobs can be queued for 24 hours then are cancelled by github.


npalm (approved):
Thanks!

Issue #1340, opened 2021-10-26
Reduce the length of the EC2 runner role and instance profile names


toast-gear:
Horse may have bolted on this one as it would be a breaking change however I think it's worthwhile doing.

https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/policies-runner.tf#L4
https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/policies-runner.tf#L11

The current role and instance profile name is quite verbose. This doesn't cause a technical issue however it isn't a great end user experience. We prefix the role name with the environment which already has to be unique and so having such a long name is unnecessary and makes self-service a bit of a faff. 

Perhaps something less verbose like `"${var.environment}-runner-role` | `"${var.environment}-runner-profile"` (my preferred) or `"${var.environment}-actions-runner-role` | `"${var.environment}-actions-runner-profile"`. Tags are the preferred way of categoring and filtering resources and so I don't think we need to go too overboard on the name, we just need to ensure it will be unique for multiple deployments of this stack and it relates to the GitHub Actions platform.

Further more, paths are already supported and so it seems unnecessary having such a long name. I know changing this is a breaking change which will have implications on people's IAM trust relationships however this is easily enough to fix in the deployment role by just adding a second trust relationship in preperation to upgrading their deployments of this module and so I think it's worth simplifying for the sake of the long term health of the project.

If you don't want to release a breaking change (I think sometimes this is actually the better long term option and something to consider) then a set of optional variables to define the role name and instance profile name will do.

PR #1358, opened 2021-10-30
feat: reducing verbosity of role and profile names

toast-gear:
Fixes: https://github.com/philips-labs/terraform-aws-github-runner/issues/1340

This is a **breaking** change as it will impact people's IAM trust relationships. Please do say if you don't feel comfortable with it, you have my reasoning on the raised issue for why I've raised this PR in this way.


npalm (approved):
LGTM

Issue #1341, opened 2021-10-26
Support literal `\n` characters for GitHub App RSA Key


toast-gear:
Clarity on this section in the README.md:

```
Note that github_app.key_base64 needs to be the base64-encoded .pem file, i.e., the output of 
base64 app.private-key.pem (not directly the content of app.private-key.pem).
```

From that description I assume the base64 string once decoded must include new lines? If so I'd like to raise a feature request to support RSA keys with literal `\n` characters instead.

The reason I am asking is because the GitHub Terraform provider is a provider that you may use with this project to automate the creation of external webhooks with each runner groups API Gateway endpoint as the endpoint for its corresponding webhook. It would be cool if the behaviour of the 2 projects synced up in terms of how the provided credentials are processed. The specific behaviour I'd like see synced up is being introduced in the GitHub provider in this PR https://github.com/integrations/terraform-provider-github/pull/931. I'd like to be able to store my RSA key in a single line with literal `\n` characters, I'd like this for a few reasons:

1. It would be nice to mirror how the auth is processed with the GitHub terraform provider as it is likely that this project and that provider will be used in the same terraform apply e.g. create the webhook after deploying the philips-labs module. Keeping how credentials are handled in sync between the 2 projects will make storing and using the same credentials much easier.
2. GitHub does not handle adding masks (`add-mask`) to multi-line variables very well. This has been refactored a few times but it still isn't great.
3. GitHub makes injecting multi-line secrets into the `$GITHUB_ENV` a faff. It is far easier to inject single line variables into the `$GITHUB_ENV` and more generally work with compared to multi-line variables in GitHub Actions

EDIT I think the change is quite simple, I think we just need to do a find and replace here https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts#L51 the same that has been done in the linked PR, then add a test case for both formats.

toast-gear:
```typescript
async function createAuth(installationId: number | undefined, ghesApiUrl: string): Promise<AuthInterface> {
  const appId = parseInt(await getParameterValue(process.env.PARAMETER_GITHUB_APP_ID_NAME));
  let authOptions: StrategyOptions = {
    appId,
    privateKey: Buffer.from(
      await getParameterValue(process.env.PARAMETER_GITHUB_APP_KEY_BASE64_NAME),
      'base64',
[...]

PR #1368, opened 2021-11-02
feat: support single line private keys

toast-gear:
Fixes: https://github.com/philips-labs/terraform-aws-github-runner/issues/1341

Replace literal `\n` chars with new lines so once the base64 string is decoded the value can a single line string or a multiline string

**Rationale**

1. Single line variables are easier to deal with in GitHub Actions in multiple ways
2. As pointed out in the issue this logic is being applied to the Terraform GitHub provider and so it would be cool if the 2 projects aligned given they are a natural fit

Regex tested against Javascript in https://regex101.com/

EDIT I need to deploy this on my end to check it works as expected as I am not familiar with Typescript unless you can see it will work as is.

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
@@ -48,10 +48,12 @@ async function createAuth(installationId: number | undefined, ghesApiUrl: string
     privateKey: Buffer.from(
       await getParameterValue(process.env.PARAMETER_GITHUB_APP_KEY_BASE64_NAME),
       'base64',
-    // replace literal \n characters with new lines to allow the key to be stored as a 
-    // single line variable. This logic should match how the GitHub Terraform provider 
-    // processes private keys to retain compatibility between the projects
-    ).toString().replace('/[\\n]/g', String.fromCharCode(13)),
+      // replace literal \n characters with new lines to allow the key to be stored as a
+      // single line variable. This logic should match how the GitHub Terraform provider
+      // processes private keys to retain compatibility between the projects
+    )
+      .toString()
+      .replace('/[\\n]/g', String.fromCharCode(13)),
   };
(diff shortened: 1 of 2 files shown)


npalm (commented):
LGTM will test asap

npalm (commented):
Code char 13 is CR, correct? Did you test this is working. I would expe ct that the linefeed should be 10 (LF)

modules/runners/lambdas/runners/src/scale-runners/gh-auth.ts
```diff
@@ -48,7 +48,12 @@ async function createAuth(installationId: number | undefined, ghesApiUrl: string
     privateKey: Buffer.from(
       await getParameterValue(process.env.PARAMETER_GITHUB_APP_KEY_BASE64_NAME),
       'base64',
-    ).toString(),
+      // replace literal \n characters with new lines to allow the key to be stored as a
+      // single line variable. This logic should match how the GitHub Terraform provider
+      // processes private keys to retain compatibility between the projects
+    )
+      .toString()
+      .replace('/[\\n]/g', String.fromCharCode(13)),
```
npalm, line 56: Did you test with CR (windows line feed, char 13)? I was epecting an LF (char 10).

toast-gear: ah yeh, got to love working cross OS! I'll update it to 10.

npalm: There should also be a test case, maybe you can add a test case as well with the `\n`

toast-gear: yeh, I'll try my best but I've never written any typescript before so I'm not super confident I'l be able to do it quickly 😅

npalm (approved):

Issue #1372, opened 2021-11-03
Ephemeral runners


npalm:
## Description

The current runners are short lived but not ephemeral. A runner can process multiple jobs till it got killed by the `scale-down` lambda. Ephemeral runners guarantee a clean environment for every job to improve security. Avoiding any shared state, or left overs between jobs.

## How

GitHub have introduced the `workflow_job` to support scaling ephemeral runners. The runner agent is extended with the option `--ephemeral` which will terminate the agent after a run and de-register it.

- Enable via option `ephemeral` to runners #1375 
- Boot instances faster, e.g. via pre-build ami
- Have an optional pool available of runner
- Handle events that cannot processed due to limits or exceptions

## Implementation

For tracking implementation a GitHub [project](https://github.com/philips-labs/terraform-aws-github-runner/projects/1) is crated to track progress. All work will be first merged to the branch `feature/ephemeral`. Active development (also via PR) on ephemeral runnes can done via PR's to this branch.

skyzyx:
‡ …for GitHub.com.

For GHES, `workflow_job` does not exist in v3.2, and neither does `--ephemeral`. The latter is supposed to be coming in v3.3. There has been no public announcement of `workflow_job` coming to GHES one way or the other.

See also https://github.com/philips-labs/terraform-aws-github-runner/issues/182.

npalm:
@skyzyx thanks for the notice, we will keep the module backwards comptabile. But I planning only to support hte ephemeral runners together with the `workflow_job` event

aidan-mundy:
@skyzyx @npalm Enterprise Server 3.3.0.rc1 dropped yesterday with support for ephemeral runners and `workflow_job`.

ScottGuymer:
Implementing this is in progress. Closing this in favour of #1372. Will track the implementation of this there.

aidan-mundy:
@ScottGuymer 🤔 this _is_ #1372

ScottGuymer:
oops

ScottGuymer:
This is now implemented and released

PR #1374, opened 2021-11-03
add option ephemeral runners

npalm:
# Adding basic support for ephemeral runners

## Implementation

A flag `ENABLE_EPHEMERAL_RUNNERS` is added. Once the flag is enabled and the `workflow_job` events are received the scaler tries to create a runner for every event. 

1. Workflow job event received and ephemeral runners are enabled, wrong combination will generate an error
2. Instance is created for every event, in case of scaling failure the event goes back to SQS and will be retried
3. Once the job is finished on the instance, the instance will terminate it self. 

## Challenges

- What if we miss for some reason an event?
- We still have the scale down runner, which can potential also delete an idle runner that was created for an event.

## Tasks
- [x] update docs
- [x] add ephemeral option
- [x] add error handling and retries
- [x] add deadletter queue, to stop retries after time x
- [x] restrict instance termination, can we restrict to the instance only via a condition?
- [x] merge with changes in packer image PR #1444 

see #1372


npalm:
#fix 1375

npalm:
part of #1372

examples/ephemeral/.terraform.lock.hcl
```diff
@@ -0,0 +1,57 @@
+# This file is maintained automatically by "terraform init".
```
mcaulifn, line 1: This file should probably be excluded.

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -51,14 +55,16 @@ export const scaleUp = async (eventSource: string, payload: ActionRequestMessage
...
+  // ephemeral runners should be created on every event, will only work with `workflow_job` events.
+  if (ephemeral || isQueued) {
     const currentRunners = await listEC2Runners({
       environment,
       runnerType,
       runnerOwner,
     });
     console.info(`${runnerType} ${runnerOwner} has ${currentRunners.length}/${maximumRunners} runners`);
 
+    // TODO: how to handle the event if the max is reached in case of ephemeral runners
```
mcaulifn, line 67: It would have to be unlimited if using ephemeral.

npalm: yepz or an retry mechanism to process events once the limit is reach. Also AWS accounts have soft and hard limits.

examples/ephemeral/.terraform.lock.hcl
npalm: No, why. It is locking the versions

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
mcaulifn: Maybe it could re-queue the message with a delay?

examples/ephemeral/.terraform.lock.hcl
mcaulifn: It is the example so I was thinking it wasn't really needed.

modules/runners/templates/user-data.sh
```diff
@@ -22,4 +22,8 @@ ${install_config_runner}
 
 ${post_install}
 
-./svc.sh start
+# TODO: svc.sh start replaced by run.sh and once stopped / exitted the instance terminated
```
npalm, line 25: First merge #1444

modules/runners/policies/instance-ec2.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": "ec2:TerminateInstances",
+      "Resource": "*",
+      "Condition": {
```
npalm, line 8: Find a better condition

npalm:
@ScottGuymer Please can you have a look on the PR. Still need to update docs, but coding part should be done.

examples/default/main.tf
```diff
@@ -61,7 +63,11 @@ module "runners" {
   instance_types = ["m5.large", "c5.large"]
 
   # override delay of events in seconds
-  delay_webhook_event = 5
+  delay_webhook_event   = 10
```
ScottGuymer, line 66: Any reason for these to change as there is a distinct example for ephemeral too. Are these breaking changes?

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
-    for (const e of event.Records) {
-      await scaleUp(e.eventSource, JSON.parse(e.body));
-    }
-
-    callback(null);
-  } catch (e) {
-    logger.error(e);
-    callback('Failed handling SQS event');
+  if (event.Records.length != 1) {
+    logger.warn('Event ignored, only on record at the time can be handled, ensure the lambda batch size is set to 1.');
```
ScottGuymer, line 12:
```suggestion
    logger.warn('Event ignored, only one record at the time can be handled, ensure the lambda batch size is set to 1.');
```

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
```diff
@@ -541,5 +542,28 @@ describe('scaleUp with public GH', () => {
       expect(createRunner).toHaveBeenNthCalledWith(1, expectedRunnerParams, 'lt-1');
       expect(createRunner).toHaveBeenNthCalledWith(2, expectedRunnerParams, 'lt-2');
     });
+
+    it('ephemeral runners cannot only run with workflow_job event, others should fail.', async () => {
```
ScottGuymer, line 546:
```suggestion
    it('ephemeral runners only run with workflow_job event, others should fail.', async () => {
```

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,14 +16,25 @@ export interface ActionRequestMessage {
...
   const enableOrgLevel = yn(process.env.ENABLE_ORGANIZATION_RUNNERS, { default: true });
   const maximumRunners = parseInt(process.env.RUNNERS_MAXIMUM_COUNT || '3');
   const runnerExtraLabels = process.env.RUNNER_EXTRA_LABELS;
   const runnerGroup = process.env.RUNNER_GROUP_NAME;
   const environment = process.env.ENVIRONMENT;
   const ghesBaseUrl = process.env.GHES_URL;
+  const ephemeralEnabled = yn(process.env.ENABLE_EPHEMERAL_RUNNERS, { default: false });
 
+  if (ephemeralEnabled && payload.eventType != 'workflow_job') {
+    logger.warn(`${payload.eventType} even is not supported in combination with ephemeral runners.`);
```
ScottGuymer, line 31:
```suggestion
    logger.warn(`${payload.eventType} event is not supported in combination with ephemeral runners.`);
```

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,14 +16,25 @@ export interface ActionRequestMessage {
...
+    throw Error(
+      `The workflow_job type ${payload.eventType} is not supported in combination with ephemeral runners.` +
```
ScottGuymer, line 33:
```suggestion
      `The event type ${payload.eventType} is not supported in combination with ephemeral runners.` +
```

ScottGuymer (commented):
Partial review with some small thing

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
-  try {
-    for (const e of event.Records) {
-      await scaleUp(e.eventSource, JSON.parse(e.body));
-    }
-
-    callback(null);
-  } catch (e) {
-    logger.error(e);
-    callback('Failed handling SQS event');
+  if (event.Records.length != 1) {
```
PatrickMennen, line 11: !==

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
   }
+
+  return new Promise((resolve, reject) => {
+    scaleUp(event.Records[0].eventSource, JSON.parse(event.Records[0].body))
+      .then(() => resolve())
+      .catch((e: Error) => {
+        if (e instanceof ScaleError) {
+          reject(e);
+        } else {
+          logger.warn('Ignoring error: ', e);
```
PatrickMennen, line 23: `Ignoring error:  ${e}`

modules/runners/lambdas/runners/src/scale-runners/runners.ts
```diff
@@ -90,15 +90,17 @@ export async function createRunner(runnerParameters: RunnerInputParameters, laun
...
-  runInstancesResponse.Instances?.forEach(async (i: EC2.Instance) => {
-    await ssm
-      .putParameter({
-        Name: runnerParameters.environment + '-' + (i.InstanceId as string),
-        Value: runnerParameters.runnerServiceConfig,
-        Type: 'SecureString',
-      })
-      .promise();
-  });
+  if (runInstancesResponse.Instances != undefined) {
```
PatrickMennen, line 93: !== or if (runInstances.Instances)

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,14 +16,25 @@ export interface ActionRequestMessage {
...
   if (eventSource !== 'aws:sqs') throw Error('Cannot handle non-SQS events!');
   const enableOrgLevel = yn(process.env.ENABLE_ORGANIZATION_RUNNERS, { default: true });
   const maximumRunners = parseInt(process.env.RUNNERS_MAXIMUM_COUNT || '3');
   const runnerExtraLabels = process.env.RUNNER_EXTRA_LABELS;
   const runnerGroup = process.env.RUNNER_GROUP_NAME;
   const environment = process.env.ENVIRONMENT;
   const ghesBaseUrl = process.env.GHES_URL;
+  const ephemeralEnabled = yn(process.env.ENABLE_EPHEMERAL_RUNNERS, { default: false });
 
+  if (ephemeralEnabled && payload.eventType != 'workflow_job') {
```
PatrickMennen, line 30: !==

PatrickMennen (changes requested):

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
```diff
@@ -541,5 +542,28 @@ describe('scaleUp with public GH', () => {
...
+          ...TEST_DATA,
+          eventType: 'check_run',
+        }),
+      ).rejects.toBeInstanceOf(Error);
+    });
+
+    it('creates a ephemeral runner.', async () => {
+      process.env.ENABLE_EPHEMERAL_RUNNERS = 'true';
+      await scaleUpModule.scaleUp('aws:sqs', TEST_DATA);
+      expectedRunnerParams.runnerServiceConfig = expectedRunnerParams.runnerServiceConfig + `  --ephemeral`;
```
gertjanmaas, line 559: Does this test work? seems there are 2 spaces before `--ephemeral`, which I assume should be one.

gertjanmaas (approved):
Other than the one comment, looks good :)

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
npalm: It works, but should be a single space.

README.md
```diff
@@ -299,6 +299,7 @@ Examples are located in the [examples](./examples) directory. The following exam
 - _[Ubuntu](examples/ubuntu/README.md)_: Example usage of creating a runner using Ubuntu AMIs.
 - _[Prebuilt Images](examples/prebuilt/README.md)_: Example usages of deploying runners with a custom prebuilt image.
 - _[Windows](examples/windows/README.md)_: Example usage of creating a runner using Windows as the OS.
+- _[Prebuilt Images](examples/prebuilt/README.md)_: Example usages of deploying runners with a custom prebuilt image.
```
ScottGuymer, line 302: This is now in there twice

examples/ephemeral/README.md
```diff
@@ -0,0 +1,31 @@
+# Action runners deployment ephemeral example
+
```
ScottGuymer, line 2: Some more info on how this is different to a default setup and the key settings that make it different would be useful here

variables.tf
```diff
@@ -464,3 +469,27 @@ variable "runner_os" {
...
 }
+
+variable "fifo_build_queue" {
+  description = "Enable a FIFO queue to remain the order of events received by the webhook. Suggest to set to true for repo level runners."
+  type        = bool
+  default     = false
+}
+
+variable "redrive_build_queue" {
+  description = "Set options to attach (optional) a dead letter queue to the build queue, the queue between the webhook and the scale up lambda. You have the following options. 1. Disable by setting, `enalbed' to false. 2. Enable by setting `enabled` to `true`, `maxReceiveCount` to a number of max retries, and `deadLetterTargetArn` to null for letting the module create a queue. Or otherwise provide you own queue by setting an ARN."
```
ScottGuymer, line 480: I wonder if adding the ability to bring your own DLQ is just adding too much complexity here. Is there a need to create it outside of the module?

variables.tf
```diff
@@ -96,7 +96,7 @@ variable "runners_lambda_zip" {
 variable "runners_scale_up_lambda_timeout" {
   description = "Time out for the scale up lambda in seconds."
   type        = number
-  default     = 180
+  default     = 30
```
ScottGuymer, line 99: Is this changed for testing or a new default?

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
```diff
@@ -226,7 +227,7 @@ describe('scaleUp with GHES', () => {
       process.env.RUNNER_GROUP_NAME = 'TEST_GROUP';
       await scaleUpModule.scaleUp('aws:sqs', TEST_DATA);
       expectedRunnerParams.runnerServiceConfig =
-        expectedRunnerParams.runnerServiceConfig + `--labels label1,label2 --runnergroup TEST_GROUP`;
+        expectedRunnerParams.runnerServiceConfig + ` --labels label1,label2 --runnergroup TEST_GROUP`;
```
ScottGuymer, line 230: I think it might be better when working with these command to use a string array and then [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) then with a space. Rather than having to worry about spaces in all the strings etc.

`runnerServiceConfig` could be defined as a string array above and you can merge the arrays and join.

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -80,21 +91,25 @@ export async function scaleUp(eventSource: string, payload: ActionRequestMessage
           });
       const token = registrationToken.data.token;
 
-      const labelsArgument = runnerExtraLabels !== undefined ? `--labels ${runnerExtraLabels}` : '';
-      const runnerGroupArgument = runnerGroup !== undefined ? ` --runnergroup ${runnerGroup}` : '';
+      const labelsArgument = runnerExtraLabels !== undefined ? `--labels ${runnerExtraLabels} ` : '';
+      const runnerGroupArgument = runnerGroup !== undefined ? `--runnergroup ${runnerGroup} ` : '';
       const configBaseUrl = ghesBaseUrl ? ghesBaseUrl : 'https://github.com';
+      const ephemeralArgument = ephemeral ? '--ephemeral ' : '';
+      const runnerArgs = `--token ${token} ${labelsArgument}${ephemeralArgument}`;
```
ScottGuymer, line 98: Again it might be better to work with string arrays and join here. More precise and less concerns about spacing etc.

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
```diff
@@ -264,5 +275,15 @@ describe('handler', () => {
       expect(resp.statusCode).toBe(201);
       expect(sendActionRequest).toBeCalled();
     });
+
+    it('handles check_run events with no installation id.', async () => {
```
ScottGuymer, line 279: What happens if it is set to a string?

ScottGuymer (changes requested):
Some small comments

In good shape, almost ready to merge.

variables.tf
npalm: The rational is. The whole setup is immutable. So if you would like to process for whatever reason failures and we manage the queue, an re-create will enforce also the re-creation of this queue. It also allows other configuration, attaching you own way to process / handle messages easier. But maybe we should leave it out as long we have not a "real" use case.

variables.tf
npalm: Changed, because I think the default is a way to high. The same time the messages is blocked as well in case of failure.

modules/runners/lambdas/runners/src/scale-runners/scale-up.test.ts
npalm: Agree, but suggest we raise a seperate issue to get the whole parsing logic refactored. OK?

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
npalm: Agree, but not all properties are valid in all situation.

ScottGuymer: No but you could build the string array based on the logic and then just `join` it at the end without having to worry about what space are needed where.

variables.tf
ScottGuymer: Yeah, I think we should start simple with this and only add things as the use case becomes apparent. This module is already quite complex, we should find ways that these extra use cases can be added by extension on top of the module rather than adding them into the module.

For example, we could set the DLQ as an output that could be used as needed. Someone could take that output and use it as an input to some other module that adds extra functionality to the queue.

npalm: Problem of the output is that the use have no control over the queue, but agree in general. So I will remove the injection of the extra queue. And remain the optional DLQ.

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
npalm: Since we control the data here, and it is typed. Would suggest not to improve this part in this PR

examples/ephemeral/README.md
npalm: Added

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
ScottGuymer: ok no problem

examples/ephemeral/README.md
```diff
@@ -0,0 +1,30 @@
...
+This example is essential based on the default setup. But shows how you can use runners with the ephemeral flag enabled. Once you enable ephemeral a runner will used once, only for one job. This requires that each job needs a fresh instance. To feature should be used in combination with event `workflow_job`, see your GitHub webhook endpoint configuration. We also suggest to use the feature with a pre-build AMI, to optimize the creation time of a runner.
```
ScottGuymer, line 3:
```suggestion
This example is based on the default setup. But shows how you can use runners with the ephemeral flag enabled. Once you enable ephemeral a runner will used once, only for one job. This requires that each job needs a fresh instance. To feature should be used in combination with event `workflow_job`, see your GitHub webhook endpoint configuration. We also suggest using the feature with a pre-build AMI, to optimize the creation time of a runner.
```

ScottGuymer (approved):
All look good now. 

Do you want me to test deploying this?

npalm:
> All look good now.
> 
> Do you want me to test deploying this?

Yes would be great when you can verify. Maybe smart to try the combination with a custom AMI

ScottGuymer:
That will take me some time to get up and running.

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -80,21 +91,25 @@ export async function scaleUp(eventSource: string, payload: ActionRequestMessage
...
-          : `--url ${configBaseUrl}/${payload.repositoryOwner}/${payload.repositoryName} ` +
-            `--token ${token} ${labelsArgument}`,
+          ? `--url ${configBaseUrl}/${payload.repositoryOwner} ${runnerArgs}${runnerGroupArgument}`.trim()
+          : `--url ${configBaseUrl}/${payload.repositoryOwner}/${payload.repositoryName} ${runnerArgs}`.trim(),
         runnerOwner,
         runnerType,
       });
     } else {
       logger.info('No runner will be created, maximum number of runners reached.', LogFields.print());
+      if (ephemeral) {
```
mcaulifn, line 110: I must be missing something: Why is the error only thrown if `ephemeral`?

modules/runners/policies/instance-ec2.json
```diff
@@ -0,0 +1,15 @@
...
+        "StringEquals": {
+          "aws:ARN": "$${ec2:SourceInstanceARN}"
+        }
+      }
+    }
+  ]
+}
```
mcaulifn, line 15: Needs a new line.

modules/webhook/lambdas/webhook/src/sqs/index.ts
```diff
@@ -15,11 +10,20 @@ export interface ActionRequestMessage {
...
+
+  const useFifoQueueEnv = process.env.SQS_IS_FIFO || 'false';
+  const useFifoQueue = JSON.parse(useFifoQueueEnv) as boolean;
+
+  const sqsMessage: SQS.Types.SendMessageRequest = {
+    QueueUrl: String(process.env.SQS_URL_WEBHOOK),
+    MessageBody: JSON.stringify(message),
+  };
+
+  logger.debug(`sending message to SQS: ${JSON.stringify(sqsMessage)}`);
```
mcaulifn, line 23: Please add `, LogFields.print()`

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
-    for (const e of event.Records) {
-      await scaleUp(e.eventSource, JSON.parse(e.body));
-    }
-
-    callback(null);
-  } catch (e) {
-    logger.error(e);
-    callback('Failed handling SQS event');
+  if (event.Records.length !== 1) {
+    logger.warn('Event ignored, only one record at the time can be handled, ensure the lambda batch size is set to 1.');
```
mcaulifn, line 12: Please add `, LogFields.print()`

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
   }
+
+  return new Promise((resolve, reject) => {
+    scaleUp(event.Records[0].eventSource, JSON.parse(event.Records[0].body))
+      .then(() => resolve())
+      .catch((e: Error) => {
+        if (e instanceof ScaleError) {
+          reject(e);
+        } else {
+          logger.warn(`Ignoring error: ${e.message}`);
```
mcaulifn, line 23: Please add `, LogFields.print()`

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
-  } catch (e) {
-    logger.error(e);
-    callback('Failed');
-  }
+
+  return new Promise((resolve) => {
+    scaleDown()
+      .then(() => resolve())
+      .catch((e) => {
+        logger.error(e);
```
mcaulifn, line 36: Please add `, LogFields.print()`

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,14 +16,25 @@ export interface ActionRequestMessage {
 }
 
 export async function scaleUp(eventSource: string, payload: ActionRequestMessage): Promise<void> {
+  logger.info(`Received ${payload.eventType} from ${payload.repositoryOwner}/${payload.repositoryName}`);
```
mcaulifn, line 19: Please add `, LogFields.print()`

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,14 +16,25 @@ export interface ActionRequestMessage {
...
   const enableOrgLevel = yn(process.env.ENABLE_ORGANIZATION_RUNNERS, { default: true });
   const maximumRunners = parseInt(process.env.RUNNERS_MAXIMUM_COUNT || '3');
   const runnerExtraLabels = process.env.RUNNER_EXTRA_LABELS;
   const runnerGroup = process.env.RUNNER_GROUP_NAME;
   const environment = process.env.ENVIRONMENT;
   const ghesBaseUrl = process.env.GHES_URL;
+  const ephemeralEnabled = yn(process.env.ENABLE_EPHEMERAL_RUNNERS, { default: false });
 
+  if (ephemeralEnabled && payload.eventType !== 'workflow_job') {
+    logger.warn(`${payload.eventType} event is not supported in combination with ephemeral runners.`);
```
mcaulifn, line 31: Please add `, LogFields.print()`

mcaulifn (changes requested):

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
npalm: The error is caused due to an issue with scaling. Reached max number of runners, or nog spot available. Since each event for ephemeral needs a runner we throw an exception. Messages goes back to the queue and will be retried later. For non ephemeral an existing runner will be used most likely. That is the same behaviour we have right now. Does this explanation help @mcaulifn

modules/runners/lambdas/runners/src/lambda.ts
npalm: Will do, merge was some hell for this PR

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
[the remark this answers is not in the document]
npalm: Should be covered by unit tests, created also an issue to refactor this part

modules/runners/lambdas/runners/src/lambda.ts
```diff
@@ -2,30 +2,39 @@ import { scaleUp } from './scale-runners/scale-up';
...
-    callback(null);
-  } catch (e) {
-    logger.error(e);
-    callback('Failed handling SQS event');
+  if (event.Records.length !== 1) {
+    logger.warn('Event ignored, only one record at the time can be handled, ensure the lambda batch size is set to 1.');
+    return new Promise((resolve) => resolve());
   }
+
+  return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
npalm: A async lambda should return a Promise not a callback. Will check if a try / catch is needed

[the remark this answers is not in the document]
mcaulifn: [This](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html) is what I was just looking at. 
> For libraries that return a promise, you can return that promise directly to the runtime.

Since `scaleUp` returns a promise, that should suffice?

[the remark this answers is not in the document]
ScottGuymer: It should be enough as long as you have an `await` for any async calls to other functions.

If you forget to await something async you can get wierdness.

ScottGuymer:
@npalm tested the deployment of ephemeral example config. works great so far

ScottGuymer (approved):
Looks good!

ScottGuymer (approved):

npalm:
@ScottGuymer PR is ready for final review. 

Tested:

- [ ] example ephemeral runners
- [ ] example default on branch develop with migration to this branch

PR #1381, opened 2021-11-05
fix: update return codes, no error code for job that are ignored

### LINKED ISSUE #1348: Don't error if the webhook event only has GitHub owned labels
I have webhooks setup at the organisation level in my organisation as I don't want to install them at the repo level as that would result in having to manage hundreds of webhooks. When a `workflow_job` event is sent to my API Gateway the webhook is marked as failed as it receives a HTTP 403 response. The errors in my logs are these 2 over and over again:

```
2021-10-27T14:46:31.491Z	4ff42805-6636-4c5e-89c3-11b0c1225917	
DEBUG	Received workflow job event with labels: '["ubuntu-latest"]'. 
The event does NOT match the configured labels: 'default,self-hosted'

2021-10-27T14:46:31.491Z	4ff42805-6636-4c5e-89c3-11b0c1225917	
ERROR	Received event contains runner labels 'ubuntu-latest' that are not accepted.
```

This is confusing as it makes it look like there is a problem with the configuration of the webhook when really the webhook is fine. I know we added support for ignoring known GitHub self-hosted labels in this PR https://github.com/philips-labs/terraform-aws-github-runner/pull/1244. Given many (probably the majority) of organisations running self-hosted runners will also utilise GitHub's runners too it would be great is we could also make it so when the webhook lambda receives a `workflow_job` event that only contains `runs-on` labels for GitHub owned labels such as `ubuntu-latest`, we ignore them, log the ignore at the debug level (as we do now) and return a HTTP 200 response instead. 

Fixing this so the webhook still receives a HTTP 200 response marking it as green in the GitHub UI regardless of if the job is bound for a philips-labs runner or not would enable an administrator to more easily see real configuration errors such as a mismatched webhook secret.

EDIT the current GitHub managed labels can be found here https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources, workflow jobs that are marked only for any of these should not result in a non HTTP 200 response (or any other label combination imo)

EDIT As a more concrete example, here are what some of my webhooks look like atm on the organisation webhook screen:

image

And here are some `Recent Deliveries` in the one of the webhook settings.

image

They are all working fine but it looks like I have a major configuration issue. This leads to lots of confusion unless the adminstrator investigating the webhook errors is:

A) familiar with the GitHub Actions platform
B) familiar with the implementation details of this module

npalm:
Replaced several response codes:

- 201: for jobs that are queued
- 202: for jobs that are ignored, also details to body added

fix #1348

### DIFF
--- a/modules/webhook/lambdas/webhook/src/lambda.ts
+++ b/modules/webhook/lambdas/webhook/src/lambda.ts
@@ -2,14 +2,17 @@ import { handle } from './webhook/handler';
 import { APIGatewayEvent, Context, Callback } from 'aws-lambda';
 import { logger } from './webhook/logger';
 
+export interface Response {
+  statusCode: number;
+  body?: string;
+}
+
 export const githubWebhook = async (event: APIGatewayEvent, context: Context, callback: Callback): Promise<void> => {
   logger.setSettings({ requestId: context.awsRequestId });
   logger.debug(JSON.stringify(event));
   try {
-    const statusCode = await handle(event.headers, event.body as string);
-    callback(null, {
-      statusCode: statusCode,
-    });
+    const response = await handle(event.headers, event.body as string);
+    callback(null, response);
   } catch (e) {
     callback(e as Error);
   }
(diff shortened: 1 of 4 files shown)


JeroenKnoops (approved):
LGTM

PR #1369, opened 2021-11-02
Webhook Logging

mcaulifn:
- Move metadata to JSON object for more structured logging
- Add additional detail to logs
- Use `supportedEvents` list to quickly exit if not supported
- Making `handle` function definition consistent with other functions in lambda

PR #1401, opened 2021-11-09
fix(logging): Add context to webhook logs

mcaulifn:
- Move metadata to JSON object for more structured logging
- Add additional detail to logs
- Use supportedEvents list to quickly exit if not supported

Closes #1369

### DIFF
--- a/modules/webhook/lambdas/webhook/src/webhook/handler.ts
+++ b/modules/webhook/lambdas/webhook/src/webhook/handler.ts
@@ -30,7 +30,7 @@ export async function handle(headers: IncomingHttpHeaders, body: string): Promis
   LogFields.fields.action = payload.action;
 
   if (!supportedEvents.includes(githubEvent)) {
-    logger.error(LogFields.fields, `Unsupported event type.`);
+    logger.warn(LogFields.fields, `Unsupported event type.`);
     return {
       statusCode: 202,
       body: `Ignoring unsupported event ${githubEvent}`,
--- a/modules/webhook/lambdas/webhook/src/webhook/handler.ts
+++ b/modules/webhook/lambdas/webhook/src/webhook/handler.ts
@@ -3,9 +3,10 @@ import { Webhooks } from '@octokit/webhooks';
 import { sendActionRequest } from '../sqs';
 import { CheckRunEvent, WorkflowJobEvent } from '@octokit/webhooks-types';
 import { getParameterValue } from '../ssm';
-import { logger as rootLogger } from './logger';
+import { logger as rootLogger, LogFields } from './logger';
 import { Response } from '../lambda';
 
+const supportedEvents = ['check_run', 'workflow_job'];
 const logger = rootLogger.getChildLogger();
 
 export async function handle(headers: IncomingHttpHeaders, body: string): Promise<Response> {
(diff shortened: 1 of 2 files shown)


modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -24,33 +25,60 @@ export async function handle(headers: IncomingHttpHeaders, body: string): Promis
...
+
+  if (payload[githubEvent].started_at) {
+    LogFields.fields.started_at = payload[githubEvent].started_at;
+  }
+
+  /*
+  The app subscribes to all `check_run` and `workflow_job` events.
+  If the event status is `completed`, log the data for workflow metrics.
+  */
+  if (payload[githubEvent].completed_at) {
```
npalm, line 51: Could you not replace the if statement by a single line like ```LogFields.fields.completed_at = payload[githubEvent]?.completed_at;```?

npalm (changes requested):
No time to test yet

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -24,33 +25,60 @@ export async function handle(headers: IncomingHttpHeaders, body: string): Promis
     return response;
   }
   const payload = JSON.parse(body);
-  logger.info(`Received Github event ${githubEvent} from ${payload.repository.full_name}`);
+  LogFields.fields.event = githubEvent;
+  LogFields.fields.repository = payload.repository.full_name;
+  LogFields.fields.action = payload.action;
+
+  if (!supportedEvents.includes(githubEvent)) {
+    logger.error(LogFields.fields, `Unsupported event type.`);
```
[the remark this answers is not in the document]
mcaulifn: Changed

npalm (commented):
LGTM, will try to run a quick test tomorrow

npalm (dismissed):

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -24,33 +25,53 @@ export async function handle(headers: IncomingHttpHeaders, body: string): Promis
...
+  /*
+  The app subscribes to all `check_run` and `workflow_job` events.
+  If the event status is `completed`, log the data for workflow metrics.
+  */
+  LogFields.fields.completed_at = payload[githubEvent]?.completed_at;
+  LogFields.fields.conclusion = payload[githubEvent]?.conclusion;
 
   if (isRepoNotAllowed(payload.repository.full_name)) {
-    console.warn(`Received event from unauthorized repository ${payload.repository.full_name}`);
+    logger.error(LogFields.fields, `Received event from unauthorized repository ${payload.repository.full_name}`);
```
npalm, line 52: I think by a small change we can keep the log much cleaner and also readable. 

So would suggest to add an function to `logger.ts`
```TypeScript
  public static print(): string {
    return ' LogFields: ' + JSON.stringify(LogFields.fields);
  }
 ```
And next add the logfiels at the end of the line, for example 

```suggestion
    logger.error(`Received event from unauthorized repository ${payload.repository.full_name}. `, LogFields.fields);
```

npalm (changes requested):
@mcaulifn change is in general fine to me. But would suggest to add the logfields to the end of the line. This to keep the logs also readable. See details on review comment

npalm (approved):
LGTM

PR #1444, opened 2021-11-25
Prebuilt runner image using packer

ScottGuymer:
When looking at using ephemeral runners it becomes clear that startup time for the agents becomes a key factor in build times.

To enable us to shorten the startup times as much as possible we should build an image that has all the dependencies and the runner installed. In this way there is much less work to do when starting the runner.

To do this I have used packer to build an image that can be used for the runners.

This requires a few changes to how the current scripting is done. We need to refactor this into two phases
- installation
- startup

Installation
This is all the work that can be done before the runner starts, so installing dependencies or configuring generic settings

Startup
These are the things that can only happen when you are booting an instance of an agent. Such as collecting the GH token to connect the runner or configuring CloudWatch.


ScottGuymer:
So a few things i have started to do here

- build the image using packer with the current amzn2 image as a base
- refactored the install script to pull the runner directly from a URL (removing the need for the s3 bucket at runtime)
- Added a `startup.sh` script to `/var/lib/cloud/scripts/per-boot/` so it will run at each boot. This does the following
    - Gets the instance meta data required (but now uses tags for some non secret info like environment)
    - Gets the GH API token from SSM
    - Optionally configures cloudwatch based on tag setting
    - configures the runner
    - starts the runner and waits for it to exit (this could be made switchable to make it a long lived runner)
    - Terminates the instance when the runner exists (mostly for ephemeral workloads)
 - Given the instance rights to read tags in IAM to be able to pass config in an simple and extensible way.
 - Added a simple CI validation job to check the format and validate

Whats left to decide/do?
- [ ] How many/where are we happy with breaking changes to create simplicity?
- [ ] Do we allow both image based and startup script based config?
- [ ] Do we build this image in CI? 
- [ ] Do we make this ami public and the default image to use instead of the on demand config?
- [ ] How do we deal with the pre_install hook? 
- [ ] Do we add the post_install hook to the startup so it can run something before the agent starts?
- [ ] Do we use user_data as a point of customisation for the image on start?
- [ ] How do we handle the arm_patch functionality, should this just be bake into arm images?

main.tf
```diff
@@ -1,6 +1,7 @@
 locals {
   tags = merge(var.tags, {
-    Environment = var.environment
+    Environment       = var.environment,
+    "ghr:environment" = format("%s", var.environment)
```
ScottGuymer, line 4: I added this as an additional top level tag so we can start to move to namespaced tags and address issues like #1280 #1346 

We can then slowly start to move references to `Environment` over to `ghr:environment` which is less likely to cause clashes with org policies.

modules/runners/logging.tf
```diff
@@ -28,7 +28,7 @@ resource "aws_cloudwatch_log_group" "gh_runners" {
 }
 
 resource "aws_iam_role_policy" "cloudwatch" {
-  count = var.enable_ssm_on_runners ? 1 : 0
+  count = var.enable_cloudwatch_agent ? 1 : 0
```
ScottGuymer, line 31: This is a buigfix.

If you previously set `var.enable_cloudwatch_agent=false` and `var.enable_ssm_on_runners=true` then it caused an error on this resource.

modules/runners/main.tf
```diff
@@ -3,20 +3,18 @@ locals {
     {
       "Name" = format("%s-action-runner", var.environment)
     },
-    {
-      "Environment" = format("%s", var.environment)
```
ScottGuymer, line 7: This is already specified in the top level module and passed in here

modules/runners/policies-runner.tf
```diff
@@ -26,7 +26,10 @@ resource "aws_iam_role_policy" "ssm_parameters" {
   role = aws_iam_role.runner.name
   policy = templatefile("${path.module}/policies/instance-ssm-parameters-policy.json",
     {
-      arn_ssm_parameters = "arn:aws:ssm:${var.aws_region}:${data.aws_caller_identity.current.account_id}:parameter/${var.environment}-*"
+      arn_ssm_parameters = jsonencode([
```
ScottGuymer, line 29: This now passes a list into the policy json rather than a single line. This is to enable path based access to SSM Parameters

modules/runners/policies/instance-describe-tags-policy.json
```diff
@@ -0,0 +1,10 @@
+{
```
ScottGuymer, line 1: Although `ec2:DescribeTags` is in the `instance-cloudwatch-policy.json` it is only deployed if CloudWatch is turned on.

As we access the tags in the startup to get the `ghr:environment` tag we now always need access to read the tags.

modules/runners/policies-runner.tf
ScottGuymer: changed this to specific values so we can split read and delete and be more specific on the delete.

modules/runners/runner-config.tf
```diff
@@ -0,0 +1,21 @@
+resource "aws_ssm_parameter" "runner_config_run_as" {
+  name  = "/${var.environment}/runner/run-as"
+  type  = "String"
+  value = var.runner_as_root ? "root" : "ec2-user"
+  tags  = local.tags
+}
+
+resource "aws_ssm_parameter" "runner_agent_mode" {
```
ScottGuymer, line 8: Added this in here in preparation for allowing ephemeral runner config

modules/runners/variables.tf
```diff
@@ -292,6 +292,12 @@ variable "runner_log_files" {
       "file_path" : "/var/log/user-data.log",
       "log_stream_name" : "{instance_id}"
     },
+    {
```
ScottGuymer, line 295: The pre build image now places logs here when running the startup and we want it scooped up into cloudwatch.

ScottGuymer:
@npalm this is ready for review now.

.github/workflows/packer-build.yml
```diff
@@ -0,0 +1,35 @@
...
+    paths:
+      - "images/**"
+      - ".github/workflows/packer-build.yml"
+
+env:  
+  AWS_REGION: eu-west-1
+
+jobs:
+  verify_packer:
+    name: Verify image
```
npalm, line 17:
```suggestion
    name: Verify packer    
```

images/linux-amzn2/github_agent.linux.pkr.hcl
```diff
@@ -0,0 +1,87 @@
...
+  ssh_username = "ec2-user"
+  tags = {
+    OS_Version    = "amzn2"
+    Release       = "Latest"
+    Base_AMI_Name = "{{ .SourceAMIName }}"
+  }
+}
+
+build {
+  name = "githubactions-runner"
```
npalm, line 44: Maybe an idea to make this conigurable

modules/runners/policies/instance-describe-tags-policy.json
```diff
@@ -0,0 +1,10 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+      {          
+          "Effect": "Allow",
+          "Action": "ec2:DescribeTags",
+          "Resource": "*"
+      }
+  ]
+}
```
npalm, line 10: I think you can apply this condition to only let the instance access its own tags:

```
            "Condition": {
                "StringEquals": {
                    "aws:ARN": "${ec2:SourceInstanceARN}"
                }
            }
```

modules/runners/templates/install-runner.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Creating actions-runner directory for the GH Action installtion"
+cd /home/"$user_name"
+mkdir actions-runner && cd actions-runner
+
+
+if [[ -n "$RUNNER_TARBALL_URL" ]]; then
+  echo "Downloading the GH Action runner from $RUNNER_TARBALL_URL to $file_name"
+  curl -o $file_name -L "$RUNNER_TARBALL_URL"
+else  
+  echo "Retrieving TOKEN from AWS API"
```
npalm, line 23: Seem we are not using the token here at all

npalm (changes requested):
@ScottGuymer great work! Some minor comments.

Did you also create instructions how to setup (an automated) AMI builder? Will test the PR asap

modules/runners/policies/instance-describe-tags-policy.json
ScottGuymer: Looks like the 2021 version of this is

```
"Condition": {
  "ArnEquals": {
    "aws:SourceArn": "${ec2:SourceInstanceARN}"
  }
}
```

modules/runners/templates/install-runner.sh
ScottGuymer: Its used in the curl on line 26 to get the region

images/linux-amzn2/github_agent.linux.pkr.hcl
ScottGuymer: Its not possible to make this variable. Its only used for logging and naming the instance used for building. It doesn't have a bearing on the end result image.

modules/runners/policies/instance-describe-tags-policy.json
npalm: Just testen, the one with `aws:ARN` is working the one with `aws:SourceArn` not

examples/prebuilt/main.tf
```diff
@@ -0,0 +1,44 @@
...
+  webhook_lambda_zip                = "lambdas-download/webhook.zip"
+  runner_binaries_syncer_lambda_zip = "lambdas-download/runner-binaries-syncer.zip"
+  runners_lambda_zip                = "lambdas-download/runners.zip"
+
+  runner_extra_labels = "default,example"
+
+  # configure your pre-built AMI
+  userdata_enabled = false
+  ami_filter       = { name = ["github-runner-amzn2-2021*"] }
+  ami_owners       = [var.ami_owner]
```
npalm, line 34: You can retrieve the account id as follow

```hcl
data "aws_caller_identity" "current" {}

# usages
ami_owner = aws_caller_identity.current.account_id
```

examples/prebuilt/main.tf
```diff
@@ -0,0 +1,44 @@
...
+
+  # enable access to the runners via SSM
+  enable_ssm_on_runners = true
+
+  # override delay of events in seconds
+  delay_webhook_event = 0
```
npalm, line 40: Would not suggest to set the default to 0 for the check_run event and non ephmeral runners. You simple will create too many runners.

main.tf
```diff
@@ -125,6 +126,7 @@ module "runners" {
   role_path                 = var.role_path
   role_permissions_boundary = var.role_permissions_boundary
 
+  userdata_enabled      = var.userdata_enabled
```
npalm, line 129: Would it be an ideaa to set the enabled via a condition?
```
userdata_enabled = userdata_template != "" ? true : false
```

images/linux-amzn2/github_agent.linux.pkr.hcl
```diff
@@ -0,0 +1,87 @@
...
+}
+
+variable "region" {
+  description = "The region to build the image in"
+  type        = string
+  default     = "eu-west-1"
+}
+
+source "amazon-ebs" "githubrunner" {
+  ami_name      = "github-runner-amzn2-${formatdate("YYYYMMDDhhmm", timestamp())}"
```
npalm, line 23: Maybe also use arch type in the name

npalm (changes requested):
Tested the setup

- [ ] deploy default -> migrate to PR 1440
- [ ] uprade previous step to custom AMI failed

The instance is booting but some errors during loading, zee below

```
Configuration validation first phase succeeded
/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent -schematest -config /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml
Configuration validation second phase succeeded
Configuration validation succeeded
Redirecting to /bin/systemctl stop amazon-cloudwatch-agent.service
Redirecting to /bin/systemctl restart amazon-cloudwatch-agent.service
Get GH Runner token from AWS SSM
Delete GH Runner token from AWS SSM

An error occurred (ParameterNotFound) when calling the DeleteParameter operation:
```

I see the launch templates still contains the userdate to install the runner.

npalm:
> Tested the setup
> 
> * [x]  deploy default -> migrate to PR 1440
> * [ ]  upgrade previous step to custom AMI failed (on user-data)
> 
> The instance is booting but some errors during loading, in user_data
> 
> ```
> Configuration validation first phase succeeded
> /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent -schematest -config /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml
> Configuration validation second phase succeeded
> Configuration validation succeeded
> Redirecting to /bin/systemctl stop amazon-cloudwatch-agent.service
> Redirecting to /bin/systemctl restart amazon-cloudwatch-agent.service
> Get GH Runner token from AWS SSM
> Delete GH Runner token from AWS SSM
> 
> An error occurred (ParameterNotFound) when calling the DeleteParameter operation:
> ```
> 
> I see the launch templates still contains the userdate to install the runner. After a few minutes all my tests jobs succeeded.

Problem seems caused by I did not disabled the userdata via
```
userdata_enabled = false
```
So would suggest also mention this explicit in the readme of the images

modules/runners/variables.tf
```diff
@@ -292,6 +298,12 @@ variable "runner_log_files" {
...
+      "log_group_name" : "runner-startup",
```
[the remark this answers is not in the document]
npalm: You should add the new log_file to the end of het list, to avoid conflicts in an upgrade

modules/runners/policies/instance-describe-tags-policy.json
ScottGuymer: As discussed the above appears to work with some permissions and not others. @npalm tested with terminate instance permissions. 

I tested this condition with describe tags and found it to have no effect on my ability to query tags of other instances.

ScottGuymer: Verdict is to remove this for the describe tags policy as it has no effect.

examples/prebuilt/main.tf
ScottGuymer: fixed.

main.tf
ScottGuymer: I don't think this is possible without a breaking change.

Currently if this value is not set then the default userdata is used. 
If it is set then we try to load the template from the location in this value. If you specify an empty string it fails to load the file and errors
I could update the logic to distinguish between 
- null (which means use default userdata) 
- empty string (which means don't use userdata at all) 
- value (assume this is a file and try to load it)

But I'm not sure that using the difference between empty string and null as a trigger for this is clear enough..

WDYT?

examples/prebuilt/main.tf
ScottGuymer: set to 5

images/linux-amzn2/github_agent.linux.pkr.hcl
ScottGuymer: added

modules/runners/variables.tf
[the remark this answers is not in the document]
ScottGuymer: Done..

Not sure i like it being in both places. I think this is a TF limitation though. Will test removing and setting outer value to null.

main.tf
ScottGuymer: As discussed it makes sense to keep the explicit `userdata_enabled` to turn this on and of..

In addition i have renamed it to `enabled_userdata`

modules/runners/variables.tf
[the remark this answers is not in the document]
ScottGuymer: Tested this and it simply passes `null` from the wrapper module into the sub-module, The submodule then uses `null` as the value and ignores the default. 

Which makes sense even if it might not be what is wanted.

It would be nice if there was some option to "default on null" or rather than just when not specified.

npalm (approved):
Tested, all good!

PR #1454, opened 2021-12-01
chore: Ensure mockes are always loaded so tests can run in isolation

npalm:
## Problem
Test set for GH public in module runners, `src/scale-runners/scale-up.test.ts` cannot run in isolation. As full test suite all tests are PASSING, running a test in isolation for the set GH Public results in failures. In CI we run the full set.

## Solution
Problem is that several mocks are only loaded for GHES cases, which are loaded first. Moving the mockes one level up (suite level), ensures mocks are loaded always.


PatrickMennen (approved):
LGTM

PR #1476, opened 2021-12-02
Feat/windows support

ScottGuymer:
Merge the code that @RichiCoder1 created for windows runners into the develop branch so it can be released.

Beyond integrating this with the current develop I have made some minor changes.

Still need to update the docs etc.


main.tf
```diff
@@ -5,9 +5,6 @@ locals {
 
   s3_action_runner_url = "s3://${module.runner_binaries.bucket.id}/${module.runner_binaries.runner_distribution_object_key}"
   runner_architecture  = substr(var.instance_type, 0, 2) == "a1" || substr(var.instance_type, 1, 2) == "6g" ? "arm64" : "x64"
-
-  ami_filter = length(var.ami_filter) > 0 ? var.ami_filter : local.runner_architecture == "arm64" ? { name = ["amzn2-ami-hvm-2*-arm64-gp2"] } : { name = ["amzn2-ami-hvm-2.*-x86_64-ebs"] }
```
ScottGuymer, line 9: This is now quite complex logic so and was duplicated in `/modules/runner/main.tf` so i have removed it from here and just kept this logic in the runner module

modules/runners/variables.tf
```diff
@@ -72,10 +78,7 @@ variable "instance_types" {
 variable "ami_filter" {
   description = "Map of lists used to create the AMI filter for the action runner AMI."
   type        = map(list(string))
-
-  default = {
-    name = ["amzn2-ami-hvm-2.*-x86_64-ebs"]
-  }
+  default     = null
```
ScottGuymer, line 81: Now null so we can calculate values based on the OS later

modules/runners/scale-down.tf
```diff
@@ -18,7 +25,7 @@ resource "aws_lambda_function" "scale_down" {
       GHES_URL                             = var.ghes_url
       LOG_LEVEL                            = var.log_level
       LOG_TYPE                             = var.log_type
-      MINIMUM_RUNNING_TIME_IN_MINUTES      = var.minimum_running_time_in_minutes
+      MINIMUM_RUNNING_TIME_IN_MINUTES      = coalesce(var.minimum_running_time_in_minutes, local.min_runtime_defaults[var.runner_os])
```
ScottGuymer, line 28: The default minimum time needs to be longer for windows as it takes a lot longer to boot.

modules/runners/templates/start-runner.sh
```diff
@@ -37,11 +37,11 @@ fi
 
 ## Configure the runner
 
-echo "Get GH Runner token from AWS SSM"
+echo "Get GH Runner config from AWS SSM"
```
ScottGuymer, line 40: just some tidy up to make the logs consistent between win/linux

modules/runners/templates/start-runner.ps1
```diff
@@ -0,0 +1,97 @@
...
+    if ((Get-LocalGroup -Name "$group" -ErrorAction Ignore) -and
+        !(Get-LocalGroupMember -Group "$group" -Member $username -ErrorAction Ignore)) {
+        Add-LocalGroupMember -Group "$group" -Member $username
+        Write-Host "Added $username to $group group"
+    }
+}
+
+# Disable User Access Control (UAC)
+# TODO investigate if this is needed or if its overkill
+Set-ItemProperty HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Policies\System -Name ConsentPromptBehaviorAdmin -Value 0 -Force
```
ScottGuymer, line 82: I'm not sure if this is strictly needed, but it already existed and I haven't tested without it,

ScottGuymer:
@npalm  @RichiCoder1 this is ready for a review now!

modules/runners/variables.tf
```diff
@@ -136,7 +144,7 @@ variable "scale_down_schedule_expression" {
 variable "minimum_running_time_in_minutes" {
   description = "The time an ec2 action runner should be running at minimum before terminated if non busy."
```
npalm, line 145:
```suggestion
  description = "The time an ec2 action runner should be running at minimum before terminated if non busy. If not set the default is calculated based on the OS."
```

modules/runners/templates/start-runner.ps1
npalm: Suggest to make an issue and refer here to the issue

modules/runners/variables.tf
npalm: Should we not add what more detail, for example for linux we use amazon linux

npalm (changes requested):
Looks good to me, a few comments

npalm:
Tested windows. Works all fine, but find another weird problem. My events got declined by the lambda with a 401. Seems that secrets in github are limited. Their own example shows an hex generated value. Can you update the examples as follow 

```hcl
resource "random_id" "random" {
  byte_length = 16
}
```

Next update ref and the output tf.

npalm:
Tesated

- [ ] linux (default)
- [ ] windows

ScottGuymer:
> Tested windows. Works all fine, but find another weird problem. My events got declined by the lambda with a 401. Seems that secrets in github are limited. Their own example shows an hex generated value. Can you update the examples as follow
> 
> ```terraform
> resource "random_id" "random" {
>   byte_length = 16
> }
> ```
> 
> Next update ref and the output tf.

This may well have been the issue i was seeing the other day. I was seeing the same 401 issues. I tainted the secret and then it worked.

npalm:
> > Tested windows. Works all fine, but find another weird problem. My events got declined by the lambda with a 401. Seems that secrets in github are limited. Their own example shows an hex generated value. Can you update the examples as follow
> > ```terraform
> > resource "random_id" "random" {
> >   byte_length = 16
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Next update ref and the output tf.
> 
> This may well have been the issue i was seeing the other day. I was seeing the same 401 issues. I tainted the secret and then it worked.

No specification, but this example I found on github https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks#setting-your-secret-token

ScottGuymer:
Yeah, I can't seem to find a spec either. But all the examples show the secret converted to hex rather than just a plain string. 

Terraform doesn't seem to have a native string to hex conversion function (it does have integer to hex via `format()`). I think `random_id` with the hex output is the best we are going to get.

npalm:
> Yeah, I can't seem to find a spec either. But all the examples show the secret converted to hex rather than just a plain string.
> 
> Terraform doesn't seem to have a native string to hex conversion function (it does have integer to hex via `format()`). I think `random_id` with the hex output is the best we are going to get.

Maybe set the byte length a bit longer

examples/windows/README.md
```diff
@@ -0,0 +1,26 @@
+# Action runners deployment windows example
+
+This module shows how to create GitHub action runners using an Windows Runners. Lambda release will be downloaded from GitHub.
+
+## Usages
+
+Steps for the full setup, such as creating a GitHub app can be found in the root module's [README](../../README.md). First, download the Lambda releases from GitHub. Alternatively you can build the lambdas locally with Node or Docker, for which there is a build script available at `<root>/.ci/build.sh`. In the `main.tf` you can remove the location of the lambda zip files, the default location will work in this case.
```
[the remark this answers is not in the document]
ScottGuymer: I have set it to `false` to match the other examples.

modules/runners/variables.tf
ScottGuymer: I have added some hints on the defaults

modules/runners/templates/start-runner.ps1
ScottGuymer: done and adde dlink

npalm (approved):

ScottGuymer:
🎉

RichiCoder1:
🥳

PR #1525, opened 2021-12-18
feat: Add windows example ami

maths22:
This PR adds a packer config for building a Windows AMI based on 2019 core.
The packer config could presumably be reused to build on top of any other
Windows AMI, though I haven't tested it.  The example project for prebuilt
has been updated to allow using a windows AMI as well (mostly to show which
knobs need to be turned).

This has been tested against a personal aws account, but only with a super-
basic action that checks out the repo and does nothing else.


images/windows-core-2019/bootstrap_win.txt
```diff
@@ -0,0 +1,37 @@
...
+
+# Create a self-signed certificate to let ssl work
+$Cert = New-SelfSignedCertificate -CertstoreLocation Cert:\LocalMachine\My -DnsName "packer"
+New-Item -Path WSMan:\LocalHost\Listener -Transport HTTPS -Address * -CertificateThumbPrint $Cert.Thumbprint -Force
+
+# WinRM
+write-output "Setting up WinRM"
+write-host "(host) setting up WinRM"
+
+cmd.exe /c winrm quickconfig -q
```
ScottGuymer, line 22: I think you can do this all in PowerShell but I can't find a good example of it being done non-interactively.

images/windows-core-2019/github_agent.windows.pkr.hcl
```diff
@@ -0,0 +1,64 @@
...
+}
+
+build {
+  name = "githubactions-runner"
+  sources = [
+    "source.amazon-ebs.githubrunner"
+  ]
+
+  provisioner "file" {
+    content = templatefile("../../modules/runners/templates/start-runner.ps1", {})
```
ScottGuymer, line 55: The `start-runner.ps1` from the runners template folder needs a wrapper script to tell it where to log its output.  When its used in the runners templates its included in the user-data script which handles the logging output.

You can look at start-runner.sh in the linux image for an example. Something like this

```powershell
Start-Transcript -Path "C:\UserData.log" -Append
${start_runner}
Stop-Transcript
```

images/windows-core-2019/github_agent.windows.pkr.hcl
```diff
@@ -0,0 +1,64 @@
...
+}
+
+source "amazon-ebs" "githubrunner" {
+  ami_name      = "github-runner-windows-core-2019-${formatdate("YYYYMMDDhhmm", timestamp())}"
+  communicator  = "winrm"
+  instance_type = "t3a.medium"
+  region        = var.region
+  source_ami_filter {
+    filters = {
+      name                = "Windows_Server-2019-English-Core-Base-*"
```
ScottGuymer, line 29: In the Linux image we include docker as standard, we should do this for windows too. We should either choose the core-containers base AMI or configure docker in the windows-provisioner.ps1.

ScottGuymer:
Looks good.

Last thing I think needs to be added it to make the GH workflow a matrix job to run over the different images in the folder.

https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/.github/workflows/packer-build.yml#L23

ScottGuymer:
Looks like the build is failing on the formatting in `github_agent.windows.pkr.hcl`

pbronneberg:
Great! looking forward to using the Windows Build agents

ScottGuymer (approved):
LGTM

ScottGuymer:
@npalm anything else to add?

npalm (approved):
LGTM

PR #1537, opened 2021-12-20
fix: add --preserve-env to start-runner.sh to enable RUNNER_ALLOW_RUNASROOT

### LINKED ISSUE #1524: RUNNER_ALLOW_RUNASROOT variable is not set during start-runner operations
During instance startup when trying to launch the runner as root, the following logs are produced and the runner fails to launch. 

```
  | 2021-12-16T12:34:20.628-06:00 | Dec 16 18:34:19 ip-10-100-76-217 user-data: Delete GH Runner token from AWS SSM
  | 2021-12-16T12:34:20.628-06:00 | Dec 16 18:34:20 ip-10-100-76-217 user-data: Configure GH Runner as user root
  | 2021-12-16T12:34:20.628-06:00 | Dec 16 18:34:20 ip-10-100-76-217 systemd: Created slice User Slice of root.
  | 2021-12-16T12:34:20.628-06:00 | Dec 16 18:34:20 ip-10-100-76-217 systemd: Started Session c1 of user root.
  | 2021-12-16T12:34:20.628-06:00 | Dec 16 18:34:20 ip-10-100-76-217 user-data: Must not run with sudo
```

The logs never reach "Starting the runner as user" as expected so it seems that the `./config.sh` call is failing. To hackfix this, I'm approaching it like so:

```shell
...

if [[ "$run_as" == "root" ]]; then
    run_as_root_flag='RUNNER_ALLOW_RUNASROOT=1'
fi

echo "Configure GH Runner as user $run_as"
sudo -u "$run_as" -- $run_as_root_flag ./config.sh --unattended --name "$instance_id" --work "_work" $${config}

## Start the runner
echo "Starting runner after $(awk '{print int($1/3600)":"int(($1%3600)/60)":"int($1%60)}' /proc/uptime)"
echo "Starting the runner as user $run_as"

if [[ $agent_mode = "ephemeral" ]]; then  
  echo "Starting the runner in ephemeral mode"
  sudo -u "$run_as" -- $run_as_root_flag ./run.sh
  echo "Runner has finished"
  
  echo "Stopping cloudwatch service"
  service awslogsd stop
  echo "Terminating instance"
  aws ec2 terminate-instances --instance-ids "$instance_id" --region "$region"
else 
  echo "Installing the runner as a service"
  ./svc.sh install "$run_as"
  echo "Starting the runner in persistent mode"
  ./svc.sh start
fi
```

Please let me know if there's a config I'm missing or there's a better way to accomplish this, otherwise I can put in a PR with this.

inhumantsar:
Fixes #1524

### DIFF
--- a/modules/runners/templates/start-runner.sh
+++ b/modules/runners/templates/start-runner.sh
@@ -54,8 +54,12 @@ if [ -z "$run_as" ]; then
     run_as="ec2-user"
 fi
 
+if [[ "$run_as" == "root" ]]; then
+    export RUNNER_ALLOW_RUNASROOT=1
+fi
+
 echo "Configure GH Runner as user $run_as"
-sudo -u "$run_as" -- ./config.sh --unattended --name "$instance_id" --work "_work" $${config}
+sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh --unattended --name "$instance_id" --work "_work" $${config}
 
 ## Start the runner
 echo "Starting runner after $(awk '{print int($1/3600)":"int(($1%3600)/60)":"int($1%60)}' /proc/uptime)"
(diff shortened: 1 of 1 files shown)


modules/runners/templates/start-runner.sh
```diff
@@ -54,16 +54,20 @@ if [ -z "$run_as" ]; then
     run_as="ec2-user"
 fi
 
+if [[ "$run_as" == "root" ]]; then
```
ScottGuymer, line 57: This is already exported (and not used) in the `install-runner.sh` file without a condition.

I think the condition here is fine and it belongs in this file. Will remove it from the other file in a different PR.

ScottGuymer (approved):

PR #1539, opened 2021-12-21
Ubuntu exemple with on-premise GHES using network http proxy

axel3rd:
PoC about #1303

Requires #1256

Result can be found [here](https://github.com/GroupePSA/terraform-aws-github-runner/tree/exemple-ubuntu-vpc-routed-proxy/examples/ubuntu-ghes-vpc-routed-proxy)


axel3rd:
Set to draft: AWS credentials & perhaps proxy are useless in user-data template.

axel3rd:
> AWS credentials are useless in user-data template.
> userdata_pre_install / userdata_post_install could probably be used to pass company + admin PAT temporary

As draft until #1256 review/merge.

ScottGuymer:
I think it might be useful if this example deployed some kind of proxy that could be used to fully test this config without having to have a proxy configured elsewhere.

axel3rd:
> I think it might be useful if this example deployed some kind of proxy that could be used to fully test this config without having to have a proxy configured elsewhere.

Agree with the target, but not trivial to achieve 😕.
A VPC should probably be customized (including the proxy), which has direct connection with the customer GHES (could depend of network layers in place).

ScottGuymer:
Would it work to just have a straight proxy to the internet and test against GH SaaS?

axel3rd:
> Would it work to just have a straight proxy to the internet and test against GH SaaS?

It is a pure network problem. If on one VPC (which host Lambda [scaleup|scaledown|pool]-runners) we can allow GH SaaS URL but deny AWS API URLs (to "enforce" to use the proxy for runners), I guess yes 🤞.

PR #1542, opened 2021-12-22
fix: Update filter to match with new GitHub release assets.

npalm:
## Problem
The latest release of GitHub runners contains a lot more artifacts. The syncer Lambda is downloading the GitHub runner release based on a prefix with architecture and os. Per arch and os this leads now to mutliple matches wich cause the error "Cannot find GitHub release asset". 

## Solution
Precizer match to find the release asset


npalm:
fix #1542

ScottGuymer (dismissed):
LGTM :shipit:

ScottGuymer (approved):
:shipit:

Issue #1400, opened 2021-11-09
Support for EC2 Spot Best practices - Diversification of instances using ABS


ruecarlo:
Currently the lambda in charge of scaling up nodes (including Spot configurations), uses the RunInstance API to create a Spot instance. Spot instances as spare capacity might be limited. Best approach when using Spot instance is to diversify across a set of instances that qualify for the workload and use one of the API that allows for that diversification.

The suggestion here is to change the `RunInstance` call and instead use the Drop In replacement API for [EC2 Fleet in instant and the Spot Capacity-Optimized allocation strategy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instant-fleet.html). EC2 Fleet allows diversification and still provides a synchronous API that adhere to Spot best practices providing the spot instance types that are selected to minimise the frequency of interruptions for the workload.  More examples [here](https://ec2spotworkshops.com/launching_ec2_spot_instances/ec2_fleet.html).

Another thing for consideration in the implementation is to use the newly released [attribute based instance selection](https://aws.amazon.com/blogs/aws/new-attribute-based-instance-type-selection-for-ec2-auto-scaling-and-ec2-fleet/)

npalm:
I did a short experiment, see [gist](https://gist.github.com/npalm/f9491e661cbf6aabc99b1656e3cf0127). It creates the instances but caues 2 issues

- it also tries to launch instances in the default vpc, which fails since the required security group is not available in the default VPC. Not figured out, how to avoid possible creation. The result of the createFleet contains 1 instance in a valid subnet. But also the error mentioned before.
- Creating the SSM properties does not work, did not investigate at all yet

npalm:
With replacing the runInstance API call by createFleet also the dynamic launch templates can be removed. Finding a right spot instances (or on demand) should be moved to this API call. https://github.com/philips-labs/terraform-aws-github-runner/blob/dbba705038828c86f6f5adef18f7a7a35643c359/modules/runners/main.tf#L57

npalm:
Currently we specify in the launch template via the option [`market_type`](https://github.com/philips-labs/terraform-aws-github-runner/blob/dbba705038828c86f6f5adef18f7a7a35643c359/modules/runners/main.tf#L96) whenever we create a spot or on-demand. Maybe we should move this logic also to the API call for creating a fleet.

npalm:
[...]

PR #1556, opened 2021-12-23
feat: Replace run instance API by create fleet API

npalm:
## Description

As suggested in #1400 it is a best practice to use the fleet API for creating instances. This PR will replace the way we create instances by using the create fleet API instead of run instances. Moving to the create fleet API instead of the run isntances API will remove the compleity of the lambda to retry instance creation once failed. 

## Implementation

The lambda will take a list of instance types, lift of subnets. And wil create based on the the opstio target capacity type a spot (default) or an on-demand instance.  The creat fleet API will choose the instance type and subnet based on tthe specified criteria.


npalm:
@ruecarlo would you have a bit of time to check the API call for create fleet? Which options should we set to cover scenarios like:
- only spot
- only on-demand
- spot prefered, on-demand as fall back

npalm:
No clue why the diff is showing the moved files as new files

```
 rename modules/runners/lambdas/runners/src/{scale-runners => aws}/runners.test.ts (99%)
 rename modules/runners/lambdas/runners/src/{scale-runners => aws}/runners.ts (98%)
 rename modules/runners/lambdas/runners/src/{scale-runners => aws}/ssm.test.ts (100%)
 rename modules/runners/lambdas/runners/src/{scale-runners => aws}/ssm.ts (100%)
 rename modules/runners/lambdas/runners/src/{scale-runners => gh-auth}/gh-auth.test.ts (98%)
 rename modules/runners/lambdas/runners/src/{scale-runners => gh-auth}/gh-auth.ts (95%)
 rename modules/runners/lambdas/runners/src/{scale-runners => }/logger.ts (100%)
 rename modules/runners/lambdas/runners/src/{scale-runners => }/modules.d.ts (100%)
```

README.md
```diff
@@ -87,6 +87,7 @@ To be able to support a number of use-cases the module has quite a lot configura
 - Linux vs Windows. you can configure the os types linux and win. Linux will be used by default.
 - Re-use vs Ephemeral. By default runners are re-used for till detected idle, once idle they will be removed from the pool. To improve security we are introducing ephemeral runners. Those runners are only used for one job. Ephemeral runners are only working in combination with the workflow job event. We also suggest to use a pre-build AMI to improve the start time of jobs.
 - GitHub cloud vs GitHub enterprise server (GHES). The runner support GitHub cloud as well GitHub enterprise service. For GHES we rely on our community to test and support. We have no possibility to test ourselves on GHES.
+- Spot vs on-demand. The runners using either the EC2 spot or on-demand life cycle. Runners will be created via the AWC [CreateFLeet API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html). THe module (scale up lambda) will request via the create fleet API aan instance in one of the subnets and matching one of the specified instances types.
```
ScottGuymer, line 90:
```suggestion
- Spot vs on-demand. The runners using either the EC2 spot or on-demand life cycle. Runners will be created via the AWS [CreateFleet API](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html). The module (scale up lambda) will request an instance via the create fleet API in one of the subnets and matching one of the specified instance types.
```

README.md
```diff
@@ -347,15 +348,7 @@ The following sub modules are optional and are provided as example or utility:
...
-When not using the top-level module and specifying an `a1`, `t4g` or `*6g*` (6th-gen Graviton2) `instance_type`, the `runner-binaries-syncer` and `runners` submodules need to be configured appropriately for pulling the ARM64 GitHub action runner binary and leveraging the arm64 AMI for the runners.
-
-When configuring `runner-binaries-syncer`
-
-- _runner_architecture_ - set to `arm64`, defaults to `x64`
-
-When configuring `runners`
-
-- _ami_filter_ - set to `["amzn2-ami-hvm-2*-arm64-gp2"]`, defaults to `["amzn2-ami-hvm-2.*-x86_64-ebs"]`
+When using the top level module configure `runner_architecture = arm64` and insure the list of `instance_types` matches. When not using the top-level ensure the bot properties are set on the submodules.
```
ScottGuymer, line 351:
```suggestion
When using the top-level module configure `runner_architecture = arm64` and ensure the list of `instance_types` matches. When not using the top-level ensure the bot properties are set on the submodules. 
```

main.tf
```diff
@@ -5,7 +5,6 @@ locals {
   })
 
   s3_action_runner_url = "s3://${module.runner_binaries.bucket.id}/${module.runner_binaries.runner_distribution_object_key}"
-  runner_architecture  = substr(var.instance_type, 0, 2) == "a1" || substr(var.instance_type, 0, 3) == "t4g" || substr(var.instance_type, 1, 2) == "6g" ? "arm64" : "x64"
```
ScottGuymer, line 8: I like that we are removing the complexity here.

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -0,0 +1,196 @@
...
+      'MaxSpotFleetRequestCountExceeded',
+    ];
+
+    if (errors.some((e) => scaleErrors.includes(e))) {
+      logger.warn('Create fleet failed, ScaleError will be thrown to trigger retry for ephemeral runners.');
+      logger.debug('Create fleet failed.', fleet.Errors);
+      throw new ScaleError('Failed to create instance, create fleet failed.');
+    } else {
+      logger.warn('Create fleet failed, error not recognized as scaling error.', fleet.Errors);
+      throw Error('Create flee failed, no instance created.');
```
ScottGuymer, line 180:
```suggestion
      throw Error('Create fleet failed, no instance created.');
```

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -0,0 +1,196 @@
...
+      'RequestLimitExceeded',
+      'ResourceLimitExceeded',
+      'MaxSpotInstanceCountExceeded',
+      'MaxSpotFleetRequestCountExceeded',
+    ];
+
+    if (errors.some((e) => scaleErrors.includes(e))) {
+      logger.warn('Create fleet failed, ScaleError will be thrown to trigger retry for ephemeral runners.');
+      logger.debug('Create fleet failed.', fleet.Errors);
+      throw new ScaleError('Failed to create instance, create fleet failed.');
```
ScottGuymer, line 177: So this will toss it back on the queue to be retried?

modules/runners/lambdas/runners/src/gh-auth/gh-auth.ts
```diff
@@ -32,7 +32,7 @@ export async function createGithubAppAuth(
 ): Promise<AppAuthentication> {
   const auth = await createAuth(installationId, ghesApiUrl);
   const appAuthOptions: AppAuthOptions = { type: 'app' };
-  return await auth(appAuthOptions);
+  return auth(appAuthOptions);
```
ScottGuymer, line 35: I'm guessing the promise returned here is now awaited by the caller?

modules/runners/scale-up.tf
```diff
@@ -29,6 +29,10 @@ resource "aws_lambda_function" "scale_up" {
       RUNNERS_MAXIMUM_COUNT                = var.runners_maximum_count
       SUBNET_IDS                           = join(",", var.subnet_ids)
       ENABLE_EPHEMERAL_RUNNERS             = var.enable_ephemeral_runners
+      INSTANCE_TYPES                       = join(",", var.instance_types)
+      INSTANCE_TARGET_CAPACITY_TYPE        = var.instance_targeet_capacity_type
```
ScottGuymer, line 33:
```suggestion
      INSTANCE_TARGET_CAPACITY_TYPE        = var.instance_target_capacity_type
```

This will likely cause issues elsewhere.

modules/runners/variables.tf
```diff
@@ -52,9 +52,48 @@ variable "block_device_mappings" {
...
+  type        = string
+  default     = null
+
+  validation {
+    condition     = anytrue([var.market_options == null])
+    error_message = "Deprecated, replaced by `instance_targeet_capacity_type`."
+  }
+}
+
+variable "instance_targeet_capacity_type" {
```
ScottGuymer, line 65:
```suggestion
variable "instance_target_capacity_type" {
```

modules/runners/variables.tf
```diff
@@ -52,9 +52,48 @@ variable "block_device_mappings" {
...
+    error_message = "The instance target capacity should be either spot or on-demand."
+  }
+}
+
+variable "instance_allocation_strategy" {
+  description = "The allocation strategy for spot instances. AWS recommends to use `capacity-optimized` however the AWS default is `lowest-price`."
+  type        = string
+  default     = "lowest-price"
+  validation {
+    condition = anytrue([
```
ScottGuymer, line 83: You can simplify this validation like this 

```
  validation {
    condition     = contains(["lowest-price", "diversified", "etc"], var.instance_allocation_strategy)
    error_message = "The instance allocation strategy does not match the allowed values."
  }
```

modules/runners/variables.tf
```diff
@@ -52,9 +52,48 @@ variable "block_device_mappings" {
...
+      var.instance_allocation_strategy == "lowest-price",
+      var.instance_allocation_strategy == "diversified",
+      var.instance_allocation_strategy == "capacity-optimized",
+      var.instance_allocation_strategy == "capacity-optimized-prioritized",
+    ])
+    error_message = "The instance allocation strategy does not matched the allowed values."
```
ScottGuymer, line 89:
```suggestion
    error_message = "The instance allocation strategy does not match the allowed values."
```

modules/runners/variables.tf
```diff
@@ -52,9 +52,48 @@ variable "block_device_mappings" {
...
+  }
+}
+
+variable "instance_max_spot_price" {
+  description = "Max price price for spot intances per hour. This variable will be passed to the create fleet as max spot price for the fleet."
```
ScottGuymer, line 94: What happens with the default `null`? is no max specified then?

variables.tf
```diff
@@ -340,9 +334,48 @@ variable "runner_additional_security_group_ids" {
...
+    error_message = "Deprecated, replaced by `instance_targeet_capacity_type`."
+  }
+}
+
+variable "instance_targeet_capacity_type" {
+  description = "Default lifecyle used runner instances, can be either `spot` or `on-demand`."
   type        = string
   default     = "spot"
+  validation {
+    condition = anytrue([
```
ScottGuymer, line 352: Same as above this can be simplified with `contains`

variables.tf
```diff
@@ -340,9 +334,48 @@ variable "runner_additional_security_group_ids" {
...
+  default     = null
+
+  validation {
+    condition     = anytrue([var.market_options == null])
+    error_message = "Deprecated, replaced by `instance_targeet_capacity_type`."
+  }
+}
+
+variable "instance_targeet_capacity_type" {
+  description = "Default lifecyle used runner instances, can be either `spot` or `on-demand`."
```
ScottGuymer, line 348:
```suggestion
  description = "Default lifecycle used for runner instances, can be either `spot` or `on-demand`."
```

variables.tf
```diff
@@ -340,9 +334,48 @@ variable "runner_additional_security_group_ids" {
...
+    error_message = "The instance target capacity should be either spot or on-demand."
+  }
+}
+
+variable "instance_allocation_strategy" {
+  description = "The allocation strategy for spot instances. AWS recommends to use `capacity-optimized` however the AWS default is `lowest-price`."
+  type        = string
+  default     = "lowest-price"
+  validation {
+    condition = anytrue([
```
ScottGuymer, line 365: This can be simplified with `contains`

variables.tf
```diff
@@ -351,10 +384,21 @@ variable "volume_size" {
...
+  default     = null
+
+  validation {
+    condition     = anytrue([var.instance_type == null])
+    error_message = "Deprecated, replaced by `instance_types`."
+  }
+}
+
 variable "instance_types" {
   description = "List of instance types for the action runner. Defaults are based on runner_os (amzn2 for linux and Windows Server Core for win)."
```
ScottGuymer, line 399: I dont think the description here makes sense. Maybe a previous copy paste issue..

variables.tf
```diff
@@ -500,3 +544,18 @@ variable "redrive_build_queue" {
...
   }
 }
+
+
+variable "runner_architecture" {
+  description = "The platform architecture of the runner instance_type."
+  type        = string
+  default     = "x64"
+  validation {
+    condition = anytrue([
```
ScottGuymer, line 554: This can be simplified with `contains`

modules/runners/variables.tf
```diff
@@ -52,9 +52,48 @@ variable "block_device_mappings" {
...
+    error_message = "Deprecated, replaced by `instance_targeet_capacity_type`."
+  }
+}
+
+variable "instance_targeet_capacity_type" {
+  description = "Default lifecyle used runner instances, can be either `spot` or `on-demand`."
   type        = string
   default     = "spot"
+  validation {
+    condition = anytrue([
```
ScottGuymer, line 70: This can be simplified with `contains`

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -88,29 +180,29 @@ export async function scaleUp(eventSource: string, payload: ActionRequestMessage
...
-      await createRunnerLoop({
-        environment,
-        runnerServiceConfig: enableOrgLevel
-          ? `--url ${configBaseUrl}/${payload.repositoryOwner} ${runnerArgs}${runnerGroupArgument}`.trim()
-          : `--url ${configBaseUrl}/${payload.repositoryOwner}/${payload.repositoryName} ${runnerArgs}`.trim(),
-        runnerOwner,
-        runnerType,
-      });
+
+      await createRunners(
```
ScottGuymer, line 184: This is so much simpler 🎉

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -15,6 +15,109 @@ export interface ActionRequestMessage {
...
+        await githubClient.apps.getRepoInstallation({
+          owner: payload.repositoryOwner,
+          repo: payload.repositoryName,
+        })
+      ).data.id;
+}
+
+async function isJobQueued(githubInstallationClient: Octokit, payload: ActionRequestMessage): Promise<boolean> {
+  let isQueued = false;
+  if (payload.eventType === 'workflow_job') {
```
ScottGuymer, line 81: The `workflow_job` string is used a few times. Maybe extract it out into a const so that it is safer to use.

There are other strings here that could have the same treatment like `Org` and `check_run`

ScottGuymer (changes requested):
This is looking a lot neater and less complex

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
npalm: Correct, but is used as a kind of enum. In type script it is called literal string types see https://www.typescriptlang.org/docs/handbook/literal-types.html

modules/runners/variables.tf
npalm: Good catch, it works but it is unclear how null is handled. AWS docs https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_SpotOptionsRequest.html does not give any details on the validtiy of the string. Will do some checks, and or make it stricter.

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
ScottGuymer: ah ok cool

modules/runners/lambdas/runners/src/aws/runners.ts
npalm: Yes for ephemeral rnners only

npalm:
The PR is ready for review, however there are 2 topics we could got some extra help.

1. When sending a create fleet request with type instant. We handling a [list of specific errors](https://github.com/philips-labs/terraform-aws-github-runner/blob/51a0070e82a3d1989f856e04bff8963d5addc5f1/modules/runners/lambdas/runners/src/aws/runners.ts#L164).  For the errors in this list we think it make sense to retry later, since the casue is a rate limit, or limit in general. Are we missing errors, or are there errors that should not trigger a retry?
2. With te create fleet api we create typically 1 instance, either spot or on-emand. But we are wondering how we can ensure we got an instant if there is no spot capacity?

modules/runners/variables.tf
ScottGuymer: Did you uncover anything on how this is handled?

ScottGuymer (approved):
LGTM :shipit:

PR #1561, opened 2021-12-27
Support t4g Graviton instance type

bliles:
Add t4g to the list of arm64 instance types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ To be able to support a number of use-cases the module has quite a lot configura
 
 #### ARM64 support via Graviton/Graviton2 instance-types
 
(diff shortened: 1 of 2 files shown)


npalm (commented):
@bliles thanks for your PR. We weill check the change asap. Be-aware that the variable `instance_type` is deprecated. I have #1556  open in which I remvoing the local runner_architecture. Feel free to comment on this PR in case you have any suggestion.

npalm (approved):

bliles:
> @bliles thanks for your PR. We weill check the change asap. Be-aware that the variable `instance_type` is deprecated. I have #1556 open in which I remvoing the local runner_architecture. Feel free to comment on this PR in case you have any suggestion.

That makes sense to me, it's probably more future-proof to just have a variable to define the runner architecture instead of inferring it from the instance type.

Issue #1560, opened 2021-12-24
Issue trying to use ubuntu based runners with 0.27.2


nparfait:
I was previously using version 0.18.1 of this module. I decided to upgrade to the latest 0.27.2. However came across an issue.

I'm trying to use ubuntu runner ( i have copied user-data.sh from examples/ubuntu/templates/ )
Using the defaults the runner was registering with GIthub.

I then tried using the flag
runner_as_root = true

this then registered the runner but caused me some permissions issues between actions/checkout and my steps running in a container.

Finally i tried updating SSM parameter xxx/runner/run-as
to: runners

and this was able to run successfully.

Looking here 
https://github.com/philips-labs/terraform-aws-github-runner/blob/v0.27.2/modules/runners/runner-config.tf#L4

the run as is either root or ec2-user. 
The ubuntu image doesn't have an ec2-user so not sure if this is a problem or not.

Can you please advise?

npalm:
Thanks for creating the issue, the ubuntu example is indeed broken. I have just created a fix in PR #1572

Issue #1562, opened 2021-12-27
The request include parameter that is not valid for this queue type.


maathor:
Hi,
I'm trying to implement this current stack. Into the application log (into github > app > advance I retrieve my attempts)
They are all failing with : 
`{"message":"Internal Server Error"}`

From cloudwatch, the webhook lambda I get:

```
    "message": "Value 4643128775 for parameter MessageGroupId is invalid. Reason: The request include parameter that is not valid for this queue type.",
    "code": "InvalidParameterValue",
    "name": "InvalidParameterValue"
```

I used the ubuntu example with the lastest lambdas version (v0.27.2).
From the app configuration side I set:

```
- Repository:
   Actions: Read-only (check for queued jobs)
   Checks: Read-only (receive events for new builds)
   Metadata: Read-only (default/required)
   Administration: Read & write (to register runner)

- Organization
     Self-hosted runners: Read & write (to register runner)

- In the "Permissions & Events" section and then "Subscribe to Events"
subsection, check "Workflow job"
```

I didn't any message into my SQS queue due to this failing webhook lambda.

npalm:
Looks like the the webhook lambda is including `MessageGroupId` in the SQS message. This value is only valid vor an FIFO queue. Can you check what type of queue is created?

maathor:
a standard one, yes, you're right ! seems legit !

npalm:
The variable `sqs_build_queue_fifo` should match the queue type. This variable is passed as input to the webhook lamda. You can check in the configuration tab in AWS lambda / webhook what value is used.

maathor:
yes, thank you so much. it works like a charm now !

PR #1479, opened 2021-12-04
fix install_config_runner -> install_runner

marekaf:
https://github.com/philips-labs/terraform-aws-github-runner/commit/060daac3568cd36f8b203d3f77f736df7aefb223#diff-e9624b388e62ca51cf1fe073eb0919588e8c36a1143ecdb49580996a89f13beb

https://github.com/philips-labs/terraform-aws-github-runner/commit/060daac3568cd36f8b203d3f77f736df7aefb223#r61361220

I believe this was forgotten to be changed in ubuntu example

Afrer upgrading I got for the ubuntu userdata.sh this:
```
│ Invalid value for "vars" parameter: vars map does not contain key "install_config_runner", referenced at ./templates/user-data.sh.tpl:74,3-24.
╵
```

PR #1572, opened 2021-12-28
fix: ubuntu example

npalm:
## Description

Due to refactoring in #1444 the ubuntu example got broken when running as non root user. 

## Solution

- Added variable `runner_run_as` to conrigure the desired run as user for the agent on the EC2 instance, with default 'ec2-user'. 
- Changed the default install dir to `/opt/action-runners/` instead of in the user home.

## Notes

closes: #1560

other potential related issues: #1479 #1562 #1524


examples/ubuntu/templates/user-data.sh
```diff
@@ -62,10 +62,12 @@ su -l $USER_NAME -c "systemctl --user enable docker"
 ${install_runner}
 
 # config runner for rootless docker
-cd /home/$USER_NAME/actions-runner/
+cd /opt/actions-runner/
```
ScottGuymer, line 65: This is a nice change it simplifies and centralises the install location.

ScottGuymer (approved):

Issue #519, opened 2021-02-06
Automate the creation of an offline runner


npalm:
The current approach requires we alwsys have an offline runner regirstered. Those offline runners are removed by github eacht 60 days. Therefore it would be convient to automate the process to keep 1 offline runner so we can scale back to 0.

### Register via ED2 instance
Use the same mechanism we use to spin up runners with an extra lambda that only execute the user data till the config step. And also ensure the ec2 instance is take down

### Reverse egineer github config process
Create lamda that is using github http calls based on reverse engineering, see https://github.com/actions/runner/issues/558

### Run the config in a lambda
Create a lambda that can exectue the config via a lambda layer

miked63017:
@npalm I've figured out most of the logic, urls, etc involved with registering a runner. I plan on making a python module to handle this using python 3x and requests. This will probably be this weekend when I'm off work.

If you would prefer to implement it yourself, I can share my notes I've gathered about the process before then. 

I honestly think I want to spend some more time tracing the entire process, so it could actually pull and perform workflows as well, but the new runner registration was a must have for me.

Basically the process I have found to work:
Get runner token via api using pat or app creds 
Post token to secretish endpoint, receive back json data with new secretish endpoint and a jwt
Use jwt to auth via auth bearer header, using newly revealed endpoint, can query existing "agents"(as they are called in the api), add a new one, or update an existing one.

There's obviously some more details involved, including creating an RSA key, and a bunch of headers. I haven't looked further than registration yet...

npalm:
@gertjanmaas I think you will like the comment above

mcaulifn:
@miked63017 let me know if you don't get to it. We are looking for this as well so I would have some time to work on it.

miked63017:
@npalm @mcaulifn here is a link, it's still pretty beta'ish and not well documented, but I guess we can say the same thing about the runners/actions API in general :-)

https://github.com/miked63017/pyghrunner

miked63017:
@mcaulifn in RE to the ghapi module it looks cool, but most of these calls are undocumented pieces of the api, and probably subject to change.

mcaulifn:
Overall looks like it should work. Are you planning on adding it to this repo?

npalm:
@gertjanmaas any opinion?

miked63017:
[...]

PR #1573, opened 2021-12-29
docs: Offline runners are not required anymore.

npalm:
Prevous jobs for self-hosted runners failing immediatly if no potential runner was available. This is changed, jobs will kept queued for 24 hours. This removed the need to create first an offline runner to allow scaling back to zero. (https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow)


npalm:
close: #519

ScottGuymer (approved):

PR #1577, opened 2021-12-31
feat: Add scheduled / pull based scaling for org level runners

npalm:
## Description

Thepool will be an optional feture (opt-in) and only available for org level runners. Thepool will remain a constant number of runners based on a cron expression (AWS event bridge).

- Only supporting Org level runners
- Supports multiple cron triggers, but only one size can be chosen for the pool

Todo:
- [x] Add Lambda
- [x] Update Terraform
- [x] Add unit tests
- [x] Update docs
- [x] Test

## Implementation notes

The required pool size is calculated as follow
1. Get the running instance for the module from EC2 based on the environment
2. Get the idle runners from GitHub
Intersection between 1 and 2 are the number of available runners to consume build. Diff in size of the the pool is the amount used for toppin gup.


README.md
```diff
@@ -250,6 +252,14 @@ module "runners" {
 
 ```
 
+### Simple Pool
+
+The module basically supports two options for keeping a pool of runners. One is via a simple pool which only supports org-level runners, the second option is [keeping runners idle](#idle-runners). 
+
+The simple pool is introduced in combination with the ephemeral runners and is primary meant to ensure if any event is unexpected dropped, and no runner was created the pool can pick up the job. THe pool is maintained by a lambda. Each time the lambda is triggered a check is preformed if the number of idler runners managed by the module are meeting the expected pool size. If not, the pool will be adjusted. Keep in mind that the scale down function is still active and will terminate instances that are detected to long as idle.
```
axel3rd, line 259: Minor typo (THe/the pool)

```suggestion
The simple pool is introduced in combination with the ephemeral runners and is primary meant to ensure if any event is unexpected dropped, and no runner was created the pool can pick up the job. The pool is maintained by a lambda. Each time the lambda is triggered a check is preformed if the number of idler runners managed by the module are meeting the expected pool size. If not, the pool will be adjusted. Keep in mind that the scale down function is still active and will terminate instances that are detected to long as idle.
```

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -57,7 +60,24 @@ export async function listEC2Runners(filters: ListRunnerFilters | undefined = un
       ec2Filters.push({ Name: `tag:Owner`, Values: [filters.runnerOwner] });
     }
   }
-  const runningInstances = await ec2.describeInstances({ Filters: ec2Filters }).promise();
+
+  //  const runningInstances = [];
```
ScottGuymer, line 64: Rogue comment/

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -57,7 +60,24 @@ export async function listEC2Runners(filters: ListRunnerFilters | undefined = un
...
+
+  const runners: RunnerList[] = [];
+  let nextToken = undefined;
```
ScottGuymer, line 67: Can you do this in typescript?

```suggestion
  let nextToken;
```

modules/runners/lambdas/runners/src/lambda.test.ts
```diff
@@ -76,22 +78,22 @@ describe('Test scale up lambda wrapper.', () => {
...
+    expect(await scaleUpHandler(sqsEvent, context)).resolves;
   });
 
   it('Scale should be rejected', async () => {
     const error = new ScaleError('some scale error');
     const mock = mocked(scaleUp);
 
     mock.mockRejectedValue(error);
-    await expect(scaleUpHandler(sqsEvent, context)).rejects.toThrow(error);
+    expect(scaleUpHandler(sqsEvent, context)).rejects.toThrow(error);
```
ScottGuymer, line 96: Does this not need an await too?

modules/runners/simple-pool.tf
```diff
@@ -0,0 +1,124 @@
+resource "aws_lambda_function" "simple_pool" {
+  count = length(var.simple_pool_config) == 0 ? 0 : 1
```
ScottGuymer, line 2: As everything in this file has a `count` you could add it a private module and have the count on the module declaration.

This would make all the code here simpler by removing the `[0]` in every reference.

modules/runners/lambdas/runners/src/simple-pool/simple-pool.ts
```diff
@@ -0,0 +1,91 @@
...
+import { listEC2Runners } from '../aws/runners';
+
+const logger = rootLogger.getChildLogger({ name: 'simple-pool' });
+
+export interface SimplePoolEvent {
+  simplePoolSize: number;
+}
+
+export async function adjust(event: SimplePoolEvent): Promise<void> {
+  logger.info(`Adjusting simple pool of size: ${event.simplePoolSize}`);
```
ScottGuymer, line 14: I don't think we are adjusting anything at this point.

```suggestion
  logger.info(`Checking current pool size against simple pool of size: ${event.simplePoolSize}`);
```

ScottGuymer (commented):
Looks quite neat. 

Some small comments

modules/runners/simple-pool.tf
npalm: Ture, but moving it into a module requires to duplicate a lot fo variables that needs to be set

modules/runners/lambdas/runners/src/lambda.test.ts
npalm: Not 100% sure, but looks like because of mocking that is not needed.

modules/runners/variables.tf
```diff
@@ -462,3 +462,30 @@ variable "enable_ephemeral_runners" {
...
 }
+
+variable "simple_pool_lambda_timeout" {
+  description = "Time out for the simple pool lambda lambda in seconds."
+  type        = number
+  default     = 60
+}
+
+variable "simple_pool_runner_owner" {
+  description = "The simple pool wil deploy runners to an GitHub org, set this value to the org to which you want the runners deployed. Repo level is not supported."
```
ScottGuymer, line 473:
```suggestion
  description = "The simple pool will deploy runners to a GitHub org, set this value to the GitHub org ID to which you want the runners deployed. Repo level is not supported."
```

modules/runners/simple-pool.tf
ScottGuymer: It's a balance between duplicating the variables and introducing complexity to the code and the possibility of future mistakes when extending this code.

npalm: refactored

ScottGuymer (approved):
:shipit:

modules/runners/lambdas/runners/src/lambda.test.ts
npalm: Updated tests according  https://jestjs.io/docs/tutorial-async

ScottGuymer (approved):

ScottGuymer (approved):

Issue #1530, opened 2021-12-20
Question about resuable workflows


teodor-daskalov:
Hi @npalm,

About 3 months ago we implemented your solution (v0.18.1) and we're really happy with it, because it provides us with a way to have dynamic EC2 instances as self-hosted runners and free up resources for other things from our own on-prem machines.

A colleague of ours is trying to implement the new feature of reusing workflows that was released last month:
https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

So far no success. The workflows that call a reusable workflow cannot trigger the spawning of new AWS runners.

Have you tried this on your end and could it work in theory with the version that we have on our end?

Thanks in advance for your support!

npalm:
Yes, we have the reusable workflows / templates working. Did you grant org level access to the repo containing the shared workflow?

teodor-daskalov:
@npalm thanks for your response. I think I found out the reason behind this, after I re-checked the webhook lambda's logs:

<img width="1273" alt="image" src="https://user-images.githubusercontent.com/86970293/146936161-3d77eae7-0406-4802-b969-ed04a1f9c586.png">

So what I fail to understand is why doesn't the lambda recognize the 3 default labels for any linux OS runner: 
`self-hosted`, `linux`, `x86` ?

In our Terraform config we have defined 3 custom labels: `aws-runner`, `ubuntu-18.04`, and `test`.
But the lambda rejects a request that has labels `self-hosted` + `linux` because the label `linux` doesn't seem to be taken into consideration.

Am I missing something or is this the expected behavior?

Of course, at this time it is an easy thing to fix, we just always have to target the exact set of labels defined or a specific part of them, without introducing an "unknown" to the lambda label.

npalm:
The webhook is checking if the labels in the event are matching the configured labels. So in your case the labels aws-runner, ubuntu-18.04, and test should be used in the workflow triggering the build. The predefined github labels are ignored in the check. You can also set `disable_check_wokflow_job_labels` which should disable the label check and accept any workflow job event to the webhook. 

The labels check is relative new, and was introduced together with the workflow job even.

teodor-daskalov:
Thanks a lot for the info! Btw, do you think all the configured labels have to be used in the workflow or one or two is enough?

npalm:
Actually, I doubt a bit about what to do with the labels. I even think to disable the check by default. So would suggest to siable them for now.

teodor-daskalov:
[...]

PR #1591, opened 2022-01-06
 Replace disable_check_wokflow_job_labels by opt in enable_workflow_job_labels_check, and check all label

npalm:
## Description
When support was added for the workflow_job even also the label check was introduced. By default the label check was enabled. And only the custome labels were supposed to be checked. This was creating confusion. When we need a label check we found out we really need a check on all labels, for example to only accept jobs for our our linux runners. There we changed the label check.

### Opt-in
The label checks are opt-in by setting `enable_workflow_job_labels_check` to `true`

### Accept events
Events are still accepted, also if labels are not acceptec. Workflow job events with labels that not match are dropped. The response message contains the reason and statusCode 202.

### Matching
The match is now strict, all lavels in the workflow job should match all labels of the runners.
- `self-hosted` label is added by design
- `runner_os` and `runner_architecture` are used for label check
- `runner_extra_labels` are used for custom labels.

## Related
Related issues (closed): #1530 #1348

## Migration
- `disable_check_wokflow_job_labels=true` simply remove no other change neede
- Otherwise, check if you would like to use a strict check on labels. If yes, enable `enable_workflow_job_labels_check`


examples/ephemeral/main.tf
```diff
@@ -28,12 +28,13 @@ module "runners" {
   }
 
   # Grab the lambda packages from local directory. Must run /.ci/build.sh first
-  webhook_lambda_zip                = "../../lambda_output/webhook.zip"
-  runner_binaries_syncer_lambda_zip = "../../lambda_output/runner-binaries-syncer.zip"
-  runners_lambda_zip                = "../../lambda_output/runners.zip"
+  # webhook_lambda_zip                = "../../lambda_output/webhook.zip"
```
ScottGuymer, line 31: Should this be commented?

modules/webhook/variables.tf
```diff
@@ -83,12 +83,18 @@ variable "kms_key_arn" {
   default     = null
 }
 
-variable "runner_extra_labels" {
-  description = "Extra labels for the runners (GitHub). Separate each label by a comma"
+variable "runner_labels" {
+  description = "Labels for the runners (GitHub). Separate each label by a comma. Labels are used to check events when `runner_enable_workflow_job_labels_check` is set to `true`."
```
ScottGuymer, line 87:
```suggestion
  description = "Extra (custom) labels for the runners (GitHub). Separate each label by a comma. Labels checks on the webhook can be enforced by setting `enable_workflow_job_labels_check`. GitHub read-only labels should not be provided."
```

examples/ephemeral/main.tf
npalm: nope

ScottGuymer (approved):

Issue #1610, opened 2022-01-11
github_agent.linux.pkr.hcl gets cannot create directory ‘actions-runner’: Permission denied on packer build


toast-gear:
Just trying to build the default hcl image images/linux-amzn2/github_agent.linux.pkr.hcl (with some small edits to the file) and I seem to get permission denied errors?:

```
# packer build log
==> githubactions-runner.amazon-ebs.githubrunner: Provisioning with shell script: /var/folders/kv/vw_r4gv96tg3n92znzrw_x8hbggjb0/T/packer-shell3159066398
    githubactions-runner.amazon-ebs.githubrunner: Creating actions-runner directory for the GH Action installtion
    githubactions-runner.amazon-ebs.githubrunner: Downloading the GH Action runner from https://github.com/actions/runner/releases/download/v2.286.0/actions-runner-linux-x64-2.286.0.tar.gz to actions-runner.tar.gz
==> githubactions-runner.amazon-ebs.githubrunner: mkdir: cannot create directory ‘actions-runner’: Permission denied
==> githubactions-runner.amazon-ebs.githubrunner:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
==> githubactions-runner.amazon-ebs.githubrunner:                                  Dload  Upload   Total   Spent    Left  Speed
==> githubactions-runner.amazon-ebs.githubrunner: 100   678  100   678    0     0   5603      0 --:--:-- --:--:-- --:--:--  5603
==> githubactions-runner.amazon-ebs.githubrunner: Warning: Failed to create the file actions-runner.tar.gz: Permission denied
==> githubactions-runner.amazon-ebs.githubrunner:   0  132M    0  1362    0     0   4515      0  8:33:27 --:--:--  8:33:27  4515
==> githubactions-runner.amazon-ebs.githubrunner: curl: (23) Failure writing output to destination
```

The mkdir seems to fail `mkdir: cannot create directory ‘actions-runner’: Permission denied`. I'm just running this from my laptop with a role assumed with admin rights.

<details>
  <summary>Packer HCL</summary>

**issued command :** packer build .

```hcl
packer {
  required_plugins {
    amazon = {
      version = ">= 0.0.2"
      source  = "github.com/hashicorp/amazon"
    }
  }
}

variable "runner_version" {
  description = "The version (no v prefix) of the runner software to install https://github.com/actions/runner/releases"
  type        = string
  default     = "2.286.0"
}

variable "region" {
[...]

PR #1611, opened 2022-01-11
feat(packer): add vars and minor clean up

toast-gear:
👋 , I wish to use your packer hcl however it needs more variables to be usable in my env:

1. SGs that are open to 0.0.0.0/0 are a big no no so I added a var to allow the SG to be passed in allowing for a more restrictive SG to be created up front. If no SG var is set then null results in the default behaviour of creating a temp SG (as happens now). Added additional vars needed for placing the builder and setting the instance type.
2. Seperated the version from the url for better readability and to perhaps allow for this to be bumped by renovate automatically later.
3. Fixed the permission denied errors raised in https://github.com/philips-labs/terraform-aws-github-runner/issues/1610. /opt and everything in it are owned by root root and so mkdir won't work without sudo and then a subsequent ownership change for the curl.

EDIT in draft mode as I need to try it with a pre-made SG first

### DIFF
--- a/modules/runners/templates/install-runner.sh
+++ b/modules/runners/templates/install-runner.sh
@@ -13,7 +13,7 @@ file_name="actions-runner.tar.gz"
 
 echo "Creating actions-runner directory for the GH Action installtion"
 cd /opt/
-mkdir actions-runner && cd actions-runner
+sudo mkdir actions-runner && sudo chown -R "$(whoami)" actions-runner && cd actions-runner
 
 if [[ -n "$RUNNER_TARBALL_URL" ]]; then
   echo "Downloading the GH Action runner from $RUNNER_TARBALL_URL to $file_name"
--- a/modules/runners/templates/install-runner.sh
+++ b/modules/runners/templates/install-runner.sh
@@ -13,7 +13,7 @@ file_name="actions-runner.tar.gz"
 
 echo "Creating actions-runner directory for the GH Action installtion"
 cd /opt/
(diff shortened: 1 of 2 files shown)


npalm (commented):
@toast-gear thanks LGTM, fully agree security groups should be configurable and subnets should be configurable.

modules/runners/templates/install-runner.sh
```diff
@@ -13,7 +13,7 @@ file_name="actions-runner.tar.gz"
 
 echo "Creating actions-runner directory for the GH Action installtion"
 cd /opt/
-mkdir actions-runner && cd actions-runner
+sudo mkdir actions-runner && sudo chown -R "$(whoami)" actions-runner && cd actions-runner
```
ScottGuymer, line 16: I'm not sure how this would behave when running as part of the user-data script where it is already running as root by default. It may cause an error.

Might be safer to use the trick I mentioned in #1610 so that these scripts are run under the same conditions in both user-data and packer.

ScottGuymer (commented):
Looks good! Thanks for the contribution, the packer templates are new and are in need of some iteration for sure.

Just one small comment on this change.

Also, would it be possible to make similar changes to the windows template?

modules/runners/templates/install-runner.sh
toast-gear: Done, resolve if you're happy

ScottGuymer (approved):

PR #1621, opened 2022-01-13
fix(runnrs): Pool runners to allow multiple event rules and targets

piscue:
if the pool list has more than 1 element, terraform cannot apply

It will throw errors like
```sh
Error: Error adding new Lambda Permission for github-pool: ResourceConflictException: The statement id (AllowExecutionFromCloudWatch) provided already exists. Please provide a new statement id, or remove the existing statement.
{
  RespMetadata: {
    StatusCode: 409,
    RequestID: "1baaaa63-a167-49a1-9eb4-e07f824a85ce"
  },
  Message_: "The statement id (AllowExecutionFromCloudWatch) provided already exists. Please provide a new statement id, or remove the existing statement.",
  Type: "User"
```

### DIFF
--- a/modules/runners/pool/main.tf
+++ b/modules/runners/pool/main.tf
@@ -99,7 +99,7 @@ data "aws_iam_policy_document" "lambda_assume_role_policy" {
 resource "aws_cloudwatch_event_rule" "pool" {
   count = length(var.config.pool)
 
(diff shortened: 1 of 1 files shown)


npalm (approved):
@piscue thanks! Looks good, and tested.

Issue #1620, opened 2022-01-13
arm-runner-patch is invalid


ndench:
The [arm-runner-patch.tpl](https://github.com/philips-labs/terraform-aws-github-runner/blob/1140f5ce8b4ee45748bc783e81254d08198afb4b/modules/runners/templates/arm-runner-patch.tpl) is trying to apply invalid patches during boot time. By checking `/var/log/user-data.log` I can see the follow errors:

```
patching file bin/Runner.Listener.runtimeconfig.json
Hunk #1 FAILED at 8.
1 out of 1 hunk FAILED -- saving rejects to file bin/Runner.Listener.runtimeconfig.json.rej
patching file bin/Runner.PluginHost.runtimeconfig.json
Hunk #1 FAILED at 8.
1 out of 1 hunk FAILED -- saving rejects to file bin/Runner.PluginHost.runtimeconfig.json.rej
patching file bin/Runner.Worker.runtimeconfig.json
Hunk #1 FAILED at 8.
1 out of 1 hunk FAILED -- saving rejects to file bin/Runner.Worker.runtimeconfig.json.rej
```

I then checked the files that it is trying to patch and the json files do not match:

```
       }
     ],
     "configProperties": {
       "System.Runtime.TieredCompilation.QuickJit": true
     }
   }
 }
```

Instead they contain:

```
      }
    ],
    "configProperties": {
      "System.Reflection.Metadata.MetadataUpdater.IsSupported": false
    }
  }

```

Which must be making the patch fail. My runners are then failing to boot and not accepting any workflow jobs.

ScottGuymer:
I'm starting to think that this should probably not live inside this module and should be something injected in by pre or post install user data hooks. 

Its likely this could change a lot over time and having to change this module each time it does seems inconvenient.

@npalm WDYT?

npalm:
Agree, supporting ARM on os and architecture level by the module is fine. But maybe better to move the tailored script to an example with indeed a pre / post hook. Any help would be welcome!

ndench:
I've had a go at fixing the patch in #1624, it makes the arm runners boot correctly and appear to accept a job. However the job never actually starts and eventually times out. Is this patch even required anymore?

If the patch is still necessary, I agree that it should be solved with an example in the documentation instead of being hardcoded. The user-data scripts are not as flexible as I would like as they are, so I think there's definitely some improvement that we can make there along with this change.

[...]

PR #1624, opened 2022-01-13
feat(runner): Replace patch by install ICU package for ARM runners

ndench:
~The runtimeconfig.json files seem to have changed on the latest Amazon Linux AMI.~
~When running the user data scripts, the patch of these files fails which causes initialization of the runner instance to fail.~

~I've run with the following patch template and the arm instance is now booting and seems to be accepting jobs. However they~ ~are not running and hang indefinitely. I'm unsure if there are any flow on changes required here?~

~My jobs all look like this:~
image

**Update**
Turns out new runner releases just keep changing the runtimeconfig.json files. They changed in each of the last 3 releases which continue to break the patch. Instead of patching the config, we can install the `icu` package which now exists in the repositories.

Closes: #1620


ScottGuymer:
I think we should move this patching into an example where we show how to do this by injecting a pre install hook into the user data.

npalm:
@ndench The short snippet of the log you added shows the created instance got connected and the agent have received the job. In the AWS webconsole you access the instance (once ssm configuration variable of module is enabled). Next you could check log files like `/var/log/user_data/` or the agent logs in the `_diag` folder of the runner.

ndench:
I've noticed the `runtimeconfig.json` files seem to be changing frequently which keeps breaking the patch. Turns out all we need to do is install `libicu60` which wasn't present in the repository when arm64 support was originally added. I've updated this PR to just install that package.

Additionally, as you can see from [this commit](https://github.com/philips-labs/terraform-aws-github-runner/pull/1624/commits/639c46e15b37935264842a360d55a1d1d8ad0051) I tested removing the arm patch entirely as has been suggested. I can then install the package using the existing `userdata_pre_install` variable:

```
  userdata_pre_install = <<EOF
sudo yum -y install libicu60
EOF
```

I'm happy to add that commit back in, and update the README to specify users need to install that package. However I prefer that the module handle installing that. It makes it much simpler from a users point of view. Plus it's not a flaky anymore when the config files change.

Let me know which approach is preferred.

npalm:
@ndench Thanks a lot for improving the support for ARM. Would you have a bit of time to add an example for ARM? It will help other a lot that would use the module with ARM support!

ndench:
No problem @npalm, I updated the README and added an example for arm64 instances.

brainlock:
We're also using this for the occasional ARM runner, and just ran into the same problem. If it helps, I tested this PR, works great. Thank you @ndench!

modules/runners/templates/arm-runner-patch.tpl
```diff
@@ -1,49 +1 @@
...
--      "System.Runtime.TieredCompilation.QuickJit": true
-+      "System.Runtime.TieredCompilation.QuickJit": true,
-+      "System.Globalization.Invariant": true
-     }
-   }
--}
-\ No newline at end of file
-+}
-ICU_PATCH
+sudo yum -y install libicu60
```
ScottGuymer, line 1: If this is all that is now needed maybe we can just add it as a conditional statement in the user-data.sh file rather than having to include this file.

npalm: Agree, but still it make sense to have an example.

ndench:
@ScottGuymer @npalm no problem, updated it to remove `arm-runner-patch.tpl` entirely. Tested on my infrastructure and it works a treat.

ScottGuymer (approved):
LGTM.

If extra things are needed in the future they can always be added by the pre or post userdata hooks.

npalm (changes requested):
@ndench thanks great work!. Tested and all worked. 
image

Only one small request. Can you update the ci workflow file `.github/workflows/terraform.yml` and add the `arm` example dir to the matrix of the job `verify_examples` See https://github.com/hyraiq/terraform-aws-github-runner/blob/519f130ec9fd36a6e327857809256178ca1aa4ba/.github/workflows/terraform.yml#L46

ndench:
No problem @npalm, I also noticed that the ephemeral and windows examples were missing from the github workflow, so I added them while I was there.

npalm:
@ndench thanks for adding the other examples as well

npalm (approved):

PR #1644, opened 2022-01-18
feat(images): Added ubuntu-focual example packer configuration

toast-gear:
I had issues with actions like actions/setup-python using the Amazon Linux 2 image. It might be quite nice to also have an Ubuntu image in the repo given action authors tend to only target Ubuntu due to GitHub only providing a Ubuntu based runner. 

Changes:
- Minimal Ubuntu based image
- Expanded tag variables for Amazon Linux 2 image (moved this https://github.com/philips-labs/terraform-aws-github-runner/pull/1629 PR into here basically)

TODO:
- [x] I actually need to test it
- [x] Add in the Cloudwatch stuff


images/install-runner.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-user_name=ec2-user
+user_name=$(cat /tmp/install-user.txt)
```
ScottGuymer, line 3: We have been moving to take the install out of the user directory and into the `/opt` directory so that this is no longer needed when installing, only when starting the runner where it is passed through via SSM.

toast-gear: we still do a `chown -R "$user_name":"$user_name" .` as part of the install, it's on line 46 of the runners module install-runner.sh script, do we think this isn't needed then?

toast-gear:
image

Struggling to get Actions to pickup some envs I'm setting in the image. Any ideas for how to fix this @ScottGuymer . I've tried setting it in the `/etc/environment` via the below but no use/

```
"echo ImageOS=ubuntu20 | sudo tee -a /etc/environment"
```

toast-gear:
Yup that worked, fun new error though 🤔 

image

toast-gear:
image

Preconfiguring the tool cache directory seems to fix the problem 🎉

toast-gear:
That's probably enough now for a review and merge if it all looks OK

toast-gear:
Friendly bump :D

npalm:
> Friendly bump :D

Thanks, but I was certainly not forgotten your PR but having it a bit too busy to get it tested. Sorry for the delay.

examples/ubuntu/main.tf
```diff
@@ -64,7 +64,7 @@ module "runners" {
     {
       "log_group_name" : "runner",
       "prefix_log_group" : true,
-      "file_path" : "/home/runners/actions-runner/_diag/Runner_**.log",
+      "file_path" : "/opt/actions-runner/_diag/Runner_**.log",
```
ScottGuymer, line 67: note: this is probably something we can set globally in https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/logging.tf#L21 as it should now be the same for all linux configs

images/install-runner.sh
ScottGuymer: Yeah its possible this is no longer needed.

modules/runners/templates/install-runner.sh
```diff
@@ -12,10 +12,13 @@ fi
 
 file_name="actions-runner.tar.gz"
 
-echo "Creating actions-runner directory for the GH Action installtion"
+echo "Setting up GH Action tool cache"
+mkdir -p /opt/hostedtoolcache
```
ScottGuymer, line 16: Im not clear what this directory is for?

toast-gear: https://docs.github.com/en/enterprise-server@3.3/admin/github-actions/managing-access-to-actions-from-githubcom/setting-up-the-tool-cache-on-self-hosted-runners-without-internet-access it's the location of the runner tool cache, it goes by many names https://github.com/actions/runner/blob/ac31fd10b26b89b21e6294c0ef7050fce1c498bf/src/Runner.Common/HostContext.cs#L260

Normally you can use any custom location however ruby appears Ruby can only support the default location due to technical reasons https://github.com/ruby/setup-ruby/issues/242#issuecomment-991103529. 

Without the folder already created some setup actions can fail so it's better to have it setup as part of creating the runner image, perhaps the Amazon Linux 2 image needs it setup too?

npalm: I test for the linux image quite a lot setups, the python one does not working, since it has hard ubuntu dependencies. THe ruby one is not covered. Will add a check as well.

ScottGuymer: Ok can you add a comment pointing to this for future reference.

ScottGuymer: Yeah, the setup-* actions from GH should work fine on ubuntu but some of them have a hard dependency on it and wont work in other distros.

toast-gear: Done, I couldn't really find a good link that explains it properly that was for the github.com. Hopefully the comments I've added explain it well enough that it's clear or can easily be found googling.

examples/ubuntu/main.tf
npalm: @scott would be great, but the module provided example is amazon linux only. So for ubuntu override is needed.

images/install-runner.sh
npalm: No, we cannot remmove the chown command, since it is setting the permissions for the action-runner agent dir. Some lines before the dir is set to `/opt/actions-runner`

```bash
cd /opt/
mkdir -p actions-runner && cd actions-runner
```

npalm (approved):
@toast-gear Great work! I have tested the PR, works like a charm. Just wondering, did you measure the boot time for ubuntu? I could bring it down to roughly 38 seconds (time after the start script is launched). So, it excludes the time for the instance to get ready.

Used the ubuntu examples with the following config:
```hcl
  delay_webhook_event      = 0
  enable_ephemeral_runners = true

  enabled_userdata = false
  ami_filter       = { name = ["github-runner-ubuntu-focal-amd64-2022*"] }
  ami_owners       = [data.aws_caller_identity.current.account_id]
```

PR #1676, opened 2022-01-28
feat: Add output image id used in launch template

toast-gear:
I would like the AMI to be output so I can expose it in a pipeline

### DIFF
--- a/modules/runners/outputs.tf
+++ b/modules/runners/outputs.tf
@@ -3,7 +3,7 @@ output "launch_template" {
 }
 
 output "image_id" {
-  value data.aws_ami.runner.id
+  value = data.aws_ami.runner.id
 }
 
 output "role_runner" {
--- a/modules/runners/outputs.tf
+++ b/modules/runners/outputs.tf
@@ -2,10 +2,6 @@ output "launch_template" {
   value = aws_launch_template.runner
 }
 
-output "image_id" {
-  value = data.aws_ami.runner.id
-}
-
 output "role_runner" {
   value = aws_iam_role.runner
 }
--- a/modules/runners/outputs.tf
+++ b/modules/runners/outputs.tf
@@ -2,6 +2,10 @@ output "launch_template" {
   value = aws_launch_template.runner
 }
 
+output "image_id" {
+  value data.aws_ami.runner.id
+}
+
 output "role_runner" {
   value = aws_iam_role.runner
 }
--- a/outputs.tf
+++ b/outputs.tf
@@ -1,13 +1,15 @@
 output "runners" {
   value = {
-    launch_template_name    = module.runners.launch_template.name
-    launch_template_id      = module.runners.launch_template.id
-    launch_template_version = module.runners.launch_template.latest_version
-    lambda_up               = module.runners.lambda_scale_up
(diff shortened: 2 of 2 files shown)


modules/runners/outputs.tf
```diff
@@ -2,6 +2,10 @@ output "launch_template" {
   value = aws_launch_template.runner
 }
 
+output "image_id" {
```
npalm, line 5: Looks no reason to output, can you remove this aoput. Used image_id is part of the launch templates whic is already outputed

npalm (changes requested):
Thanks, can you also add a small description to the PR?

modules/runners/outputs.tf
toast-gear: We output the launch template but the launch template doesn't expose anything of its settings so we need to output them as well if we want them exposed? https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#attributes-reference Apologies if it does, I'm somewhat new to Terraform still.

npalm: No there is no need to output the image_id in this way, it is already outputted as part of the object launch_template., see https://github.com/philips-labs/terraform-aws-github-runner/blob/db4fc4b7caafe36c633f7fcfcbd2257e424dbb39/modules/runners/outputs.tf#L1

outputs.tf
```diff
@@ -1,13 +1,15 @@
...
-    lambda_up               = module.runners.lambda_scale_up
-    lambda_down             = module.runners.lambda_scale_down
-    role_runner             = module.runners.role_runner
-    role_scale_up           = module.runners.role_scale_up
-    role_scale_down         = module.runners.role_scale_down
+    launch_template_name     = module.runners.launch_template.name
+    launch_template_id       = module.runners.launch_template.id
+    launch_template_version  = module.runners.launch_template.latest_version
+    launch_template_ami_id   = module.runners.launch_template.latest_version
+    launch_template_image_id = module.runners.image_id
```
npalm, line 7:
```suggestion
    launch_template_image_id = module.runners.launch_template.image_id
```

npalm (changes requested):

outputs.tf
toast-gear: For my benefit how can I tell what is already output? https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#attributes-reference I thought it was a case of looking at the attribute references for a given resource? So in this case `launch_template.id` and `launch_template.latest_version` are both documented attribute references, how can I tell `image_id` is one without just trying it?

npalm: Don't have a solid answer. One thing you can do is check state. But agree it is not always very clear what is exported.

toast-gear:
- removed output from runner module
- added image_id output as suggested
- removed `launch_template_ami_id   = module.runners.launch_template.latest_version` as the [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#attributes-reference) say that the `latest_version` is the "the latest version of the launch template", not the AMI.

npalm (approved):
LGTM

Issue #1717, opened 2022-02-09
Making security group creation step optional


krlydm:
Hi,

I'm facing with an issue on using this solution for our use cases at my company. My company is restricting who has access to create VPC or Security groups in AWS, so implicit security group creating in not allowed due to security concerns. Removing VPC creation is easy, but I can't find any good option to do it for SG. As a workaround to test this, I removed the security group creation, compiled an own runners.zip and pass my pre-created security group ID via `runner_additional_security_group_ids`. 

Is it possible to solve this differently or this would be a new feature request instead? 

It would be nice if there is option to treat security groups in a same way as VPC and subnets are handed.

npalm:
Sounds like a valid option to me, today setting up the runner require several permissions. We use permission boundaries to limit the scope of the deployment role in AWS. Just wondering, is creating IAM roles allowed?

Feel free to create a PR to optional disable the security group [creation of the runner module](https://github.com/philips-labs/terraform-aws-github-runner/blob/a49fab4703dc6eec88d83b457af268a0f802eef5/modules/runners/main.tf#L138) (default should be true).

krlydm:
Sounds good to me, I'm open to create a PR, seems like a simple change what you recommended. 
Handling IAM roles are bit different in my case, because we have more room to play with limiting permission boundaries to a specific role via ARN. Security group is like a firewall, they are handled org-wide to make sure we are applying approved restrictions everywhere.

krlydm:
I have created the pull request for addressing this request: #1718

PR #1718, opened 2022-02-10
feat(runner): Ability to disable default runner security group creation

krlydm:
Closes: #1717 

It can restricted who has access to create/modify VPCs or Security groups in AWS, so implicit security group creating in runner modul can be a problem in these cases. Removing VPC creation is easy, but there is no option to do it for the default security group. This change adds option to disable the security group creation of the runner module.


modules/runners/main.tf
```diff
@@ -136,6 +136,7 @@ resource "aws_launch_template" "runner" {
 }
 
 resource "aws_security_group" "runner_sg" {
+  count = var.enable_managed_runner_security_group ? 1 : 0
```
npalm, line 139: Please can you check line 91, here the security group is used and needs to be updated for the count. I think this should work

```hcl
  vpc_security_group_ids = compact(concat(
    var.enable_managed_runner_security_group ? [aws_security_group.runner_sg[0].id] : [],
    var.runner_additional_security_group_ids,
  ))
```

npalm (changes requested):
@krlydm Tested and is not working due to dependency of the security group.

modules/runners/main.tf
krlydm: Updated.

npalm: Thx will check asap, CI is also failing with formatting errors. Can you check that as well?

krlydm: Updated again. Sorry about these "newbie" issues, Terraform is entirely new for me, but now I know what should I run locally to avoid these issues.

npalm: Still two jobs are failing. I will have a look later, this seems to be caused by the new Terraform AWS provider 4.0 See #1739

npalm:
@krlydm can you rebase with the develop branch, I hope that will fix the build failures

krlydm:
> @krlydm can you rebase with the develop branch, I hope that will fix the build failures

@npalm Rebased.

krlydm:
@npalm Is there anything that I need to do to resolve the change request?

npalm:
@krlydm no I only need a bit of time. WIll do my best to merge asap. Thx!

npalm (approved):
@krlydm LGTM, thanks!

Issue #1745, opened 2022-02-14
Update packer scripts to remove disks on termination


ScottGuymer:
Responding to myself after noticing by looking at the AMI that it had "Delete on termination" set to "No"
<img width="1051" alt="Screen Shot 2022-02-11 at 11 55 46 AM" src="https://user-images.githubusercontent.com/5667028/153634727-5f848f5c-4565-412b-a369-b4ff7edfcd3a.png">

In `/mages/ubuntu-focal/github_agent.ubuntu.pkr.hcl`, I updated to:
```
  launch_block_device_mappings {
    device_name           = "/dev/sda1"
    volume_size           = var.root_volume_size_gb
    volume_type           = "gp3"
    delete_on_termination = true
  }
```

_Originally posted by @Fgerthoffert in https://github.com/philips-labs/terraform-aws-github-runner/discussions/1742#discussioncomment-2158915_

PR #1758, opened 2022-02-15
feat: Variablize delete_on_termination

Hunqer:
Issue: #1745
    Variablize delete_on_termination for EBS volume.
    Make default as true (which make more sense to delete volumes
    as they are no longer used by runners).

Fixes: https://github.com/philips-labs/terraform-aws-github-runner/issues/1745

https://www.packer.io/plugins/builders/amazon/ebs#delete_on_termination
By default packer use delete_on_termination set to false.

> [delete_on_termination](https://www.packer.io/plugins/builders/amazon/ebs#delete_on_termination) (bool) - Indicates whether the EBS volume is deleted on instance termination. Default false. NOTE: If this value is not explicitly set to true and volumes are not cleaned up by an alternative method, additional volumes will accumulate after every build.

### DIFF
--- a/images/linux-amzn2/github_agent.linux.pkr.hcl
+++ b/images/linux-amzn2/github_agent.linux.pkr.hcl
@@ -42,6 +42,12 @@ variable "root_volume_size_gb" {
   default = 8
 }
 
+variable "ebs_delete_on_termination" {
+  description = "Indicates whether the EBS volume is deleted on instance termination."
+  type        = bool
+  default     = true
+}
+
 variable "global_tags" {
   description = "Tags to apply to everything"
   type        = map(string)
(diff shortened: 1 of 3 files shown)


marcelloromani (approved):

ScottGuymer:
Looks good. Can you also add this to the windows tempalte?

Hunqer:
> Looks good. Can you also add this to the windows tempalte?

Added also to windows.

Hunqer:
@ScottGuymer  had to make update yesterday due to missing `device_name` which is required.

ScottGuymer:
> @ScottGuymer had to make update yesterday due to missing `device_name` which is required.

We need to be careful that it is the same device name as the initial root device of this will just add a new one and not affect the root device that come with the AMI.

Hunqer:
> We need to be careful that it is the same device name as the initial root device of this will just add a new one and not affect the root device that come with the AMI.

@ScottGuymer should I change `device_name` then? But I see that all checks have passed without problems.

Hunqer:
I've used info from this site: https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/device_naming.html
The `reserved for root` one is `/dev/sda1`.

ScottGuymer (approved):

ScottGuymer:
Thanks for the contribution!

PR #1783, opened 2022-02-23
fix: Use source_hash instead of etag to prevent drift when using KMS

marekaf:
[source_hash](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_object#source_hash) - (Optional) Triggers updates like etag but useful to address etag encryption limitations. Set using filemd5("path/to/source") (Terraform 0.11.12 or later). (The value is only stored in state and not saved by AWS.)

source_hash should help eliminate some "plan drifts" that are caused when using KMS encryption and etag together

it shouldn't affect the functionality if not using SSE KMS on the bucket.


npalm (approved):
@marekaf Thanks!

Issue #1811, opened 2022-03-03
Build error: Your kernel version indicates a revision number of 255 or greater


npalm:
## Description

Since March the 2nd we have seen in ubuntu:20.04 / latest container builds the following issue:

```
ERROR: Your kernel version indicates a revision number
of 255 or greater.  Glibc has a number of built in
assumptions that this revision number is less than 255.
```

Ther result is that build are fialing, in our cases combinations of ubuntu and python.

## Reproduce
Run a job in with ubuntu:20.04 as container and execute for example

```
apt-get update -y
apt-get install python3-pip -y
```

Result failing build with error above. 

## Solution

Investigation learned that the issue is related to the amazon linux 2 base AMI, what exact is triggering the failures is not cleear to us. Upgrade to a amazon linux 2 kernel 5x base AMI solve the issue. Other options are running on ubuntu for example:

In case you have the issue, quickest solution is setting the AMI filter for the trunner

```hcl
  ami_filter     = { name = ["amzn2-ami-kernel-5.*-hvm-*-x86_64-gp2"] } 
  ami_owners = ["137112412989"] 

```

PR #1812, opened 2022-03-03
fix(runners): Upgrade Amazon base AMI to Amazon Linux 2 kernel 5x

npalm:
Upgraded the Amazon Linux base AMI to avoid errors mentioned in #1811


npalm:
close #1811

Issue #1245, opened 2021-10-07
Module is not compatible with the GovCloud region


aidan-mundy:
I will add more problems here as I find them, initial problem in my GovCloud journey is that the GovCloud region ARNs start with "arn:aws-us-gov:" and the module hardcodes ARNs as "arn:aws:". (A similar problem will cause this module to fail in the China regions)
https://github.com/philips-labs/terraform-aws-github-runner/blob/1e6c8b02f498be71238975eebe0aee18a3668084/modules/runners/policies-runner.tf#L29

npalm:
I assume we have hard coded this even on more places. Feel free to sugges an absraction in a PR. Thanks for taking the time to create an issue.

aidan-mundy:
@npalm Surprisingly, changing that single line was enough to make the module work in my GovCloud environment with GHES. I am using the root module, default KMS, and most of the standard options. 

I will be doing more testing with some of the more niche options to find other GovCloud incompatibilities. Once I'm satisfied with my hard-coded changes, I'll write up an abstraction for it (hopefully automated, but potentially with an input variable) and put up a PR.

npalm:
@aidan-mundy is you issue resolved?

aidan-mundy:
I havent found any additional issues in the module (related to GovCloud anyway). Haven't had the chance to throw up an MR for this, but I can do that in the next week or two.

aidan-mundy:
@npalm I havent found any additional issues in the module (related to GovCloud anyway). Haven't had the chance to throw up an MR for this, but I can do that in the next week or two.

jokreliable:
I believe that with the merge mentioned above, this issue can be resolved.

aidan-mundy:
Agreed, this appears to resolve the issue I had detected, as well as other potentially related issues. It appears this fell off my radar at some point, thanks for picking up the slack @jokreliable

jokreliable:
No worries - Took me a month to get back on my radar too..  😄 -- Glad it helped!

PR #1815, opened 2022-03-03
feat: Add Support for Alternative Partitions in ARNs (like govcloud)

jokreliable:
In order to use this module in AWS, a number of ARN need to be updated to point to the proper partitions used there.

This PR provides a new setting "aws_partition" that is used as a parameters in (I hope) all the various ARNs that are hardcoded in the module/submodules. The default is set to be "aws" everywhere it was defined, so it shouldn't change the behavior for "standard" aws.

I ran pre-commit and all tests there pass. I am unsure if there is a good way to build better/more specific tests you could run.

Looking forward to your feedback. 
(And thank you for making this module public! It is going to be super helpful)

potentially addresses #1245


npalm:
@jokreliable thanks, just wondering. Is this to support for example the gov cloud or china cloud of AWS?

jokreliable:
Indeed. For govcloud in the US for example, the ARNs start with `arn:aws-us-gov` instead of `arn:aws`. I'm unsure what the partition name is for China, but I expect something along the same lines (that is what the documentation suggests anyway).

npalm (approved):
@jokreliable LGTM, tested with no partion set.

PR #1816, opened 2022-03-04
feat(images): Add associate_public_ip_address variable

ulich:
If using a non-default VPC, there is no public IP address assigned to the EC2 instance. If you specified a public subnet, the EC2 instance won't have access to the internet without setting `associate_public_ip_address = true`.


npalm (approved):
LGTM @ScottGuymer can you do a quick check as well?

ScottGuymer:
Looks good. 

Please can you add the same config to the windows packer config too?

PR #1817, opened 2022-03-04
Improve example/ubuntu runner run as user data template

marcofranssen:
This prevents issues by not aligning the `runner_run_as` variable in terraform with the `USERNAME` used in the `user-data.sh` template.

While upgrading from an older release things have changed and misalignment in the `user-data.sh` template and Terraform `main.tf` occurred in our project. Took a while to figure out and debug finding those changes and updates in the example.

This fix should prevent it for future consumers. Now terraform will fail if the user did not define the `runner_run_as` variable in the `user-data.sh` template, which is more transparent and applies the fail fast fail cheap principle. So as opposed to waiting an debugging the runner when it comes online it will now fail when applying terraform.

cc @gjkamstra

### DIFF
--- a/examples/ubuntu/templates/user-data.sh
+++ b/examples/ubuntu/templates/user-data.sh
@@ -15,9 +15,12 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y \
     build-essential \
     unzip
 
-USER_NAME=runners
+USER_NAME=${runner_run_as}
 useradd -m -s /bin/bash $USER_NAME
 USER_ID=$(id -ru $USER_NAME)
+
+# uncomment following line to be able to debug and login as the user via ssm
+# echo -e "test1234\ntest1234" | passwd $USER_NAME
 
 # install and configure cloudwatch logging agent
 wget https://s3.amazonaws.com/amazoncloudwatch-agent/ubuntu/amd64/latest/amazon-cloudwatch-agent.deb
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -120,6 +120,7 @@ resource "aws_launch_template" "runner" {
       S3_LOCATION_RUNNER_DISTRIBUTION = var.s3_location_runner_binaries
       RUNNER_ARCHITECTURE             = var.runner_architecture
     })
+    runner_run_as   = var.runner_run_as
     post_install    = var.userdata_post_install
     start_runner    = templatefile(local.userdata_start_runner[var.runner_os], {})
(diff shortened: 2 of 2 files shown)


ScottGuymer:
The number of arguments that are replaced in the userdata template has been reduced on purpose to better support configuration of pre-built AMIs and allow the passing of config to them at runtime.

The intent is to pass information to runners with the use of SSM parameters. 

The run_as value is assigned as a [parameter here](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/runner-config.tf#L1). 

You can see it pulled out in the default [`start-runner.sh`](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/start-runner.sh#L23)

This mechanism allows you to easily pass further parameters by creating new SSM params outside of the module that can be consumed in a custom user-data (or other) script.

marcofranssen:
@ScottGuymer where can I find those changes documented? What do you suggest to improve in the current example? Wanted to ensure the example is more aligned with the latest v0.36.0 release.

ScottGuymer:
Yes please update this PR to use the SSM parameter approach.

marcofranssen:
It seems this whole part is required for that run_as variable.

https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/start-runner.sh#L5-L21

Now I'm not deep enough in the code. Can I reuse those ip addresses etc. in the example? In the end I just want to align the `run_as` variable from terraform with the user being created on the machine.

This seems as a lot of added complexity to the `user_data.sh` if I have to include all of that just to create this user.

Could you elaborate/assist a bit further.

ScottGuymer:
> It seems this whole part is required for that run_as variable.
> 
> https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/start-runner.sh#L5-L21

Yes you would need all this to get the properties from SSM.

> Now I'm not deep enough in the code. Can I reuse those ip addresses etc. in the example? In the end I just want to align the `run_as` variable from terraform with the user being created on the machine.

These are standard IPs provided by AWS for querying metadata from an instance.

> This seems as a lot of added complexity to the `user_data.sh` if I have to include all of that just to create this user.

Bear in mind this is an example and should be taken by users and modified. If the usecase requires modifying the username as in the example then it would be fine to have that hardcoded in your custom scripts as required.

The intent with the examples is to have some working examples of how to do some common configurations and not to provide fully customisable solutions.

PR #1818, opened 2022-03-04
Improve example/ubuntu user-data.sh (shellcheck fixes + use apt to install docker)

marcofranssen:
This builds further on #1817 and therefore also contains that change.

This PR resolves some shellcheck issues in `user-data.sh`. As well it takes the same approach to install Docker like now being done in the packer ami build script for `ubuntu-focal`.

The `user-data.sh` still install docker as rootless as was done in the original example.

cc: @gjkamstra

### DIFF
--- a/examples/ubuntu/templates/user-data.sh
+++ b/examples/ubuntu/templates/user-data.sh
@@ -6,6 +6,10 @@ ${pre_install}
 # Install AWS CLI
 apt-get update
 DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+    gnupg \
+    lsb-release \
     awscli \
     jq \
     curl \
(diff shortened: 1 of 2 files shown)


examples/ubuntu/templates/user-data.sh
```diff
@@ -15,59 +19,54 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y \
...
     unzip
 
-USER_NAME=runners
-useradd -m -s /bin/bash $USER_NAME
-USER_ID=$(id -ru $USER_NAME)
+USER_NAME=${runner_run_as}
+useradd -m -s /bin/bash "$USER_NAME"
+USER_ID=$(id -ru "$USER_NAME")
+
+# uncomment following line to be able to debug and login as the user via ssm
```
npalm, line 26: No need to set a password for SSM

examples/ubuntu/templates/user-data.sh
```diff
@@ -15,59 +19,54 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y \
...
-echo export DOCKER_HOST=unix:///run/user/$USER_ID/docker.sock >>/home/$USER_NAME/.profile
-echo export PATH=/home/$USER_NAME/bin:$PATH >>/home/$USER_NAME/.profile
+systemctl stop docker
+systemctl disable docker
 
 # Run docker service by default
-loginctl enable-linger $USER_NAME
-su -l $USER_NAME -c "systemctl --user enable docker"
+loginctl enable-linger "$USER_NAME"
+su -l "$USER_NAME" -c "systemctl --user enable docker"
```
npalm, line 52: Docker service was not started, by using the docker rootless setup it got started by default.
```suggestion
su -l "$USER_NAME" -c "dockerd-rootless-setuptool.sh install"
```

npalm (changes requested):
@marcofranssen thx for your contribution. Sorry for keep you waiting. Checked your updates and found one problem. The docker deamon was not started. Added a suggestion whiched solved the problem.

npalm:
@marcofranssen did you already had time to check the comments?

PR #1832, opened 2022-03-09
Dont delete busy runners

### LINKED ISSUE #166: Improve runner deletion by using `busy` flag
When building this solution the Github API couldn't tell if a runner was busy or not, so we resorted to trying to delete each runner via de API. If that returned a 500 Internal Server Error, we would know if it was busy.

Just played with the API again and saw the `busy` flag was added for runners. See https://docs.github.com/en/rest/reference/actions#list-self-hosted-runners-for-a-repository

Instead of trying to delete a runner we should use this flag so reduce the number of API calls on Github.

ulich:
Fixes #166

I tested this with our setup and it was working fine. But we only use ephemeral runners.

### DIFF
--- a/modules/runners/lambdas/runners/src/gh-auth/gh-auth.ts
+++ b/modules/runners/lambdas/runners/src/gh-auth/gh-auth.ts
@@ -19,6 +19,7 @@ const logger = rootLogger.getChildLogger({ name: 'gh-auth' });
 export async function createOctoClient(token: string, ghesApiUrl = ''): Promise<Octokit> {
   const ocktokitOptions: OctokitOptions = {
     auth: token,
+    log: console,
   };
   if (ghesApiUrl) {
     ocktokitOptions.baseUrl = ghesApiUrl;
--- a/modules/runners/lambdas/runners/src/gh-auth/gh-auth.ts
+++ b/modules/runners/lambdas/runners/src/gh-auth/gh-auth.ts
@@ -19,7 +19,6 @@ const logger = rootLogger.getChildLogger({ name: 'gh-auth' });
 export async function createOctoClient(token: string, ghesApiUrl = ''): Promise<Octokit> {
   const ocktokitOptions: OctokitOptions = {
     auth: token,
-    log: console,
   };
   if (ghesApiUrl) {
     ocktokitOptions.baseUrl = ghesApiUrl;
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
@@ -143,6 +143,14 @@ const DEFAULT_RUNNERS_ORIGINAL = [
     launchTime: moment(new Date()).toDate(),
     repo: `${TEST_DATA.repositoryOwner}/${TEST_DATA.repositoryName}`,
   },
+  {
+    instanceId: 'i-busy-112',
+    launchTime: moment(new Date())
+      .subtract(minimumRunningTimeInMinutes + 27, 'minutes')
+      .toDate(),
+    type: 'Org',
+    owner: TEST_DATA.repositoryOwner,
+  },
 ];
 
 const DEFAULT_REGISTERED_RUNNERS = [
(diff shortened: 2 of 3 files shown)


ulich:
@samuelb i adjusted the logic from you a little, i believe some things weren't correct [in your implementation](https://github.com/philips-labs/terraform-aws-github-runner/pull/1832/commits/7d36fa8188f5847be0e41b8d781a99652c4cceb9) (decrementing the idleCounter in every case). Let me know if I have missed something.

ulich:
You are right that a runner could not be busy at the time of listing but then be busy again when sending the delete request. In this case it would fail.

The other scenario, where the runner is busy at the time of listing but then is no longer busy could also happen, but maybe wouldn't be that bad, then it would be deleted the next time the scale-down lambda runs.

Even when performing another "status-check" request right before sending the delete request, it still could lead to both of the aforementioned situations, just that it's less probable because there is less time in between, especially when there were multiple delete requests in that run. The downside with the additional status check would be quite some API requests more which maybe lead to throttling?

Maybe, as a starting point, we use the approach of this PR which already improves, I'd say 90% of the requests. For the cases where the runner becomes idle after the list request, the downside would be that the runner would run for a few minutes longer until the next scale-down lambda runs. For the cases where the runner becomes busy after the list request, the behavior remains as before, a delete request which leads to an error. Maybe I should revert [this line](https://github.com/philips-labs/terraform-aws-github-runner/pull/1832/files#diff-e89df71fb6beb7e4c15d585e6294f24e2fa340f53ed762d40cd8dc024e907bc1R114) back to how it was before?

npalm:
> You are right that a runner could not be busy at the time of listing but then be busy again when sending the delete request. In this case it would fail.
> 
> The other scenario, where the runner is busy at the time of listing but then is no longer busy could also happen, but maybe wouldn't be that bad, then it would be deleted the next time the scale-down lambda runs.
> 
> Even when performing another "status-check" request right before sending the delete request, it still could lead to both of the aforementioned situations, just that it's less probable because there is less time in between, especially when there were multiple delete requests in that run. The downside with the additional status check would be quite some API requests more which maybe lead to throttling?
> 
> Maybe, as a starting point, we use the approach of this PR which already improves, I'd say 90% of the requests. For the cases where the runner becomes idle after the list request, the downside would be that the runner would run for a few minutes longer until the next scale-down lambda runs. For the cases where the runner becomes busy after the list request, the behavior remains as before, a delete request which leads to an error. Maybe I should revert [this line](https://github.com/philips-labs/terraform-aws-github-runner/pull/1832/files#diff-e89df71fb6beb7e4c15d585e6294f24e2fa340f53ed762d40cd8dc024e907bc1R114) back to how it was before?

What would you think about not using busy but using the the status?

ulich:
I don't know what is better. I'm also not familiar with these 2 fields from the GitHub API. But yeah, maybe status is better

ulich:
regarding the additional status check request: What we also could do is check for the status flag from the list runners response and then, if status != active, before actually sending the delete request, do another get request for the particular runner, to double check that the runner didn't become active in the meantime.

This wouldn't cause extremely many more API requests but also reduce the amount of requests that lead to errors because the runner is active again.

Wdyt?

samuelb:
What about not using caching when getting the list of runners (`listGitHubRunners()`)? Then there shouldn't be any significant delay between getting the current state of the runner and sending the deletion requests.

ulich:
As far as I can see, the cache is invalidated when the scale down lambda is invoked. So I don't think this is a huge problem either, or am I missing something?

krlydm:
Apologize for the duplicated work, but I believe the scale down improvement I made in #1859 is also addressing the problem reported in #166 and implemented by this PR. Feel free to raise any question about it.

ulich:
I now tested the code checking for `status` but actually the value for me always was `online`. So I think the `busy` flag is better

npalm:
> I now tested the code checking for `status` but actually the value for me always was `online`. So I think the `busy` flag is the way to go

I will ask the GitHub team if the can point to docs or the meaning of busy. I don't like educated guesses

npalm (approved):
@ulich Thanks, LGTM

PR #1833, opened 2022-03-10
feat: Add option for KMS encryption for cloudwatch log groups

julada:
* added variable to allow cloudwatch log encryption via kms


npalm (approved):
@julada Thanks LGTM!

PR #1838, opened 2022-03-10
feat: Add variable to specify custom commands while building the AMI

ulich:
This can be used to install additional software or to customize anything else for the final AMI


ScottGuymer:
Looks like a nice way of hooking in some extra commands.

Can you do the same with the windows template too?

ulich:
done :)

ScottGuymer (approved):

Issue #1522, opened 2021-12-14
Refactor parsing variables to start runner agent


npalm:
The way we parse runner parameters is becoming hard to maintain and extend. Some refactoring required.

## Review on PR #1374
I think it might be better when working with these command to use a string array and then [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) then with a space. Rather than having to worry about spaces in all the strings etc.

`runnerServiceConfig` could be defined as a string array above and you can merge the arrays and join.

_Originally posted by @ScottGuymer in https://github.com/philips-labs/terraform-aws-github-runner/pull/1374#discussion_r768531638_

ulich:
I would be happy to pick this up

npalm:
Great, you help is very welcome!

PR #1842, opened 2022-03-10
chore: Refactor runnerServiceConfig to be an array of strings

ulich:
Fixes https://github.com/philips-labs/terraform-aws-github-runner/issues/1522


ulich:
I see Sonar complaining about some duplication of code in the scale-up.test.ts. I actually disagree here with Sonar, trying to reduce the duplication would result in less readable/understandable testing code.

What is your approach usually in this case?

npalm:
@ulich please ignore Sonar for now, it was an experiment to add it. Still have to see how we can value out of sonar, but for now ignore it. Will remove it next week

ulich:
Ok. Then the PR is ready for review

npalm (approved):
@ulich thanks LGTM

Issue #1841, opened 2022-03-10
Hitting rate limits because ssm calls in runners.ts are not staggered or spread out


devshah1:
**Issue**
We ran into rate limits yesterday when the pool configuration adjusts the size of EC2 instances by scaling up.

After digging into the code, I saw that the calls to the [ssm `putParameter`](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/lambdas/runners/src/aws/runners.ts#L205) method are not staggered or spread out in `runners.ts`.  

Concurrent calls to this method triggers the rate limits since the parameter store only allows 40 requests per second by default.

**How to reproduce?**
If the pool adjusts more than 40 instances at a time, you'll see throttling errors from the AWS SDK.

**Proposed resolution**
Use waits to stagger calls to the ssm parameter store in the specified file.

**Root cause file reference**:
https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/lambdas/runners/src/aws/runners.ts#L205

devsh4:
@npalm I added a fix for this, please review the PR.

npalm:
According this AWS docs the rate limit should be 100 per second for non mutating and 200 for mutation API calls. Several other API calles are done before parameters are writeen, and as well other lamba's can consume from the same bucket for API rate limits. 

Also during the boot process the instances is retrieving via API calls config via SSM. Those could potential have the same issue. By introducing a delay we should keep in mind that parameters should be created before the instances starts retrieving them. Typically at leas 25 seconds. Furthemore the Lambda have default time out. So that should match as well.

in PR #1843 a wait of 25 ms is introduced after each call. Which limits the put parameter to call no more than 40 times the API in a second. Sound to me reasonable. 

@devshah1 can you test the PR?

devsh4:
@npalm As far as the rate limit for AWS goes I confirmed with the AWS team that the current rate limit was 100 req/s and this is after increasing the throughput for our account. 

Good point about the other calls that happen underneath however I monitored the behavior extensively for our workloads and whenever the pool topped up > 50 EC2 instances (this number changed to 100 once we increased the throughput), the SSM calls met with throttling which means the parameters weren't added to the store and thus EC2 instances couldn't fetch the parameters from SSM on boot. This led to a scenario where we had 2-3x EC2 instances but only half of them got registered successfully as GH runners.

I'll test out the changes in PR #1843 with production level workloads and get back with the results.

[...]

PR #1843, opened 2022-03-12
Add delay to prevent ssm rate limits using setTimeout

devsh4:
Adding a fix for bug #1841 which prevents hitting AWS SSM rate limits by adding a delay between subsequent `putParameter` calls.

These lines had no code coverage, however all pre commit checks passed and I was able to successfully build it.

**Note**: 25ms delay is based on the [AWS service quotas](https://docs.aws.amazon.com/general/latest/gr/ssm.html#limits_ssm) as they only allow 40 requests per second to the parameter store by default.


npalm:
Did a quick check, with 40 runners I get throttle exceptions. Please can you have a look at the CI builds as well.

devsh4:
Yes CI should be fixed now.

npalm:
In you next PR's an approval would not be required anymore :)

devsh4:
Regarding the test failure, it occurred when I ran locally even without my changes.

npalm:
Did you use node14?

devsh4:
> Did you use node14?

Yes it's possible I used the latest version.

devsh4:
@npalm I fixed the failing test. Please approve the build when you can.

image

ScottGuymer (commented):
I'm not sure I'm happy with the approach of adding a fixed delay in here. It effectively increases the execution time of every execution of the lambda by 25ms regardless of if it would be throttled or not. Which translates to a direct cost for each lambda execution.

Are there some better approaches we can come up with such as building in retries for failure and/or some kind of circuit breaker (or anything else)?

devsh4:
> I'm not sure I'm happy with the approach of adding a fixed delay in here. It effectively increases the execution time of every execution of the lambda by 25ms regardless of if it would be throttled or not. Which translates to a direct cost for each lambda execution.
> 
> Are there some better approaches we can come up with such as building in retries for failure and/or some kind of circuit breaker (or anything else)?

Good point, I'll add a check on the instance count and add a fixed delay only if it's more than 40 instances (according to the rate limit).

devsh4:
Ah rebase went wrong, closing this PR in lieu of new one.

Issue #1769, opened 2022-02-21
Pre-build  AMIs cause runners to have additional unused ebs when nothing is provided to block_device


Hunqer:
So we are using [ubuntu-focal](https://github.com/philips-labs/terraform-aws-github-runner/tree/develop/images/ubuntu-focal)
for our runners as we have ebs created from this part of code:
```  launch_block_device_mappings {
    device_name           = "/dev/sda1"
    volume_size           = "${var.root_volume_size_gb}"
    volume_type           = "gp3"
    delete_on_termination = "${var.ebs_delete_on_termination}"
  }
 ```
  
  
But this part cause images to have other ebs with default value of 30GB.
``` dynamic "block_device_mappings" {
    for_each = [var.block_device_mappings]
    content {
      device_name = lookup(block_device_mappings.value, "device_name", "/dev/xvda")

      ebs {
        delete_on_termination = lookup(block_device_mappings.value, "delete_on_termination", true)
        volume_type           = lookup(block_device_mappings.value, "volume_type", "gp3")
        volume_size           = lookup(block_device_mappings.value, "volume_size", var.volume_size)
        encrypted             = lookup(block_device_mappings.value, "encrypted", true)
        iops                  = lookup(block_device_mappings.value, "iops", null)
      }
    }
}  
```
  
  ![Screenshot 2022-02-21 at 16 00 13](https://user-images.githubusercontent.com/19205612/154980161-0a463770-a1e7-42a6-aee4-ddc6136121fc.png)
  
Defaults are taken from:
```variable "block_device_mappings" {
  description = "The EC2 instance block device configuration. Takes the following keys: `device_name`, `delete_on_termination`, `volume_type`, `volume_size`, `encrypted`, `iops`"
  type        = map(string)
  default     = {}
}

variable "volume_size" {
  description = "Size of runner volume"
  type        = number
  default     = 30
}
```

 What should we do about that? Should module accept null and not configure or ignore a empty {}?

mahela-aws:
@Hunqer how did you managed to get it working with [ubuntu-focal](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/images/ubuntu-focal/github_agent.ubuntu.pkr.hcl) ? I have used the same packer template given here with few changes to suit our environment but runners are not getting registered with github. appreciate any help here

toast-gear:
[...]

PR #1807, opened 2022-03-01
feat(runners): Add possibility to not create ebs

Hunqer:
For issue: https://github.com/philips-labs/terraform-aws-github-runner/issues/1769

PR #1845, opened 2022-03-14
Add possibility to create multiple ebs

next0262:
I think this terraform module doesn't work for creating multiple ebs.  
I made some changes to enable allocating multiple ebs to runner. 
I guess my changes can resolve the issue #1769 as well.


npalm:
@next0262 thanks, did you test the change?

toast-gear:
https://github.com/philips-labs/terraform-aws-github-runner/pull/1807 please don't break the ability to provide `null` as part of this PR

npalm:
@toast-gear thanks for the note. Maybe I overlook something. But providing the default `[]` empty array results in the same behavior as `null` in PR #1807

next0262:
@toast-gear Thanks for your correction. I fixed it for keeping the ability for null.
@npalm yes, I did some tests with github and aws and i checked there is no problem.

toast-gear:
I haven't tried your build I just wanted to highlight not to cause a regression with this change, if an empty list doesn't then crack on with removing null!

next0262:
@toast-gear I understand. Thank you for your advice. Please tell me if my additional change has another problem.

npalm:
cc @Hunqer woould you have time for a quick chck of this PR, to be sure it not conflicts your work

next0262:
@Hunqer I am so looking forward to your check. I would really appreciate it .

Hunqer:
@npalm  @next0262 Will check that today, and come back to you.

Hunqer (approved):
From my test looks like it not broking our workflow. @next0262 @npalm

next0262:
@npalm I ask you to approve my changes if there is no problem

npalm:
Will review and check the PR asap

next0262:
@npalm Your check is done?

modules/runners/variables.tf
```diff
@@ -47,8 +47,8 @@ variable "s3_location_runner_binaries" {
 
 variable "block_device_mappings" {
   description = "The EC2 instance block device configuration. Takes the following keys: `device_name`, `delete_on_termination`, `volume_type`, `volume_size`, `encrypted`, `iops`"
```
ScottGuymer, line 49: Can you add a type constraint for the mentioned keys instead of `map(string)`.

variables.tf
```diff
@@ -225,8 +225,8 @@ variable "runner_allow_prerelease_binaries" {
 
 variable "block_device_mappings" {
   description = "The EC2 instance block device configuration. Takes the following keys: `device_name`, `delete_on_termination`, `volume_type`, `volume_size`, `encrypted`, `iops`"
-  type        = map(string)
-  default     = {}
+  type        = list(map(string))
```
ScottGuymer, line 228: Same comment as above.

modules/runners/variables.tf
next0262: @ScottGuymer  I added type constraints on block_device_mappings

variables.tf
next0262: Same reply as above.

next0262:
@ScottGuymer @npalm I modified block_device_mappings of ubuntu example based on type constraints

npalm (approved):

PR #1858, opened 2022-03-16
feat: Replace environment variable by prefix

surminus:
We're looking at using this module to deploy multiple sets of runner types (x86_64 and arm architectures) but within the same conceptual "environment".

We use the "Environment" tag throughout our tooling, but the constraints of using the "environment" variable for resource naming mean that we need to essentially supply different environment names (eg "env-amd64" and "env-arm64"), even though they are not in different environments.

We also use the dot character (".") in our environment names, which isn't allowed in some resource names (eg SQS queue name).

This PR replaces the "environmet" variable by "prefix" to prefix resources crated by the module. The prefix is also used to set the tag: "ghr:environment" with the value of prefix for lambda's to orchestrate the instnaces.

You can still set the tag "environment" to all resources via the AWS provider.


npalm:
@surminus can you push any tiny commit, the ci (actions) did not run. Most likely since github actions was down the last days twice.

surminus:
Done, I amended the commit :)

surminus:
OK, I think that makes sense... I'll take a blast at making these changes.

npalm:
@surminus did you already had time to have a look?

surminus:
Sorry, I started to look and then got a bit stuck in the weeds, since removing this tag entirely is a little more involved. I won't be able to look for at least a couple of weeks now, but I do intend to get back to this eventually!

npalm:
No problem, thanks for the quick update!

surminus:
@npalm I've pushed an update that completely removes the `environment` variable and replaces it where possible with `var.prefix`.

My main conundrum is how to pass in the `ghr:environment` tag to places that need it. At the moment I've added a `lookup` on the tags for this, but am aware that we're not enforcing it's presence, so may be prone to failure. Do we have any thoughts on the best way to pass this along between modules?

I've also added a default `prefix` as `github-actions` - let me know if you think this should be a required setting.

This update will also introduce breaking changes to users environments. They'll need to add an `Environment` tag to `var.tags` and specifically set the prefix to match their previous names.

ScottGuymer:
You are right `lookup(local.tags, "ghr:environment")` definitely feels fragile.

The original implementation used the environment value (now prefix) as the value for the tag. I can see that this change decouples the prefix from the tag and uses a random value. Was there a reason to make this decoupling? 

I think to keep the decoupling and not rely on the lookup would need new variables for the internal modules that use the tag value.

surminus:
> The original implementation used the environment value (now prefix) as the value for the tag. I can see that this change decouples the prefix from the tag and uses a random value. Was there a reason to make this decoupling?

Yes, I felt that it would be nicer to generate a completely new random value rather than depend on the prefix, which didn't feel like it's purpose. However, given that we'd have to create a new variable in all the downstream modules, I will just set it to `var.prefix` :+1:

surminus:
@npalm @ScottGuymer I've now updated this so that we're just using `var.prefix` everywhere.

modules/runners/pool/main.tf
```diff
@@ -18,7 +18,7 @@ resource "aws_lambda_function" "pool" {
     variables = {
       DISABLE_RUNNER_AUTOUPDATE            = var.config.runner.disable_runner_autoupdate
       ENABLE_EPHEMERAL_RUNNERS             = var.config.runner.ephemeral
-      ENVIRONMENT                          = var.config.environment
+      ENVIRONMENT                          = lookup(var.config.tags, "ghr:environment")
```
ScottGuymer, line 21: Should this be `var.prefix`?

ScottGuymer:
```suggestion
      ENVIRONMENT                          = var.config.prexif
```

ScottGuymer (changes requested):
Just one small thing to update

ScottGuymer:
closes #1280

ScottGuymer (dismissed):

modules/runners/scale-down.tf
```diff
@@ -21,7 +21,7 @@ resource "aws_lambda_function" "scale_down" {
 
   environment {
     variables = {
-      ENVIRONMENT                          = var.environment
+      ENVIRONMENT                          = var.prefix
```
npalm, line 24: Removing the environment variable "ENVIRONMENT" breaks the scae down functionality of the lambda. The lambda is scaling down only for resources having this tag.

modules/runners/scale-up.tf
```diff
@@ -19,7 +19,7 @@ resource "aws_lambda_function" "scale_up" {
       ENABLE_EPHEMERAL_RUNNERS             = var.enable_ephemeral_runners
       ENABLE_JOB_QUEUED_CHECK              = local.enable_job_queued_check
       ENABLE_ORGANIZATION_RUNNERS          = var.enable_organization_runners
-      ENVIRONMENT                          = var.environment
+      ENVIRONMENT                          = var.prefix
```
npalm, line 22: The variable is used in the lambda, removing the varialbe is not possible without refacoring the lambda.

modules/runners/pool/main.tf
```diff
@@ -18,7 +18,7 @@ resource "aws_lambda_function" "pool" {
     variables = {
       DISABLE_RUNNER_AUTOUPDATE            = var.config.runner.disable_runner_autoupdate
       ENABLE_EPHEMERAL_RUNNERS             = var.config.runner.ephemeral
-      ENVIRONMENT                          = var.config.environment
+      ENVIRONMENT                          = var.config.prefix
```
npalm, line 21: The pool is using more of less the same logic as the scale-up lambda. The environment is used for listing and tagging. The tag is most likely informational. The listing is the same as in the scale down.

modules/runners/scale-down.tf
surminus: The environment variable isn't being removed, we're just changing the variable that's setting it. Can I clarify what change you would like to see here?

surminus:
> I also would prefer not to dropt direct the environment variable in the the variables.tf But see if we can provide a better migration part the consumers of the module. For example using a terraform validation rule for varialbes to indicate users how to migrate.

I originally wrote this using an optional prefix, but was [then asked to remove the environment variable](https://github.com/philips-labs/terraform-aws-github-runner/pull/1858#issuecomment-1075334643) completely (unless I misunderstood).

Can I confirm, you would like to see the `environment` variable added back in, but with a validation rule that always evaluates to `false`, and explains how to migrate? I'll look at adding a commit that does this.

Can I also confirm about the Lambda `ENVIRONMENT` variable? This is still being passed a value.

npalm:
> > I also would prefer not to dropt direct the environment variable in the the variables.tf But see if we can provide a better migration part the consumers of the module. For example using a terraform validation rule for varialbes to indicate users how to migrate.
> 
> I originally wrote this using an optional prefix, but was [then asked to remove the environment variable](https://github.com/philips-labs/terraform-aws-github-runner/pull/1858#issuecomment-1075334643) completely (unless I misunderstood).
> 
> Can I confirm, you would like to see the `environment` variable added back in, but with a validation rule that always evaluates to `false`, and explains how to migrate? I'll look at adding a commit that does this.
> 
> Can I also confirm about the Lambda `ENVIRONMENT` variable? This is still being passed a value.

I merged the branch in a local branch up-to-date with development, for some reason something went wrong and I was in the impression the ENVIRONMENT variable was gone for the lambda's. Which isn't so. that comment can be ingored. Will resolve the comments.

Indeed, I would like to have the variable back for some time. So when uers are upgrading and have the environment set the got a clear messag how to migrate to the prefix. Or any other option that avoid users to get a hard time. I prefer to avoid a breaking release as well.

modules/runners/pool/main.tf
npalm: sorry false comment

modules/runners/scale-down.tf
npalm: sorry false comment

surminus:
OK, I can think of two options:

1. Add a validation rule that forces a change to the `environment` variable, something like:
```
variable "environment" {
  description = "A name that identifies the environment, used as prefix and for tagging."
  type        = string
  default     = null

  validation {
    condition     = var.environment == null
    error_message = "The \"environment\" variable is no longer used. To migrate, set the \"prefix\" variable to the original value of \"environment\" and optionally, add \"Environment\" to the \"tags\" variable map with the same value."
  }
}
```

2. Add the environment variable back in, and add a conditionals like `var.environment != null ? var.environment : var.prefix` everywhere. This is basically what my first commit looked like: https://github.com/philips-labs/terraform-aws-github-runner/pull/1858/commits/3179c196bd22945dda9f461c27601d98283f554a

I appreciate you don't want breaking changes, but if the intention is to eventually remove the `environment` variable entirely, then there'll have to be a breaking change at some point. Can you let me know which is your preference?

npalm:
> OK, I can think of two options:
> 
> 1. Add a validation rule that forces a change to the `environment` variable, something like:
> 
> ```
> variable "environment" {
>   description = "A name that identifies the environment, used as prefix and for tagging."
>   type        = string
>   default     = null
> 
>   validation {
>     condition     = var.environment == null
>     error_message = "The \"environment\" variable is no longer used. To migrate, set the \"prefix\" variable to the original value of \"environment\" and optionally, add \"Environment\" to the \"tags\" variable map with the same value."
>   }
> }
> ```
> 
> 2. Add the environment variable back in, and add a conditionals like `var.environment != null ? var.environment : var.prefix` everywhere. This is basically what my first commit looked like: [3179c19](https://github.com/philips-labs/terraform-aws-github-runner/commit/3179c196bd22945dda9f461c27601d98283f554a)
> 
> I appreciate you don't want breaking changes, but if the intention is to eventually remove the `environment` variable entirely, then there'll have to be a breaking change at some point. Can you let me know which is your preference?

Yes agree, prefer the validation rule. Can you also rebase your branch with the origin?

surminus:
I've added a commit that includes the validation rule, and also rebased.

modules/runners/main.tf
```diff
@@ -1,15 +1,15 @@
 locals {
   tags = merge(
     {
-      "Name" = format("%s-action-runner", var.environment)
+      "Name" = format("%s-action-runner", var.prefix)
     },
     var.tags,
```
npalm, line 6: As result of refactoring the tag "Environmet is not injected and added to instances". Which breaks functionality in the lambda's.

main.tf
```diff
@@ -1,7 +1,6 @@
 locals {
   tags = merge(var.tags, {
-    Environment       = var.environment,
```
npalm, line 3: Removing this tag creates problems in the lambda. At least the scale down is not proper working anymore.

npalm (changes requested):
Greate work! But removing the tag "Enviornment" vorm the instances is causing probems in the lambda. Scale Down is not working, and also issues in listing instance to make scale up decisions. 

## Migration test

First got this response: 
```
╷
│ Error: Invalid value for variable
│
│   on main.tf line 22, in module "runners":
│   22:   environment = local.environment
│
│ The "environment" variable is no longer used. To migrate, set the "prefix" variable to the original value of "environment" and optionally, add
│ "Environment" to the "tags" variable map with the same value.
│
│ This was checked by the validation rule at ../../variables.tf:27,3-13.
```

Next replaced the environment variable by prefix
```
Plan: 0 to add, 31 to change, 0 to destroy.
```

surminus:
OK, will investigate the Lambdas and see how they use the `Environment` tag :+1:

npalm:
See: https://github.com/surminus/terraform-aws-github-runner/pull/1

Tested: scaling up and down works, pool not checked.

surminus:
Great, thank you! I've merged that in :tada:

npalm (commented):
Will run once more test today

npalm (approved):
LGTM!

Tested:
- scale down and up
- pool with ephemeral
- migration

surminus:
@npalm amazing thank you for spending the time to test it!

ScottGuymer:
🚀 !

sammort:
Thanks for change @surminus my current Organization (with whom I think you are familiar 😉 ) will find this very useful.

surminus:
> Thanks for the change @surminus my current Organization (with whom I think you are familiar wink ) will find this very useful.

You're welcome @sammort, hope all is well over there!

PR #1739, opened 2022-02-10
feat: Support AWS 4.x Terraform provider

shashidhar087:
Fixes #1739 

Terraform released new version of AWS provider 4.0.0 today one hour ago.  The version mentioned in our terraform is version: ">= 3.38". terraform init will install AWS version 4.0.0.  The version 4.0.0 is a breaking change for S3 resource created in "runner-binaries-syncer" module. Some of the arguments like "acl", "lifecycle_rule", "lifecycle_configuration" became "read_only" in new version.

Reference - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket

PR #1915, opened 2022-04-07
feat: improve syncer s3 kms encryption

julada:
**this pr also includes the terraform provider update. [This](https://github.com/philips-labs/terraform-aws-github-runner/pull/1739) PR should be merged first.**

our cloud security tool requires us to set a bucket policy to enforce the kms encryption:

* create bucket policy when encryption is activated
* explicitly set encryption params in lambda if set
* allow lambda to use the kms key


npalm:
Thx for you contribution, due to conflicts I have merged the changes manual in #2013 (add you as co author).

Issue #1926, opened 2022-04-10
runner-binaries-syncer : "source_hash" is not expected here


AliAllomani:
"aws_s3_bucket_object" "trigger" :  `etag` was replaced with `source_hash` in #1783.

However, terraform provider version is still locked to `v3.38` which doesn't have `source_hash` parameter for `aws_s3_bucket_object`

https://github.com/philips-labs/terraform-aws-github-runner/blob/e3e63367178eb0c65062268555c04ad1f6fb0f83/modules/runner-binaries-syncer/versions.tf#L4-L8

https://registry.terraform.io/providers/hashicorp/aws/3.38.0/docs/resources/s3_bucket_object

```
│ Error: Unsupported argument
│ 
│   on .terraform/modules/runners/modules/runner-binaries-syncer/runner-binaries-syncer.tf line 126, in resource "aws_s3_bucket_object" "trigger":
│  126:   source_hash = filemd5("${path.module}/trigger.json")
│ 
│ An argument named "source_hash" is not expected here.
```

PR #1937, opened 2022-04-13
fix: Set the minimal AWS provider to 3.50

npalm:
See for details #1926 

Set minimal provider to 3.50 to support source_hash for AWS S3 object.


npalm:
close: #1926

PR #1956, opened 2022-04-21
feat: Make userdata gzipped in case the user provided userdata is long

Makeshift:
I provide big userdata, `aws_launch_template` bombs out because user_data is simply base64ified and not gzipped, making it larger than 16kb.

This PR should fix that issue. ~I'm testing on my infrastructure now so will let you know if it works :)~

Appears to work fine, though note I am editing out the exec descriptor stuff (#1959) in the fork I'm running on my infrastructure (https://github.com/Makeshift/terraform-aws-github-runner/commit/16d4c04c3f30e9890830eec7e4df6a517ee553f0)

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -40,6 +40,8 @@ locals {
 }
 
 data "template_cloudinit_config" "userdata" {
+  count = var.enabled_userdata ? 1 : 0
+
   gzip          = true
   base64_encode = true
   part {
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -142,7 +142,7 @@ resource "aws_launch_template" "runner" {
   }
 
 
-  user_data = var.enabled_userdata ? data.template_cloudinit_config.userdata.rendered : ""
+  user_data = var.enabled_userdata ? data.template_cloudinit_config.userdata[0].rendered : ""
 
   tags = local.tags
 
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -40,7 +40,7 @@ locals {
 }
 
 data "template_cloudinit_config" "userdata" {
-  gzip          = false
+  gzip          = true
   base64_encode = true
   part {
     filename     = "00-init"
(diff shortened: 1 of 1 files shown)


Makeshift:
Amusingly, I've since modified my userdata to be less than 16kb when base64'd, but I still think this is a beneficial change.

npalm:
> Amusingly, I've since modified my userdata to be less than 16kb when base64'd, but I still think this is a beneficial change.

Agree, but still need to test. Was also thinking should we provide it configurable. With defaults set to zipped

ScottGuymer:
I have not tried this myself yet but reading the docs/issues I'm not sure this will work for a launch template.

I came across this https://github.com/hashicorp/terraform-provider-aws/issues/497

On an instance you have [user_data](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#user_data) and [user_data_base64](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#user_data_base64). The former explicitly states not to pass gzipped data to it, the latter accepts it.

There is no such distinction on a launch template and no matching gzip accepting variant of [user_data](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#user_data).

Makeshift:
> I have not tried this myself yet but reading the docs/issues I'm not sure this will work for a launch template.

I also thought this, as I vaguely remembered that launch templates didn't support compressed base64 (in one of my commits I even remove the gzip because I thought it wasn't supported, then I realised it worked and put it back), but it absolutely does work in my environment (eu-west-2), so it may have been a recent change to cloudinit?

Makeshift:
Is anybody able to attempt this in another region to see if it's consistently supported, or are we considering it unsupported (because the docs say it should be) and I should therefore remove the gzip?

It's known working in `eu-west-2` only at the moment.

npalm:
I will run a test as well today.

ScottGuymer:
There is a terraform function for base64gzip that we could use.

https://www.terraform.io/language/functions/base64gzip

PR #1962, opened 2022-04-22
docs: specify the `private` for subnet_ids field

chenrui333:
Specify `private` for subnet_ids field in the module

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ module "github-runner" {
 
   aws_region = "eu-west-1"
   vpc_id     = "vpc-123"
-  subnet_ids = ["subnet-123", "subnet-456"]
+  subnet_ids = ["private-subnet-123", "private-subnet-456"]
 
   environment = "gh-ci"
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,7 +167,7 @@ module "github-runner" {
 
   aws_region = "eu-west-1"
   vpc_id     = "vpc-123"
-  subnet_ids = ["subnet-123", "subnet-456"]
+  subnet_ids = ["private-subnet-123", "private-subnet-456"]
```
npalm, line 170: This is just an example, so needs completely replaced by actual values, typically something like subnet-<id>

README.md
```diff
@@ -472,7 +472,7 @@ In case the setup does not work as intended follow the trace of events:
...
 | <a name="input_scale_down_schedule_expression"></a> [scale\_down\_schedule\_expression](#input\_scale\_down\_schedule\_expression) | Scheduler expression to check every x for scale down. | `string` | `"cron(*/5 * * * ? *)"` | no |
 | <a name="input_scale_up_reserved_concurrent_executions"></a> [scale\_up\_reserved\_concurrent\_executions](#input\_scale\_up\_reserved\_concurrent\_executions) | Amount of reserved concurrent executions for the scale-up lambda function. A value of 0 disables lambda from being triggered and -1 removes any concurrency limitations. | `number` | `1` | no |
-| <a name="input_subnet_ids"></a> [subnet\_ids](#input\_subnet\_ids) | List of subnets in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`. | `list(string)` | n/a | yes |
+| <a name="input_subnet_ids"></a> [subnet\_ids](#input\_subnet\_ids) | List of private subnet ids in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`. | `list(string)` | n/a | yes |
```
npalm, line 475: Private subnets is a choice of the consumer, not mandated by the module.

npalm (approved):

README.md
npalm: Al terraform variables in the readme files are generated via terraform docs based on the actuals in variables.tf

npalm:
Thanks for taking the time to make a PR. The subnets are not managed by the module. If they are private or not is the choice of the user. So no need to make this explicit. For the default example we use indeed an example based on a public / private subnet setup.

npalm:
You are ok with closing the PR?

npalm:
Closing feel free to re-open when you think it still need to updated

PR #1964, opened 2022-04-23
docs: ensure webhook event content-type is `application/json`

chenrui333:
Update docs to explicitly call out webhook event `content-type`.


npalm (approved):
Thanks, LGTM

PR #1965, opened 2022-04-25
fix: Output the role for the pool management

ScottGuymer:
Sometimes extra policies need to be attached to roles such as for custom KMS encryption of EBS. We need to output the pool lambda role so it can be configured.


npalm (approved):
Thanks!

Issue #1954, opened 2022-04-18
`volume_size` being ignored and runner starts with minimum storage


gusfune:
I am running a setup with `0.40.0` which uses ephemeral runners but in this setup it seems the variable `volume_size` is being ignored. I have set it up for a value of `40`, but it only deploys instances with 8Gb. Also running ARM instances in this setup.

<img width="1167" alt="Screenshot 2022-04-18 at 18 36 41" src="https://user-images.githubusercontent.com/1147240/163881786-ecc7bcb6-b380-4902-91a9-7c40cd40ecc8.png">

I don't have any custom block configuration or any heavy customisation. Pasting my config from `main.tf` file below:
```terraform
module "runners" {
  source                          = "philips-labs/github-runner/aws"
  version                         = "0.40.0"
  create_service_linked_role_spot = true
  aws_region                      = local.aws_region
  vpc_id                          = module.vpc.vpc_id
  subnet_ids                      = module.vpc.private_subnets

  environment = local.environment

  github_app = {
    key_base64     = nonsensitive(OMITTED)
    id             = nonsensitive(OMITTED)
    webhook_secret = nonsensitive(OMITTED)
  }

  # Grab zip files via lambda_download
  runner_binaries_syncer_lambda_zip = "lambdas-download/runner-binaries-syncer.zip"
  runners_lambda_zip                = "lambdas-download/runners.zip"
  webhook_lambda_zip                = "lambdas-download/webhook.zip"

  # Runner setup
  enable_organization_runners = true
  enable_ssm_on_runners       = true
  instance_types              = ["t4g.large", "c6g.large"]
  runner_architecture         = "arm64"
  runner_extra_labels         = "default,aws"
  runner_os                   = "linux"
  runners_maximum_count       = 8
  volume_size                 = 40

  # ephermeral instance rules
  delay_webhook_event             = 0
  enable_job_queued_check         = true
  fifo_build_queue                = false
  job_queue_retention_in_seconds  = 300
  minimum_running_time_in_minutes = 1
  redrive_build_queue = {
    enabled         = true
    maxReceiveCount = 3
  }

  # override scaling down
  scale_down_schedule_expression = "cron(* * * * ? *)"

  tags = {
    Project = "GitHubAction"
  }
}

resource "aws_iam_service_linked_role" "spot" {
  aws_service_name = "spot.amazonaws.com"
}

# Other parts omitted as they are unrelated.
```

npalm:
[...]

PR #2013, opened 2022-05-04
feat: AWS Terraform provider 4.x

npalm:
Branch to merge PR to support Terraform Provider 4.x

See #1938

PR #2014, opened 2022-05-04
fix: Volume size is ingored

npalm:
The problen is most likely introduces in one of the PR's to support multiple ebs.

The following options for volumes are now supported.
- Default, create an ebs for amazon linux
- Provide explicit ebs mappings via block_device_mappings
- No ebs, set block_device_mappings to []
- Deprecated! Overwrite the default volume size by setting volume_size

close: #1954

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -64,7 +64,7 @@ resource "aws_launch_template" "runner" {
       ebs {
         delete_on_termination = lookup(block_device_mappings.value, "delete_on_termination", true)
         volume_type           = lookup(block_device_mappings.value, "volume_type", "gp3")
-        volume_size           = var.volume_size == null ? lookup(block_device_mappings.value, "volume_size", "30"): var.volume_size
+        volume_size           = var.volume_size == null ? lookup(block_device_mappings.value, "volume_size", "30") : var.volume_size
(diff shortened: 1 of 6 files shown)


JeroenKnoops (approved):
LGTM

modules/runners/main.tf
```diff
@@ -64,7 +64,7 @@ resource "aws_launch_template" "runner" {
       ebs {
         delete_on_termination = lookup(block_device_mappings.value, "delete_on_termination", true)
```
ScottGuymer, line 65: I think we should take the opportunity to remove this complexity here and just use the array of block mappings and remove the lookups.

npalm: Shall we do this in a separate PR. And combine that PR with the breaking changes in #2013

ScottGuymer: Can we make use of https://www.terraform.io/language/functions/defaults to help with this?

npalm: This feature is still experimental, so for that reason I was not using them yet.

npalm: In this PR only trying to fix broken functionality

ScottGuymer: ok np

ScottGuymer (approved):
:shipit:

Issue #2017, opened 2022-05-05
No explicit block is causing errors


gusfune:
After the release of 0.40.3 I have noticed that the runners were not triggering when having no `block_device_mapping` set in the terraform file.

The error show in AWS was:
> ErrorMessage: 'Invalid device name /dev/xvd'

After some digging the issue found was that this block:
```
default = [{
    device_name           = "/dev/xvd"
    delete_on_termination = true
    volume_type           = "gp3"
    volume_size           = 30
    encrypted             = true
    iops                  = null
  }]
```

Actually should be
```
default = [{
    device_name           = "/dev/xvda"
    delete_on_termination = true
    volume_type           = "gp3"
    volume_size           = 30
    encrypted             = true
    iops                  = null
  }]
```

I managed to fix my `main.tf` file by setting this block below:
```
block_device_mappings = [{
    device_name           = "/dev/xvda"
    delete_on_termination = true
    volume_type           = "gp3"
    volume_size           = 30
    encrypted             = true
    iops                  = null
  }]
``

But, the default configuration should be reviewed. This was caused by #2014 to fix the issue I have reported on #1954.

npalm:
Thx! Just find the same issue.

gusfune:
I have already pushed a fix for this @npalm on #2018 if that helps.

npalm:
@gusfune thx, tested and one file was missed for the fix. Provided another fix on top of yours. Thanks for the help!

PR #2018, opened 2022-05-05
fix: block name should default to `dev/xvda`

gusfune:
This fixes the issue reported on #2017.

PR #2019, opened 2022-05-05
fix: Wrong block device mapping

npalm:
Adding some small fixes on top of PR #2018

close: #2017


npalm:
Tested end-2-end with default example

Screenshot 2022-05-05 at 21 36 44

JeroenKnoops (approved):

PR #2024, opened 2022-05-09
feat: Add option to enable detailed monitoring for runner launch template

mckern:
Howdy. We had a need for [detailed monitoring](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-cloudwatch-new.html) for our ephemeral runner instances, and thought others may have a use for it as well. I've done my best to plumb it through and it's working in our local environment but if there's anything y'all would like changed to better suit the overall structure for this module, I'm happy to hammer it out.


npalm (approved):
@mckern Thanks! Checked and look all good.

image

Issue #1737, opened 2022-02-10
Cannot disable cloudwatch agent


oradwell:
When "enable_cloudwatch_agent" is set to "false" the runners fail to set up

```
Feb 10 17:03:40 user-data: Retrieved /gh-ci/runner/run-as parameter - (ec2-user)
Feb 10 17:03:40 user-data: Retrieved /gh-ci/runner/enable-cloudwatch parameter - (false)
Feb 10 17:03:40 user-data: Retrieved /gh-ci/runner/agent-mode parameter - (persistent)
Feb 10 17:03:40 user-data: Cloudwatch is enabled
Feb 10 17:03:40 user-data: ****** processing amazon-cloudwatch-agent ******
Feb 10 17:03:40 user-data: /opt/aws/amazon-cloudwatch-agent/bin/config-downloader --output-dir /opt/aws/amazon-cloudwatch-agent
/etc/amazon-cloudwatch-agent.d --download-source ssm:gh-ci-cloudwatch_agent_config_runner --mode ec2 --config /opt/aws/amazon
-cloudwatch-agent/etc/common-config.toml --multi-config default
Feb 10 17:03:40 user-data: Region: eu-west-1
Feb 10 17:03:40 user-data: credsConfig: map[]
Feb 10 17:03:40 user-data: Error in retrieving parameter store content: ParameterNotFound: 
Feb 10 17:03:40 user-data: Fail to fetch/remove json config: ParameterNotFound: 
Feb 10 17:03:40 user-data: 
Feb 10 17:03:40 user-data: Fail to fetch the config!
```

I know how to fix it so I'll create a PR soon

PR #2051, opened 2022-05-16
fix(runner): Don't treat $enable_cloudwatch_agent=="false" as true in Powershell

kring:
#1737 reported a bug where the userdata script tried to enable the CloudWatch Agent even though it was specified to be disabled. The problem was fixed for `start-runner.sh` in #1738, but the Powershell version has the same bug. This PR fixes the Powershell version.


ScottGuymer (approved):
Nice catch! Thanks.

Issue #1887, opened 2022-03-25
Feature: Expose throughput for block device mappings


Octogonapus:
This is a feature request to please expose the throughput parameter for block device mappings. It will be [this parameter](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#throughput).

npalm:
We expose the parameter, see iops https://github.com/philips-labs/terraform-aws-github-runner#input_block_device_mappings

PR #2052, opened 2022-05-16
feat: Add options extra option to ebs block device mapping

npalm:
## Enhancement

Added options missing options to ebs block device mapping. 
- `throughput`
- `kms_key_id`
- `snapshot_id`

## Migration
Since we not using the Terraform experimental features an upgrade can trigger a change for anyone specifying the variable `block_device_mappings`. In that case you can simply add the three new options with a value null.


npalm:
Close: #1887

sblack4 (approved):
being able to configure `throughput` was going to be my next request. Thank you! It looks good

npalm:
@ScottGuymer can you review the PR?

lorenzo:
I just wanted to say that I would love to see this merged :)

npalm:
@ScottGuymer please can you check this PR?

examples/ubuntu/main.tf
```diff
@@ -64,6 +64,9 @@ module "runners" {
     volume_size           = 30
     encrypted             = true
     iops                  = null
+    throughput            = null
```
ScottGuymer, line 67: What happens if I don't explicitly set these to `null`?

Can/should we use the `defaults` function to set them?

npalm: Since we have no optional object paremeters yet, this is a way to set defaults. Indeed another option is the default function, but in most cases we are using this way of setting defaults. 

For the end user it remains the same. When providing a block_dvice _mapping all parameters needs to be set.

ScottGuymer (approved):

Issue #2072, opened 2022-05-20
Support Node 16


npalm:
Update module to Node 16 since this is now supported by [AWS](https://aws.amazon.com/about-aws/whats-new/2022/05/aws-lambda-adds-support-node-js-16/)

Currently some unit test are failing while bumping Node

PR #2073, opened 2022-05-21
feat: Support Node16 for AWS Lambda

npalm:
AWS now support Node 16 as Lambda runtime. This PR prepares the module to shift to Node16 but will not enable it by default. Upgrade to node requires [AWS Terraform provide 4.15](https://github.com/hashicorp/terraform-provider-aws/releases/tag/v4.15.0)

- Fix unit test to run with Node 16
- Add option to set runtime per Lambda
- Update terraform providers in examples
- Add CI job to check Lambda against node 14 and 16

The new runtime is not enabled by default, see #2074


npalm:
close #2072

JeroenKnoops (approved):

PR #2121, opened 2022-06-07
feat: Support arm64 lambda functions

### LINKED ISSUE #2119: Lambda functions should support running on arm64 architecture
Nodejs 14.x is fully supported on arm64 and this is more cost-effective than x86_64. 

https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-arm-based-aws-graviton2-processors/

dylanmtaylor:
This feature enables the use of Amazon's Graviton ARM64 Lambda functions. I tested this in our AWS environment by changing the architecture of the lambdas. It seems to still work without issues. This is desirable to us as a cost-savings measure.

image

Closes #2119


modules/runner-binaries-syncer/variables.tf
```diff
@@ -187,3 +187,9 @@ variable "lambda_runtime" {
   type        = string
   default     = "nodejs14.x"
 }
+
+variable "lambda_architecture" {
```
ScottGuymer, line 191: Can you add some validation for the possible values here?

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#architectures

variables.tf
```diff
@@ -641,3 +641,9 @@ variable "lambda_runtime" {
   type        = string
   default     = "nodejs14.x"
 }
+
+variable "lambda_architecture" {
+  description = "AWS Lambda architecture."
```
ScottGuymer, line 646: Can you expand on this description a little to explain why you might want to set this value and what the possible options are to help make the documentation more robust.

ScottGuymer (commented):
Looks good. Just a couple of small comments.

Is this worth an addition to the readme to explain why you might set this.

I would also be curious to understand what the potential cost savings are here and if we might want to make arm64 the default.

ScottGuymer:
You also need to run `terraform fmt`

dylanmtaylor:
> I would also be curious to understand what the potential cost savings are here and if we might want to make arm64 the default.

I'm not sure about the actual numbers, but for any workload that supports it, the Graviton2 systems are almost always cheaper than x86_64. Amazon themselves claim "34% better price/performance", so it's sizable. I am fine with making it the default and will update the PR. I just wasn't sure if anyone would be against that. https://aws.amazon.com/about-aws/whats-new/2021/09/better-price-performance-aws-lambda-functions-aws-graviton2-processor/

modules/runner-binaries-syncer/variables.tf
dylanmtaylor: Done :)

dylanmtaylor:
> Is this worth an addition to the readme to explain why you might set this.

Maybe. Honestly, as long as it works, and I haven't seen any evidence there are issues with it, arm64 seems to be a sane default.

npalm:
Looks good, need to test later this week. Please can you also update the pool module? See https://github.com/philips-labs/terraform-aws-github-runner/tree/develop/modules/runners/pool

dylanmtaylor:
> Looks good, need to test later this week. Please can you also update the pool module? See https://github.com/philips-labs/terraform-aws-github-runner/tree/develop/modules/runners/pool

Sure thing. I'll do that now.

dylanmtaylor:
I updated it and think it's good to go now, I thought about rebasing and squashing those commits but figured it'd be preferable to have the commit history and the maintainers could squash it down at merge time.

modules/runner-binaries-syncer/variables.tf
```diff
@@ -187,3 +187,13 @@ variable "lambda_runtime" {
...
+  description = "AWS Lambda architecture. Lambda functions using Graviton processors ('arm64') tend to have better price/performance than 'x86_64' functions. "
+  type        = string
+  default     = "arm64"
```
npalm, line 194: I prefer to keep the default the same as the current default architecture
```suggestion
  default     = "x86_64"
```

modules/runners/variables.tf
```diff
@@ -551,3 +551,13 @@ variable "lambda_runtime" {
   type        = string
   default     = "nodejs14.x"
 }
+
+variable "lambda_architecture" {
+  description = "AWS Lambda architecture. Lambda functions using Graviton processors ('arm64') tend to have better price/performance than 'x86_64' functions. "
+  type        = string
+  default     = "arm64"
```
npalm, line 558: I prefer to keep the default the same as the current default architecture
```suggestion
  default     = "x86_64"
```

modules/runners/pool/main.tf
```diff
@@ -8,6 +8,7 @@ resource "aws_lambda_function" "pool" {
   function_name                  = "${var.config.prefix}-pool"
   role                           = aws_iam_role.pool.arn
   handler                        = "index.adjustPool"
+  architectures                  = [var.config.lambda_architecture]
```
npalm, line 11:
```suggestion
  architectures                  = [var.config.lambda.architecture]
```

modules/webhook/variables.tf
```diff
@@ -162,3 +162,13 @@ variable "lambda_runtime" {
   type        = string
   default     = "nodejs14.x"
 }
+
+variable "lambda_architecture" {
+  description = "AWS Lambda architecture. Lambda functions using Graviton processors ('arm64') tend to have better price/performance than 'x86_64' functions. "
+  type        = string
+  default     = "arm64"
```
npalm, line 169: I prefer to keep the default the same as the current default architecture
```suggestion
  default     = "x86_64"
```

variables.tf
```diff
@@ -641,3 +641,13 @@ variable "lambda_runtime" {
   type        = string
   default     = "nodejs14.x"
 }
+
+variable "lambda_architecture" {
+  description = "AWS Lambda architecture. Lambda functions using Graviton processors ('arm64') tend to have better price/performance than 'x86_64' functions. "
+  type        = string
+  default     = "arm64"
```
npalm, line 648: I prefer to keep the default the same as the current default architecture
```suggestion
  default     = "x86_64"
```

dylanmtaylor:
I see the test is failing, sorry about that. Let me try to get it in good shape.

npalm (changes requested):
@dylanmtaylor thanks for the updates. Tested at all works. I prefer to keep the default to x64 for the moment (made code suggestions). Also made a small correction suggestion in the change for the pool

npalm (approved):

PR #2146, opened 2022-06-13
fix: replaced old environment variable

julada:
* removed old usage of environment variable which causes an error on template interpolation
Bildschirmfoto 2022-06-13 um 12 44 38


npalm (approved):
LGTM

Issue #2105, opened 2022-06-02
Fail to create runner-binaries-syncer when S3 SSE is enabled


pinkplus:
I got a 403 error when I tried to appy the terraform configs.

After I did an investigation, I found that it's because I have enabled server-side encryption for the S3 bucket. As a result the terraform module applies a policy to reject any object that is not encrypted (69-103 lines in [main.tf](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runner-binaries-syncer/main.tf)).

On the other hand, the trigger object is not configured to use SSE (135-142 lines in [runner-binaries-syncer.tf](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runner-binaries-syncer/runner-binaries-syncer.tf)). That's the cause of the 403 error.

I can make a PR to fix this issue if that's ok.

pinkplus:
I wonder what's the right fix. I'm thinking of removing the policy (69-103 lines in [main.tf](https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runner-binaries-syncer/main.tf)), which denies upload if the upload doesn't set a server-side encryption header. If the policy is not created, the object will be encrypted with the default encryption setting when the SSE header is not set. This sounds like a better idea to me than requiring every upload request to have the SSE header.

Does it sound reasonable?

PR #2147, opened 2022-06-13
fix: set kms key on aws_s3_object when encryption is enabled

julada:
* fixes an error during apply on aws_s3_object when encryption is enabled
* set kms key id when encryption is enabled
* dont set unsupported etag property when encryption is enabled

### DIFF
--- a/modules/runner-binaries-syncer/runner-binaries-syncer.tf
+++ b/modules/runner-binaries-syncer/runner-binaries-syncer.tf
@@ -137,8 +137,8 @@ resource "aws_s3_object" "trigger" {
   bucket     = aws_s3_bucket.action_dist.id
   key        = "triggers/${aws_lambda_function.syncer.id}-trigger.json"
   source     = "${path.module}/trigger.json"
-  etag       = lookup(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default, "kms_master_key_id", null) == null ? filemd5("${path.module}/trigger.json") : null
-  kms_key_id = lookup(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default, "kms_master_key_id", null)
+  etag       = try(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default.kms_master_key_id, null) == null ? filemd5("${path.module}/trigger.json") : null
+  kms_key_id = try(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default.kms_master_key_id, null)
 
   depends_on = [aws_s3_bucket_notification.on_deploy]
 }
--- a/modules/runner-binaries-syncer/runner-binaries-syncer.tf
+++ b/modules/runner-binaries-syncer/runner-binaries-syncer.tf
@@ -134,10 +134,11 @@ resource "aws_lambda_permission" "syncer" {
 ###################################################################################
 
 resource "aws_s3_object" "trigger" {
-  bucket = aws_s3_bucket.action_dist.id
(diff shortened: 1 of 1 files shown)


npalm:
Reference to docs: https://registry.terraform.io/providers/hashicorp%20%20/aws/latest/docs/resources/s3_object#etag

npalm:
fix: #2105

modules/runner-binaries-syncer/runner-binaries-syncer.tf
```diff
@@ -134,10 +134,11 @@ resource "aws_lambda_permission" "syncer" {
...
 
 resource "aws_s3_object" "trigger" {
-  bucket = aws_s3_bucket.action_dist.id
-  key    = "triggers/${aws_lambda_function.syncer.id}-trigger.json"
-  source = "${path.module}/trigger.json"
-  etag   = filemd5("${path.module}/trigger.json")
+  bucket     = aws_s3_bucket.action_dist.id
+  key        = "triggers/${aws_lambda_function.syncer.id}-trigger.json"
+  source     = "${path.module}/trigger.json"
+  etag       = lookup(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default, "kms_master_key_id", null) == null ? filemd5("${path.module}/trigger.json") : null
```
npalm, line 140: Apply is failing in the case SSE is not configured since `server_side_encryption_configuration` is not set. By using`try` this can be avoided.
```suggestion
  etag       = try(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default.kms_master_key_id, null) == null ? filemd5("${path.module}/trigger.json") : null
```

modules/runner-binaries-syncer/runner-binaries-syncer.tf
```diff
@@ -134,10 +134,11 @@ resource "aws_lambda_permission" "syncer" {
...
+  kms_key_id = lookup(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default, "kms_master_key_id", null)
```
npalm, line 141: Apply is failing in the case SSE is not configured since `server_side_encryption_configuration` is not set. By using`try` this can be avoided.
```suggestion
  kms_key_id = try(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default.kms_master_key_id, null)
```

npalm (changes requested):
@julada tested with the default example, got an error. See fixed in the comments.

modules/runner-binaries-syncer/runner-binaries-syncer.tf
julada: You are absolutely right, i should have checked that. Thanks for fixing it. I will apply your suggestion.

npalm (approved):

npalm:
@julada thanks!

shashidhar087:
@julada Could you pls provide an example of where to set KMS Key ID. I am running into this issue where I have enabled SSE encryption and not KMS type on S3.

julada:
Can you provide an example of your actual issue?

julada:
Ah i got it. Pls try to add the server_side_encryption property on the aws_s3_object ressource. It must be set to the configured sse algorithm. In your case this attribute should be set to "AES256". If you dont want to create a fix for this case u can switch to kms encryption by adjusting your sse configuration.

shashidhar087:
Thank you for the quick response. Issue got fixed after setting the server_side_encryption key in 'aws_s3_object' resource. 

 server_side_encryption = try(var.server_side_encryption_configuration.rule.apply_server_side_encryption_by_default.sse_algorithm, null)

PR #2182, opened 2022-06-22
fix(tag): Namespace `Application` tag

mcaulifn:
Namespaces the `Application` tag to avoid conflicting with `Application` tag set by policies.

### DIFF
--- a/modules/runners/lambdas/runners/src/aws/runners.ts
+++ b/modules/runners/lambdas/runners/src/aws/runners.ts
@@ -48,7 +48,7 @@ export async function listEC2Runners(filters: ListRunnerFilters | undefined = un
   const ec2Statuses = filters?.statuses ? filters.statuses : ['running', 'pending'];
   const ec2 = new EC2();
   const ec2Filters = [
-    { Name: 'tag:Application', Values: ['github-action-runner'] },
+    { Name: 'tag:ghr:Application', Values: ['github-action-runner'] },
     { Name: 'instance-state-name', Values: ec2Statuses },
   ];
 
@@ -158,7 +158,7 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
           {
             ResourceType: 'instance',
             Tags: [
-              { Key: 'Application', Value: 'github-action-runner' },
+              { Key: 'ghr:Application', Value: 'github-action-runner' },
               { Key: 'Type', Value: runnerParameters.runnerType },
               { Key: 'Owner', Value: runnerParameters.runnerOwner },
             ],


mcaulifn:
@npalm Seems the TF checks are broken?

npalm:
For some reason the CI is failing, can you double check you are in sync? Should not fail since no TF files are changed.

npalm:
Strange, since develop and main are building without an issue. Including todays release.

mcaulifn:
@npalm Looks like one of the policies had some unintended formatting applied. I revered and all checks are passing.

mcaulifn:
@npalm Looks like this might be a one-time break where users will have to clean out old runners. OK to keep as is or do you have suggestions?

npalm:
@mcaulifn not a really good idea rigiht now.  Only option coming to my mind is let the scale down lambda be aware of the old and new tag. And remove the old tag in a next major release.

npalm:
@mcaulifn it the PR ready for review?

mcaulifn:
@npalm Should be good now

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -44,24 +44,47 @@ export interface RunnerInputParameters {
...
 
+interface Ec2Filter {
+  Name: string;
+  Values: string[];
+}
+
 export async function listEC2Runners(filters: ListRunnerFilters | undefined = undefined): Promise<RunnerList[]> {
-  const ec2Statuses = filters?.statuses ? filters.statuses : ['running', 'pending'];
-  const ec2 = new EC2();
-  const ec2Filters = [
```
npalm, line 50: Not sure if it is really needed, but back in the days we have implemented here hardcoded that if not set a filter. A filter got enforced. I think this was done to avoid the scale down can terminate anything that is not owned. 

Don't like this hidden magic and we should got cleaned it at some moment. I also updated the tested to see the impact when we remove the depecrated tag `Application`.  In that case you will find that the test for instance we no tags `'Instances with no tags.'` failes. Which is related to these default.s

npalm: The test `Instances with no tags.` on L134 in on this branch. Sorry for the late reply

mcaulifn: `'Instances with no tags.'` should be changed to expect 0 and now it is failing for me.

Despite `mockReturnValueOnce` being used throughout, and not being used in the test in question, I am getting the two instances used at the top of the test. This appears to be an issue with the test rather than with the code. I'm still trying to track down how it is getting injected.

mcaulifn: Seems [this bug](https://github.com/facebook/jest/issues/10967) is being hit. I'll try to work up a fix.

mcaulifn: I have not been able to find a work-around for the testing bug. @npalm any suggestions?

npalm:
@mcaulifn back from PTO, will try to check what I can do on this PR next week.

npalm:
Will dig in the PR soon, we are working on several breaking changes. So would like to got this one merged soon. That gives the option to clean up logic together with #2517

npalm:
@mcaulifn I just started digging in the PR and running some test, keep you posted.

modules/runners/policies/lambda-scale-down.json
```diff
@@ -21,7 +21,7 @@
             ],
             "Condition": {
                 "StringEquals": {
-                    "ec2:ResourceTag/Application": "github-action-runner"
+                    "ec2:ResourceTag/ghr:Application": "github-action-runner"
```
npalm, line 24: Can you add one extra block to allow to remove old tag runners. Without allowing the lambda to terminate we end up first with orphan runners, runner got remove from github, deletion fails. Next the orphan runner cleanup fails as well

Please add this block
```json
        {
          "Effect": "Allow",
          "Action": [
              "ec2:TerminateInstances"
          ],
          "Resource": [
              "*"
          ],
          "Condition": {
              "StringEquals": {
                  "ec2:ResourceTag/Application": "github-action-runner"
              }
          }
        },
```

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -45,24 +45,47 @@ export interface RunnerInputParameters {
...
-    { Name: 'tag:Application', Values: ['github-action-runner'] },
-    { Name: 'instance-state-name', Values: ec2Statuses },
-  ];
+  const ec2Filters = constructFilters(filters);
+  const runners: RunnerList[] = [];
+  for (const filter of ec2Filters) {
+    runners.push(...(await getRunners(filter)));
```
npalm, line 57: This line is creating a problem in the tests, once removing the deprecated code, test starts failing. The cause, mock once. Ant that is logical. But the side effect of this change is we start calling the AWS API more often, one time for each filter. Before 1 time for all the filters. 

When you start scaling to a lot fo runners the API rate limites start counting. I like the way y9u you extract the function. But we should find a way to du it in such way we not calling the API more then needed.

npalm:
@mcaulifn found 2 issues, would you have tim for a quick look?

modules/runners/lambdas/runners/src/aws/runners.ts
npalm: I check the tests again.

npalm: At least I know what the problem is. When I remove the support vor the tag Application in (runner.ts). Merge the two mocks back in a single mock like before, returning 2 objects. The test "Instances with no tags." starts failing. Running the test in isolation is not causing a problem. So somehow the mocks are not resetting well. 

Was not able to find a solution, but this eems the issue.

modules/runners/lambdas/runners/src/aws/runners.test.ts
```diff
@@ -142,7 +174,7 @@ describe('list instances', () => {
         },
       ],
     };
-    mockDescribeInstances.promise.mockReturnValue(noInstances);
+    mockDescribeInstances.promise.mockReturnValueOnce(noInstances).mockReturnValue({});
```
npalm, line 177: Missed reverting this line, which causes the test failure

npalm (changes requested):
@mcaulifn PR is good with one small request, can you add the permission as suggested in the comment. Thanks!

modules/runners/lambdas/runners/src/aws/runners.ts
mcaulifn: Is this one good or do you want me to find another solution?

modules/runners/policies/lambda-scale-down.json
mcaulifn: Fixed

modules/runners/lambdas/runners/src/aws/runners.test.ts
mcaulifn: Reverting

modules/runners/policies/lambda-scale-down.json
```diff
@@ -1,48 +1,64 @@
...
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ssm:GetParameter"
+      ],
+      "Resource": [
+        "${github_app_key_base64_arn}",
+        "${github_app_id_arn}"
+      ]
+%{ if kms_key_arn != "" ~
```
npalm, line 51: See your CI is failing, looks like something went wrong adding with this conditonal. Annyoing since json formating typicall destroys it all. Can you check this part?

mcaulifn: This appears to be caused by Editorconfig:

```
terraform-aws-github-runner/modules/runners/policies/lambda-scale-down.json: {"insertSpaces":true,"tabSize":2}
terraform-aws-github-runner/modules/runners/policies/lambda-scale-down.json: Using EditorConfig core...
modules/runners/policies/lambda-scale-down.json: editor.action.trimTrailingWhitespace
```

mcaulifn: I cannot figure out how to make the auto-formatting not remove that. I ended up editing the file in nano.

npalm:
@mcaulifn thanks will have asap another check on the PR!

npalm (approved):
@mcaulifn once again thanks for you support. I already have prepared the PR to remove the legacy tag (#2705)

PR #2207, opened 2022-06-27
fix: added server_side_encryption key in resource aws_s3_object

### LINKED ISSUE #2105: Fail to create runner-binaries-syncer when S3 SSE is enabled

shashidhar087:
fixes #2105 

@julada Here is the PR to fix S3 object trigger when server_side_encryption algorithm type is AES256


npalm (approved):
LGTM, thx

PR #2209, opened 2022-06-27
feat: Webhook accept jobs where not all labels are provided in job.

ivanprjcts:
Thanks for this awesome project! It is really useful :)

I think that `webhook` should allow run jobs in `self-hosted` runners without requiring pass all labels.

Labels are usually intended for describing capabilities (in runners) and requirements (in jobs). If I only want to require that my job is run in a `self-hosted` runner no matters what other capabilities runner have, passing only `self-hosted` label should be enough.

On the other hand, if I require to run my job in a `self-hosted` and `ubuntu` runner using these labels, but I haven't any `ubuntu` runner, it shouldn't be run.

Thank you for your time.

### DIFF
--- a/modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
+++ b/modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
@@ -159,7 +159,7 @@ describe('handler', () => {
       expect(sendActionRequest).toBeCalled();
     });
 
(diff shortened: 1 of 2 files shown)


npalm:
Sorry for keep you wainting, the PR is still on our radar.

npalm (approved):

PR #2214, opened 2022-07-01
fix(images): avoid wrong AMI could be selected for ubuntu focal

LevkovychBorya:
With the previous filter you could also get following ami:
 - ubuntu-eks/k8s_1.18/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*
We are looking specifically for this type:
- ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*
Which is unintended, as far as I understand.


npalm (approved):
@LevkovychBorya thanks for your contribution! LGTM

PR #2103, opened 2022-05-31
chore: Bump ubuntu example to latest LTS, 22.04

dylanmtaylor:

PR #2250, opened 2022-07-14
fix(examples): Upgrading ubuntu example to 22.04

tobiasehlert:
A fresh installation on Ubuntu 22.04 is missing iptables, which is required for rootless docker do work.

- adding iptables apt-get install list
- sorting apt-get install list

### DIFF
--- a/examples/ubuntu/main.tf
+++ b/examples/ubuntu/main.tf
@@ -46,7 +46,7 @@ module "runners" {
   ami_owners        = ["099720109477"] # Canonical's Amazon account ID
 
   ami_filter = {
-    name = ["ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-*"]
+    name = ["ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-*"]
   }
 
   # Custom build AMI, no custom userdata needed.
--- a/examples/ubuntu/templates/user-data.sh
+++ b/examples/ubuntu/templates/user-data.sh
@@ -44,7 +44,7 @@ WantedBy=default.target
 
 EOF
 
-echo export XDG_RUNTIME_DIR=/run/user/$user_id >>/home/$user_name/.profile
+echo export XDG_RUNTIME_DIR=/run/user/$user_id >>/home/$user_name/.bashrc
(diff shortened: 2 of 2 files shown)


npalm (commented):
@tobiasehlert thx for your PR. The example is still on ubuntu20.04 which seems not require the iptables. So suggest we combine the PR with an upgade of the example to ubuntu 22.04. Feel free to update the PR. I leave the PR open for now.

npalm:
Checking now to #2103 to upgrade example to ubuntu 22.04

npalm:
Just tried the example, for some reason the runner cannot find the docker command. On ubuntu-20.04 it all works fine. Tried the PR with and without in combination with #2250

tobiasehlert:
hi @npalm,

I ended up removing docker rootless part and using the "normal" docker as installed from apt after adding the docker repo. I also added some additional steps to add permission to user accounts to use the docker command.
For me that was kind of the best way to get going.

What about adding docker.io as a package to the installing point and not using rootless?
Any reason for us not do to it like that?

tobiasehlert:
@npalm, ah interesting point actually. I've not thought about the volume thing, but we wipe the instances pretty quick, so it's not often they remain even when not using ephemeral runners.

I do actually use an Ansible playbook to setup a instance in the first place and install a couple of software on it (eg. terraform, pip and Docker), since the team want a runner up and running in under 2 minutes. With too much tools getting installed during user-data.sh, the installation takes too long time to finish.
So I have therefore stripped my user-data.sh file quite a lot.

But I can lab today with using rootless and get back to you with my result :)

tobiasehlert:
Ah yeah.. I recall now that docker rootless didn't work for me, since the `/home/$user_name/.profile` file is not read for some reason. I was not able to pinpoint exactly why, so I went for the normal installation instead.

But when I try now to run the export commands to point to `/home/$user_name/.bashrc` instead, it works.

tobiasehlert:
@npalm, so.. not the PR should work better.
I included the changes that are suggested in PR #2103 in the last commit as well, so this update should be complete :)

npalm (approved):
Thx, tested and looks good!

Issue #2210, opened 2022-06-29
Prevent that lambda webhook fails when it tries to process a installation_repositories event


jgutierrezglez:
Hello team,

I hope you're doing well. I noticed the lambda webhook fails when it receives an installation_repositories event from GitHub. Even if in the GitHub app is configured to receive only workflow_job events - this event is still received by the application that we have created to manage our GitHub runners (our application is installed at the organizational level and any project from this organization can use it). 

The error returned by the lambda in this particular scenario is the following:

```
ERROR [webhook:ab4e8da5-4c70-488d-a47f-f1f62cb4356c index.js:91029 Runtime.handler]
TypeError Cannot read property 'full_name' of undefined
error stack:
• index.js:84926 handle
  index.js:84926:63
• index.js:91026 githubWebhook
  index.js:91026:18
```

There is a real quick fix that be can be applied to the webhook lambda to prevent this issue:

https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/webhook/lambdas/webhook/src/webhook/handler.ts#L32

```
  const payload = JSON.parse(body);
  LogFields.fields.event = githubEvent;
  LogFields.fields.repository = payload.repository.full_name;
  LogFields.fields.action = payload.action;

  if (!supportedEvents.includes(githubEvent)) {
    logger.warn(`Unsupported event type.`, LogFields.print());
    return {
      statusCode: 202,
      body: `Ignoring unsupported event ${githubEvent}`,
    };
  }
 
 LogFields.fields.name = payload[githubEvent].name;
 LogFields.fields.status = payload[githubEvent].status;
 LogFields.fields.started_at = payload[githubEvent]?.started_at;
```

if the 4 first lines of the snipped are moved just below the   if (!supportedEvents.includes(githubEvent)) line

```
  if (!supportedEvents.includes(githubEvent)) {
    logger.warn(`Unsupported event type.`, LogFields.print());
    return {
      statusCode: 202,
      body: `Ignoring unsupported event ${githubEvent}`,
    };
  }
  const payload = JSON.parse(body);
  LogFields.fields.event = githubEvent;
  LogFields.fields.repository = payload.repository.full_name;
  LogFields.fields.action = payload.action;
  LogFields.fields.name = payload[githubEvent].name;
  LogFields.fields.status = payload[githubEvent].status;
  LogFields.fields.started_at = payload[githubEvent]?.started_at;
   
```
[...]

PR #2288, opened 2022-07-26
fix: Preventing that lambda webhook fails when it tries to process a installation_repositories event

jgutierrezglez:
https://github.com/philips-labs/terraform-aws-github-runner/issues/2210


npalm (approved):
LGTM

npalm (approved):
@jgutierrezglez thx!

PR #2291, opened 2022-08-01
fix: Webhook lambda misleading log

davidl-talon:
In webhook lambda `handler.ts` there is a misleading log that suggests that the job was queued.
When the action is not in the queued state (e.g. completed or canceled), this log may be misleading.


npalm (approved):
Thanks good catch!

PR #2302, opened 2022-08-02
fix: Update ubuntu example to fix /opt/hostedtoolcache

MarkArts:
This fixes and issue we where having where [actions/setup-python](https://github.com/actions/setup-python) was unable to install python because it didn't have access to `/opt/hostedtoolcache`. 

It seems like the issue comes from the example `user_data.sh` script setting the variable `$USER_NAME` while the `install-runner.sh` uses the variable `$user_name` to set the `/opt/hostedtoolcache` permissions
https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/modules/runners/templates/install-runner.sh#L55 

This also changed the user_name to "ubuntu" to match the `runner_run_as ` as option in: 
https://github.com/philips-labs/terraform-aws-github-runner/blob/develop/examples/ubuntu/main.tf#L40

This issue seems to be related:
https://github.com/actions/setup-python/issues/459

### DIFF
--- a/examples/ubuntu/templates/user-data.sh
+++ b/examples/ubuntu/templates/user-data.sh
@@ -15,9 +15,8 @@ DEBIAN_FRONTEND=noninteractive apt-get install -y \
     build-essential \
     unzip
 
-USER_NAME=runners
-useradd -m -s /bin/bash $USER_NAME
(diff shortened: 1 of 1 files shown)


npalm (approved):
@MarkArts  thanks for fixing this!

PR #2305, opened 2022-08-02
fix: Add delay to prevent ssm rate limits if instance count is greater than max throughput

devsh4:
Fix for bug #1841 which prevents hitting AWS SSM rate limits by adding a delay between subsequent putParameter calls.

**Note**: 25ms delay is based on the [AWS service quotas](https://docs.aws.amazon.com/general/latest/gr/ssm.html#limits_ssm) as they only allow 40 requests per second to the parameter store by default.

**[Tests]**:
- Pre commit checks ✅ 
- Ran build locally ✅ 
- Ran linter ✅ 
- Ran jest tests ✅ 

**Related PR** (never got merged): https://github.com/philips-labs/terraform-aws-github-runner/pull/1843


devsh4:
@npalm Looks like the jest step is failing because the coverage is dropping below 92%, however all the tests itself are passing.

```Jest: "global" coverage threshold for branches (92%) not met: 91.37%```

**Comment by npalm** [comment]
Guess this is caused since condional delay black is not touched in the test. This requires to add a test with adding 40 runners. Right now not more time to dig in further. Will check later.

**Comment by devsh4** [comment]
@npalm Any chance you can look at this? I tried adding some tests with 40 runners but it didn't help.

**Comment by npalm** [comment]
Yes will do. But will be after next week.

**Comment by devsh4** [comment]
Any update on this? Looking to upgrade the module version but cannot until we have this PR merged in.

**Comment by npalm** [comment]
Sorry no updates, bit restricted by time. Sorry.

File: modules/runners/lambdas/runners/src/aws/runners.ts [hunk 1]
```diff
@@ -203,6 +203,14 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
   logger.info('Created instance(s): ', instances.join(','), LogFields.print());
 
   const ssm = new SSM();
+  const delay = async (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
GuptaNavdeep1983, line 206:
```suggestion
throttle(
      async () =>
        await ssm
          .putParameter({
            Name: `${runnerParameters.environment}-${instance}`,
            Value: runnerParameters.runnerServiceConfig.join(' '),
            Type: 'SecureString',
          })
          .promise(),
      isDelay ? 25 : 0,
    );
```

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -203,6 +203,14 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
...
+  const ssmParameterStoreMaxThroughput = 40;
+  let isDelay = false;
+
+  if (instances.length >= ssmParameterStoreMaxThroughput) {
+    isDelay = true;
+  }
+
```
GuptaNavdeep1983, line 213:
```suggestion
  const ssmParameterStoreMaxThroughput = 40;
  let isDelay = false;

  if (instances.length >= ssmParameterStoreMaxThroughput) {
    isDelay = true;
  }

```

GuptaNavdeep1983:
@devsh4 can you have a look at the recent changes?

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -203,13 +204,24 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
   logger.info('Created instance(s): ', instances.join(','), LogFields.print());
 
   const ssm = new SSM();
+  const ssmParameterStoreMaxThroughput = 40;
+  let isDelay = false;
+
+  if (instances.length >= ssmParameterStoreMaxThroughput) {
+    isDelay = true;
+  }
+  const throttled = throttle(
```
npalm, line 213: The lodash throttle function seems not delay but discoarding calls duing the thrutttle time. I think for delay API call sw do not nee a extra labrary. We can use simple the setTimeout fucntion to avoid we hitting the API to hard.

Not sure why we start only throttling above 40, maybe better to start earlier. There is a unit test for creating 2 instances. Which failes when you set the limit 40 to 2. This because putParameter is never called.

npalm:
Ther refered AWS document only mention GetParameters to be rate limitted. The Lambda only calls putParatameter. I can't see a specific limit for the putParamaters. Hoever I have seen RateLimits as well when we try to scale aggressive. AWS also used a bucket to rate limite. As soon you bucket is empty ou got rate limitted. The moudle makes several AWS calls. PutParameter wil certainly tak a part in this. 

But not sure if delaying the SSM put parameter calal will solve the problem. Once solving the proble here, maybe call the GetParamter from the instance start can be the next limit.

modules/runners/lambdas/runners/src/aws/runners.ts
devsh4: I agree with @npalm, we should revert this to what implementation I had before as I had tried something similar before and it didn't work.

devsh4:
> AWS service quotas

@npalm yes I believe `putParameter` call does play a role in hitting rate limits. We are running the [forked version](https://github.com/viewthespace/terraform-aws-github-runner) of this repo with the [initial change I made here](https://github.com/philips-labs/terraform-aws-github-runner/pull/2305/commits/1661e6eab6af9e2beef6c0408ffd3457c0f6d332) and it has been running seamlessly since the past 6 months for spinning up more than ~250 runners at once without hitting rate limits.

npalm:
> > AWS service quotas
> 
> @npalm yes I believe `putParameter` call does play a role in hitting rate limits. We are running the [forked version](https://github.com/viewthespace/terraform-aws-github-runner) of this repo with the [initial change I made here](https://github.com/philips-labs/terraform-aws-github-runner/pull/2305/commits/1661e6eab6af9e2beef6c0408ffd3457c0f6d332) and it has been running seamlessly since the past 6 months for spinning up more than ~250 runners at once without hitting rate limits.

Can you refer to the fork you running here (in case it is OS)?

devsh4:
@npalm Sure forked version is [here](https://github.com/viewthespace/terraform-aws-github-runner), it's linked above as well. Further we are using Linux runners.

devsh4:
@npalm Any update on this?

Issue #2274, opened 2022-07-22
Allow runner-binaries-syncer lambda deployment to be optional


rashidnhm:
We are currently using pre-built AMIs for deploying our runners and it is working well. However, as far as I'm aware, if the runner software is pre-baked into the image, then it will not be pulling a fresh version of the github actions software from S3 on boot, but currently even if you have a pre-built AMI, the runner-binaries-syncer module is still deployed, which creates the syncer lambda and the distribution bucket.

I was wondering if a feature could be introduced to disable it? Maybe a new variable? So if you have pre-baked images being produced, no need to create this portion of the infrastructure

npalm:
Making the submodule optional is ineed a good idea.

PR #2314, opened 2022-08-03
feat: Add option to disable lambda to sync runner binaries

npalm:
Create option to disable the deployment of the binary syncer module, typically not needed when having pre-build AMI containing the GitHub runner.

## What is changed?
- Added variable `enable_runner_binaries_syncer` to disable syncer
- Refactored the internals of the runner moduel. When using the internal modules directoy be-aware of breaking changes, you have to set `s3_runner_binaries`.
- Updated example for pre-build AMI to disable the syncer module.
- Narrowed down the bucket permission of the runner to only the relevant distribution.

## Testes
- Default example
- Pre-build example


npalm:
Close #2274

GuptaNavdeep1983 (dismissed):

main.tf
```diff
@@ -134,8 +133,14 @@ module "runners" {
   prefix        = var.prefix
   tags          = local.tags
 
-  s3_bucket_runner_binaries   = module.runner_binaries.bucket
-  s3_location_runner_binaries = local.s3_action_runner_url
+  # s3_bucket_runner_binaries   = var.enable_runner_binaries_syncer ? module.runner_binaries[0].bucket : null
```
GuptaNavdeep1983, line 136: Should have this commented code?

modules/runners/variables.tf
```diff
@@ -566,4 +564,10 @@ variable "lambda_architecture" {
...
     error_message = "`lambda_architecture` value is not valid, valid values are: `arm64` and `x86_64`."
   }
-}
\ No newline at end of file
+}
+
+variable "enable_runner_binaries_syncer" {
+  description = "Option to disable the lambda to sync GitHub runner distribution, usefull when using a pre-build AMI."
+  type        = bool
+  default     = true
```
GuptaNavdeep1983, line 572: Do we want to enable it by default?

npalm: Yes since it is current enabled by design and this way we don't force migration from current users.

GuptaNavdeep1983 (approved):

PR #2315, opened 2022-08-04
fix: Ignore case for runner labels.

npalm:
GitHub is not specifying if labels of runners are case senstive or not. But experiments make it clear that runners accept jobs without checking the cases of labels. This PR will ensure all labels are always set in lower case and all checks are don with lower case.

### DIFF
--- a/modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
+++ b/modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
@@ -129,7 +129,7 @@ describe('handler', () => {
         ...workflowjob_event,
         workflow_job: {
           ...workflowjob_event.workflow_job,
-          labels: ['self-hosted', 'test'],
+          labels: ['self-hosted', 'Test'],
         },
       });
       const resp = await handle(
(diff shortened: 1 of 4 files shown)


modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
```diff
@@ -141,7 +141,7 @@ describe('handler', () => {
     });
 
     it('Check runner labels accept job with mixed order.', async () => {
-      process.env.RUNNER_LABELS = '["linux", "test", "self-hosted"]';
+      process.env.RUNNER_LABELS = '["linux", "TEST", "self-hosted"]';
```
JeroenKnoops, line 144: This is now combined with the mixed order test. I think it does not matter too much, but just saying.

JeroenKnoops (approved):
Small remark on a combined test, but nothing major.

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
npalm: No, does not matter. Just took one random label and swapped cases. Also forced via Terraform all labels are convert to lower case as well.

PR #2325, opened 2022-08-11
fix(runners): Set the default Windows AMI to Server 2022

colinrymer:
AWS removed the Windows Server 20H2 AMIs on [2022-08-09](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-windows-ami-version-history.html), causing this
module to fail when trying to lookup AMIs with that name.

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -15,7 +15,7 @@ locals {
   kms_key_arn           = var.kms_key_arn != null ? var.kms_key_arn : ""
 
   default_ami = {
-    "windows" = { name = ["Windows_Server-2019-English-Core-ContainersLatest-*"] }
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
Should this not be 2022? It's probably better to use the newer image with longer support and more features.

colinrymer:
@dylanmtaylor I didn't see the 2022 images by AWS in the AMI marketplace, but the 2021.09.01 update on https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-windows-ami-version-history.html does list a 2022 version of the English core containers build, so setting this to 2022 seems totally reasonable to me.

ScottGuymer:
I thin this should be 2022 as it is the ltsc version

ScottGuymer (approved):

PR #2345, opened 2022-08-16
fix(runners): Pass allocation strategy

rasmus:
Seems that the allocation strategy isn't passed.

*Note: First time doing anything in type script*


npalm (approved):
Good catch, thanks!

PR #2346, opened 2022-08-17
fix(runners): Fetch instance environment tag though metadata

rasmus:
Instead of relying on a rate limit not being hit getting the tags, fetch the EC2 environment tag directly from the metadata layer. Its rate limit is unique for the instance.


npalm:
Sorry for the late response, due to PTO I was offline. Will check the PR the next days.

npalm:
Checked the PR and work as exported

npalm:
AWS is a bit vague about throttling. But I agree seems not the same bucket as the API.

Query throttling
We throttle queries to the instance metadata service on a per-instance basis, and we place limits on the number of simultaneous connections from an instance to the instance metadata service. See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html

npalm (approved):

npalm:
Note to @ScottGuymer @GuptaNavdeep1983  we should also make an update in our inhouse user_data scripts.

PR #2355, opened 2022-08-20
chore: Use `https` instead of `git` in pre-commit config

kmaehashi:
This fixes the problem that `pre-commit` cannot be newly configured because `git://` protocol has been shutdown in GitHub on March 15, 2022.

https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git


npalm (approved):
@kmaehashi thanks, good catch. LGTM

PR #2369, opened 2022-08-24
feat: Allow passing instance type when building windows image

kmaehashi:
Support `instance_type` variable in Windows runner image, similar to other runner images.


npalm (approved):
LGTM, thanks

npalm:
Sorry for the late response, was offline due to PTO

PR #2371, opened 2022-08-29
feat: add s3_location_runner_distribution var as expandable for userdata

sgtoj:
Add the `s3_location_runner_distribution` variable as expandable variable for the `userdata_template`. Although the native module will not need this variable to be expand for `userdata_template`, it is much easier for users that use custom `userdata_template`.

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -6,13 +6,14 @@ locals {
     var.tags,
   )
 
-  name_sg               = var.overrides["name_sg"] == "" ? local.tags["Name"] : var.overrides["name_sg"]
-  name_runner           = var.overrides["name_runner"] == "" ? local.tags["Name"] : var.overrides["name_runner"]
-  role_path             = var.role_path == null ? "/${var.prefix}/" : var.role_path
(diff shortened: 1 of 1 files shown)


npalm:
Sorry for the delay, due to PTO was not able to catch up with PR's. Will check the PR this week. Can you provide a usages example for custom AMI. In our custom AMI we pre-bake the runner binary, so not download it via S3 at all.

npalm (approved):
LGTM

ScottGuymer (approved):

sgtoj:
Our custom user-data script will detect the number of instance storage devices (e.g. m5ad.large, m5d.large, m6id.large, etc). If it detects two or more, configure them in a `RAID0` configuration, then format and mount it. If there is one instance storage device, it only formats and mount

Once the instance storage is ready, it installs configures docker and github-action-runner to the new (faster) mount. It downloads the _updated_ runner from the module's s3 distro bucket. Hence the reason for this PR. 

---

We achieve the above by passing our custom user-data (and not using any of the script provided by the module) script.

npalm:
Thx, looks good. Maybe we should document which variables can be used in a custom user_data and will be replaced.

PR #2472, opened 2022-09-27
feat: Add multi-runner capability

GuptaNavdeep1983:
This PR introduces the option to create a multi runner setup with one GitHup app and a single webhook. Currently the only way to run multiple runner configurations is deploy the module multiple times with the burden of managing different apps.

This PR introduces a submodule `multi-runner` to allow to deploy a configuration for having multiple runners. The webhook will filter events based on labels via a matcher and delegate the events to dedicated queues for each runner configurations.

We have chosen not to refactor the module from the bottom but build on what we have today. This results in that we just deploy for each runner configuration the runner module ones. And per os/arch configuration the distribution syncer (if enabled).

See also https://github.com/philips-labs/terraform-aws-github-runner/discussions/1428

## Todo

- [x] Implement multi runner module
- [x] Add example
- [x] Add documentation to the main README as well of the module
- [x] Desribe migration impact  (below)

## Migration/Upgrade

- Does the main module needs to be redeployed after this feature is released?
  There is no change in terraform version for the main module, so it will not require TF upgrade. There is only one change in main module, which is a breaking change, the variable `runner_enable_workflow_job_labels_check` is not available under this module as we now only support the flag `runner_enable_workflow_job_labels_check_all` to check all/partial workflow labels to be present in the runner labels. Running TF apply on existing deployment will apply some changes done in internal modules to existing infrastructure.
- Describe impact on using internal (submodules).
  There are some changes in the `webhook` internal module as per this feature. So, If you have been using it directly, the changes will be towards accepting runner configuration (queue details and label matchers) as a single object instead of earlier configuration of scattered queue details and criteria to match the workflow labels against the runner labels.


modules/multi-runner/queues.tf
```diff
@@ -0,0 +1,76 @@
...
+      values   = ["false"]
+    }
+  }
+}
+
+
+resource "aws_sqs_queue" "queued_builds" {
+  count                       = length(var.multi_runner_config)
+  name                        = "${var.prefix}-${var.multi_runner_config[count.index]["runner_config"]["id"]}-queued-builds${var.multi_runner_config[count.index]["fifo"] ? ".fifo" : ""}"
+  delay_seconds               = var.delay_webhook_event
```
npalm, line 37: I think the delay should also move to the runner config. For ephemeral I would set it to zero in other cases from 5 to 30.

modules/multi-runner/queues.tf
```diff
@@ -0,0 +1,76 @@
...
+  visibility_timeout_seconds  = var.runners_scale_up_lambda_timeout
```
npalm, line 38: Not sure yet. But the visibility timeout should match the lambda timeout.  I don't see a reason yet to differ per runner. I think could be the same for all configs.

modules/multi-runner/queues.tf
```diff
@@ -0,0 +1,76 @@
+
+locals {
+  multi_runner_queues_config = tolist([for index, queue in var.multi_runner_config : merge(aws_sqs_queue.queued_builds[index], queue)])
+  unique_os_and_arch         = distinct([for index, config in local.multi_runner_queues_config : { "os_type" : config["runner_config"]["runner_os"], "architecture" : config["runner_config"]["runner_architecture"] } if config["runner_config"]["enable_runner_binaries_syncer"]])
```
npalm, line 4: Is it needed to check `enabled_runner_binaries_syncer`? The local name unique_os_and_arch suggest that you try list all supported os and arch bij the config.

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+    "ghr:environment" = "${var.prefix}-${local.multi_runner_queues_config[count.index]["runner_config"]["id"]}"
+  })
+
+  s3_runner_binaries = local.multi_runner_queues_config[count.index]["runner_config"]["enable_runner_binaries_syncer"] ? local.runner_binaries_by_os_and_arch_map["${local.multi_runner_queues_config[count.index]["runner_config"]["runner_os"]}_${local.multi_runner_queues_config[count.index]["runner_config"]["runner_architecture"]}"] : {}
+
+  runner_os                     = local.multi_runner_queues_config[count.index]["runner_config"]["runner_os"]
+  instance_types                = local.multi_runner_queues_config[count.index]["runner_config"]["instance_types"]
+  instance_target_capacity_type = var.instance_target_capacity_type
+  instance_allocation_strategy  = var.instance_allocation_strategy
+  instance_max_spot_price       = var.instance_max_spot_price
```
npalm, line 19: I think it make sense to move those properties to the runner config, this will support the option to run a spot-fleet and on-demand-fleet. Max spot price will differ based on instance type.

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+  enable_job_queued_check              = local.multi_runner_queues_config[count.index]["runner_config"]["enable_job_queued_check"]
+  disable_runner_autoupdate            = local.multi_runner_queues_config[count.index]["runner_config"]["disable_runner_autoupdate"]
+  enable_managed_runner_security_group = var.enable_managed_runner_security_group
+  enable_runner_detailed_monitoring    = var.enable_runner_detailed_monitoring
+  scale_down_schedule_expression       = local.multi_runner_queues_config[count.index]["runner_config"]["scale_down_schedule_expression"]
+  minimum_running_time_in_minutes      = local.multi_runner_queues_config[count.index]["runner_config"]["minimum_running_time_in_minutes"]
+  runner_boot_time_in_minutes          = local.multi_runner_queues_config[count.index]["runner_config"]["runner_boot_time_in_minutes"]
+  runner_extra_labels                  = local.multi_runner_queues_config[count.index]["runner_config"]["runner_extra_labels"]
+  runner_as_root                       = local.multi_runner_queues_config[count.index]["runner_config"]["runner_as_root"]
+  runner_run_as                        = var.runner_run_as
```
npalm, line 39: I would move this also to the runner config

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+  runners_maximum_count                = local.multi_runner_queues_config[count.index]["runner_config"]["runners_maximum_count"]
+  idle_config                          = var.idle_config
```
npalm, line 41: Would also move to the runner config, what do you think?

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+  lambda_timeout_scale_up          = var.runners_scale_up_lambda_timeout
+  lambda_timeout_scale_down        = var.runners_scale_down_lambda_timeout
+  lambda_subnet_ids                = var.lambda_subnet_ids
+  lambda_security_group_ids        = var.lambda_security_group_ids
+  logging_retention_in_days        = var.logging_retention_in_days
+  logging_kms_key_id               = var.logging_kms_key_id
+  enable_cloudwatch_agent          = var.enable_cloudwatch_agent
+  cloudwatch_config                = var.cloudwatch_config
+  runner_log_files                 = var.runner_log_files
+  runner_group_name                = local.multi_runner_queues_config[count.index]["runner_config"]["runner_group_name"]
```
npalm, line 63: would multi_runner_config not be a better name?

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  scale_up_reserved_concurrent_executions = var.scale_up_reserved_concurrent_executions
```
npalm, line 65: This is to control the concurrent lambda execution. Assume you have a config supporting a big fleet and small fleet. I think you also would like to set this differently. Therefore I think we should move it to the config as wel.

modules/multi-runner/runners.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  instance_profile_path     = var.instance_profile_path
+  role_path                 = var.role_path
+  role_permissions_boundary = var.role_permissions_boundary
+
+  enabled_userdata      = var.enabled_userdata
```
npalm, line 71: userdata / cloudinit is baded on the runner AMI, so would move this to the config.

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -124,48 +111,34 @@ async function verifySignature(
...
-      id: body.workflow_job.id,
-      repositoryName: body.repository.name,
-      repositoryOwner: body.repository.owner.login,
-      eventType: githubEvent,
-      installationId: installationId,
-    });
-    logger.info(`Successfully queued job for ${body.repository.full_name}`, LogFields.print());
+  for (const queue of queuesConfig) {
+    if (canRunJob(body.workflow_job.labels, queue.labelMatchers, queue.exactMatch)) {
+      if (body.action === 'queued') {
```
npalm, line 119: Would it not more efficient to check this before the for loop?

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -124,48 +111,34 @@ async function verifySignature(
...
+        await sendActionRequest({
+          id: body.workflow_job.id,
+          repositoryName: body.repository.name,
+          repositoryOwner: body.repository.owner.login,
+          eventType: githubEvent,
+          installationId: installationId,
+          queueId: queue.id,
+          queueFifo: queue.fifo,
+        });
+        logger.info(`Successfully queued job for ${body.repository.full_name}`, LogFields.print());
```
npalm, line 129: We should update this log line, so we know to which queue the event was submitted, and which config was applied.

npalm (changes requested):
Partial first review. Great to see we finally can create runners for different configurations. I did a first partial review.

- Can you also update the main README and add a section to explain the mulit runner setup.
- Can you add a readme to the multi. runner example.

modules/webhook/lambdas/webhook/src/webhook/handler.ts
GuptaNavdeep1983: moved.

modules/webhook/lambdas/webhook/src/webhook/handler.ts
GuptaNavdeep1983: updated to add queue id.

modules/multi-runner/runners.tf
GuptaNavdeep1983: updated.

modules/multi-runner/runners.tf
GuptaNavdeep1983: i think you already updated this.

modules/multi-runner/runners.tf
GuptaNavdeep1983: updated.

modules/multi-runner/runners.tf
GuptaNavdeep1983: you updated.

modules/multi-runner/runners.tf
GuptaNavdeep1983: updated.

modules/multi-runner/queues.tf
GuptaNavdeep1983: updated.

examples/multi-runner/README.md
```diff
@@ -0,0 +1,39 @@
+# Action runners deployment of Multiple-Runner-Configurations-Together example
+
+This module shows how to create GitHub action runners with multiple runner configuration together in one deployment.
```
npalm, line 3: I think it could be helpful to explain how the filters are working, that the order is important, and if relevant the exact match. Would also mention that a single App is used for all runners.

examples/multi-runner/main.tf
```diff
@@ -0,0 +1,88 @@
...
+      "labelMatchers" : ["self-hosted", "linux", "x64", "staging"]
+      "exactMatch" : true,
+      "runner_config" : {
+        "id" : "linux-x64",
+        "runner_os" : "linux",
+        "runner_architecture" : "x64",
+        "create_service_linked_role_spot" : true,
+        "enable_ssm_on_runners" : true,                              # enable access to the runners via SSM
+        "instance_types" : ["m5ad.large", "m5a.large", "c5.xlarge"], # c5.xlarge is backup, we select based on price
+        "runner_extra_labels" : "staging"
```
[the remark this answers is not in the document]
GuptaNavdeep1983: Do we need to change something here?

modules/multi-runner/queues.tf
GuptaNavdeep1983: I think it can be outside runner configuration.

examples/multi-runner/README.md
GuptaNavdeep1983: Updated the readme.

modules/multi-runner/queues.tf
npalm: Agree, we koop them on top level.

modules/multi-runner/queues.tf
npalm: agree,

examples/multi-runner/main.tf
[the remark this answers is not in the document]
npalm: I will have a look again

modules/multi-runner/variables.tf
```diff
@@ -0,0 +1,493 @@
...
+}
+
+variable "runner_extra_labels" {
+  description = "Extra (custom) labels for the runners (GitHub). Separate each label by a comma. Labels checks on the webhook can be enforced by setting `enable_workflow_job_labels_check`. GitHub read-only labels should not be provided."
+  type        = string
+  default     = ""
+}
+
+variable "multi_runner_config" {
+  type = map(object({
```
npalm, line 35: Here we missing a description. We should point out that in general all the variables are based on the one in the underlying modules and documente there. But explain the one we introduce here like matcher config

npalm:
Checked an upgrade for the default example, went smooth without an issue

modules/multi-runner/variables.tf
GuptaNavdeep1983: Added the description already in this block, is there anything else required?

examples/multi-runner/main.tf
```diff
@@ -0,0 +1,165 @@
+locals {
+  environment = var.environment != null ? var.environment : "multi-runner"
+  aws_region  = "eu-central-1"
```
npalm, line 3:
```suggestion
  aws_region  = "eu-west-1"
```

examples/multi-runner/versions.tf
```diff
@@ -0,0 +1,15 @@
...
+      version = "~> 4.0"
+    }
+    local = {
+      source = "hashicorp/local"
+    }
+    random = {
+      source = "hashicorp/random"
+    }
+  }
+  required_version = ">= 1"
```
npalm, line 14:
```suggestion
  required_version = ">= 1.3"
```

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -3,16 +3,15 @@ import { CheckRunEvent, WorkflowJobEvent } from '@octokit/webhooks-types';
...
 
 const supportedEvents = ['workflow_job'];
 const logger = rootLogger.getChildLogger();
 
 export async function handle(headers: IncomingHttpHeaders, body: string): Promise<Response> {
-  const { environment, repositoryWhiteList, enableWorkflowLabelCheck, workflowLabelCheckAll, runnerLabels } =
-    readEnvironmentVariables();
+  const { environment, repositoryWhiteList, queuesConfig } = readEnvironmentVariables();
 
   // ensure header keys lower case since github headers can contain capitals.
```
npalm, line 16:
```suggestion
  # ensure header keys lower case since github headers can contain capitals.
```

Issue #2417, opened 2022-09-13
Resource Groups are no longer working 


jake-naughton:
It looks like after a change went out to update the `Environment` tag to `ghr:environment` that the Resource Group was not updated to match this.

When I navigate to the Resource Group in AWS I cannot see any related infrastructure since the tag query it is using is:
Environment: prefix

If I update this to:
ghr:environment: prefix

It fixes the issue.

jake-naughton:
It should be a simple fix, just need to update the resource-group.json from 

```
{
  "ResourceTypeFilters": ["AWS::AllSupported"],
  "TagFilters": [
    {
      "Key": "Environment",
      "Values": ["${environment}"]
    }
  ]
}
```

to

```
{
  "ResourceTypeFilters": ["AWS::AllSupported"],
  "TagFilters": [
    {
      "Key": "ghr:environment",
      "Values": ["${environment}"]
    }
  ]
}
```

Happy to put up a PR.

npalm:
Good catch, please feel free to create a PR

npalm:
We are not setting a predefined tag to all resources. Which means we cannot create a tag based resource group vai the module. Via the provider you can set a tag to all resources. This means we have to remove the resource group from the module at let it to the user to define a resource group in the deployment.

jake-naughton:
@npalm I have added a PR (#2512) to fix this if you could please review when you get a chance.

npalm:
I have created a PR #2514 to create resource groups for the examples

PR #2512, opened 2022-10-10
fix: Remove resource group from module

jake-naughton:
Fix for bug #2417 to remove resource group from module since we are no longer setting a predefined tag to all resources.

Tests all passed locally.

Pre commit checks updated terraform docs to include missing documentation from a recently merged pull request for S3. Happy to remove these commits and leave them for a separate pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,7 +380,6 @@ In case the setup does not work as intended follow the trace of events:
 
 | Name | Type |
 |------|------|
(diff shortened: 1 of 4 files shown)


npalm (approved):
Thanks looks good, will run a quick test later

npalm:
Thx for you help, all good!

PR #2514, opened 2022-10-11
chore: Add resource groups for examples

npalm:
Adding resource groups for exampls, and refactory examples.

- Add resource groups for examples
- Add default Example tag
- Moved VPC to generic base module


examples/base/main.tf
```diff
@@ -0,0 +1,8 @@
+# resource "aws_resourcegroups_group" "resourcegroups_group" {
```
GuptaNavdeep1983, line 1: do we want this code to be commented?

examples/default/main.tf
```diff
@@ -1,23 +1,25 @@
 locals {
-  environment = "default"
+  environment = "npalm"
```
GuptaNavdeep1983, line 2: do you want this to be named after you?

GuptaNavdeep1983 (approved):

PR #2520, opened 2022-10-13
Add support for looking up pre-built runner AMI ID from an SSM parameter at instance launch time

jpalomaki:
👋 

Use case: we build custom runner AMIs periodically (as new actions/runner versions are released), and we have disabled auto-updates on the runners (to keep things more predictable and fast). Now the problem is that currently we need to re-apply the runner terraform config, to pick up a newly-built AMI (since it is looked up by the terraform data source).

This PR introduces an option to have the scale up lambda look up the runner AMI ID from an SSM parameter at _runner instance launch time_, thereby allowing us to update the AMI, without having to re-apply runner stacks.

To make this work fully automated, we'd also amend our AMI build workflow to update this AMI ID SSM parameter, as a new AMI version is built. Then it would be automatically picked up by the runner stack.

This PR is a draft for now, because I assume we'd need to add some tests and amend documentation.

I have quickly smoke-tested the code in an actual GitHub repo/AWS account.

Let me know what you think.

Thanks,
Jukka

### DIFF
--- a/modules/runners/lambdas/runners/src/aws/runners.ts
+++ b/modules/runners/lambdas/runners/src/aws/runners.ts
@@ -42,6 +42,7 @@ export interface RunnerInputParameters {
     instanceAllocationStrategy: EC2.SpotAllocationStrategy;
   };
   numberOfRunners?: number;
+  amiIdSsmParam?: string;
 }
 
 export async function listEC2Runners(filters: ListRunnerFilters | undefined = undefined): Promise<RunnerList[]> {
@@ -107,17 +108,27 @@ export async function terminateRunner(instanceId: string): Promise<void> {
   logger.info(`Runner ${instanceId} has been terminated.`, LogFields.print());
 }
 
-function generateFleeOverrides(
+function generateFleetOverrides(
   subnetIds: string[],
   instancesTypes: string[],
+  amiId?: string,
 ): EC2.FleetLaunchTemplateOverridesListRequest {
+  type Spec = {
+    SubnetId: string;
+    InstanceType: string;
+    ImageId?: string;
+  };
   const result: EC2.FleetLaunchTemplateOverridesListRequest = [];
   subnetIds.forEach((s) => {
     instancesTypes.forEach((i) => {
-      result.push({
+      const item: Spec = {
         SubnetId: s,
         InstanceType: i,
-      });
+      };
+      if (amiId) {
+        item.ImageId = amiId;
+      }
+      result.push(item);
     });
   });
   return result;
@@ -127,6 +138,21 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
   logger.debug('Runner configuration: ' + JSON.stringify(runnerParameters), LogFields.print());
 
   const ec2 = new EC2();
+  const ssm = new SSM();
+
+  let amiIdFromSsm = undefined;
+
+  if (runnerParameters.amiIdSsmParam) {
+    logger.debug(`Looking up runner AMI ID from SSM parameter: ${runnerParameters.amiIdSsmParam}`);
+    try {
+      const result: AWS.SSM.GetParameterResult = await ssm.getParameter({Name: runnerParameters.amiIdSsmParam}).promise();
+      amiIdFromSsm = result.Parameter?.Value;
+    } catch (e) {
+      logger.error(`Failed to lookup runner AMI ID from SSM parameter: ${runnerParameters.amiIdSsmParam}`, e);
+      throw e;
+    }
+  }
+
   const numberOfRunners = runnerParameters.numberOfRunners ? runnerParameters.numberOfRunners : 1;
 
   let fleet: AWS.EC2.CreateFleetResult;
@@ -140,9 +166,10 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
               LaunchTemplateName: runnerParameters.launchTemplateName,
               Version: '$Default',
             },
-            Overrides: generateFleeOverrides(
+            Overrides: generateFleetOverrides(
               runnerParameters.subnets,
               runnerParameters.ec2instanceCriteria.instanceTypes,
+              amiIdFromSsm,
             ),
           },
         ],
@@ -202,7 +229,6 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
 
   logger.info('Created instance(s): ', instances.join(','), LogFields.print());
 
-  const ssm = new SSM();
   for (const instance of instances) {
     await ssm
       .putParameter({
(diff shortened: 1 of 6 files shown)


GuptaNavdeep1983:
@jpalomaki the PR looks nice, can you add some more tests around this PR to meet the coverage goals, its failing the build.

variables.tf
```diff
@@ -302,11 +302,19 @@ variable "ami_filter" {
...
   default     = null
 }
+
 variable "ami_owners" {
   description = "The list of owners used to select the AMI of action runner instances."
   type        = list(string)
   default     = ["amazon"]
 }
+
+variable "ami_id_ssm_param" {
```
npalm, line 312: I thgink it could be confusing how to use this. I would suggest the folloing

- change the name to `ami_id_ssm_paramateter_name`
- Make in the descripton clear the het parameter is not managed by the module, but externally.
- Add a short section to the global readme, maybe in the section [usage](https://github.com/philips-labs/terraform-aws-github-runner#usages)

npalm (changes requested):
Looks nice, need to test and added some comments. Can you also add the default example with a commented out part to show how the AMI override works?

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
```diff
@@ -32,6 +32,7 @@ interface CreateEC2RunnerConfig {
   launchTemplateName: string;
   ec2instanceCriteria: RunnerInputParameters['ec2instanceCriteria'];
   numberOfRunners?: number;
+  amiIdSsmParam?: string;
```
GuptaNavdeep1983, line 35: can you add or update existing test case to check if the createRunner function is called with given ami id from the environment variable?

npalm:
@jpalomaki did you had time to check our comments?

jpalomaki:
> @jpalomaki did you had time to check our comments?

@npalm Not yet (I was afk this week). I will have a look at them on Monday. 👍

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -127,6 +138,23 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
...
+
+  if (runnerParameters.amiIdSsmParam) {
+    logger.debug(`Looking up runner AMI ID from SSM parameter: ${runnerParameters.amiIdSsmParam}`);
+    try {
+      const result: AWS.SSM.GetParameterResult = await ssm
+        .getParameter({ Name: runnerParameters.amiIdSsmParam })
+        .promise();
+      amiIdFromSsm = result.Parameter?.Value;
+    } catch (e) {
+      logger.error(`Failed to lookup runner AMI ID from SSM parameter: ${runnerParameters.amiIdSsmParam}`, e);
```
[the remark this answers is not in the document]
jpalomaki: Fixed in b2d3bc5dfc640661296f91fa2f65a271581ea843

modules/runners/lambdas/runners/src/aws/runners.test.ts
```diff
@@ -370,6 +410,7 @@ function createRunnerConfig(runnerConfig: RunnerConfig): RunnerInputParameters {
       instanceAllocationStrategy: runnerConfig.allocationStrategy,
     },
     subnets: ['subnet-123', 'subnet-456'],
+    amiIdSsmParam: runnerConfig.amiIdSsmParam,
```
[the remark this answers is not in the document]
jpalomaki: Fixed

modules/runners/scale-up.tf
```diff
@@ -36,6 +36,7 @@ resource "aws_lambda_function" "scale_up" {
       RUNNER_GROUP_NAME                    = var.runner_group_name
       RUNNERS_MAXIMUM_COUNT                = var.runners_maximum_count
       SUBNET_IDS                           = join(",", var.subnet_ids)
+      AMI_ID_SSM_PARAM                     = var.ami_id_ssm_param
```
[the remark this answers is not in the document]
jpalomaki: Fixed

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -127,6 +138,23 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
   logger.debug('Runner configuration: ' + JSON.stringify(runnerParameters), LogFields.print());
 
   const ec2 = new EC2();
+  const ssm = new SSM();
+
+  let amiIdFromSsm = undefined;
```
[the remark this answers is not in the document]
jpalomaki: Fixed

variables.tf
jpalomaki: Fixed

modules/runners/lambdas/runners/src/scale-runners/scale-up.ts
jpalomaki: Fixed

jpalomaki:
@npalm @GuptaNavdeep1983 Comments addressed. Let me know what you think

jpalomaki:
@npalm 

> Can you also add the default example with a commented out part to show how the AMI override works?

This is now also done in https://github.com/philips-labs/terraform-aws-github-runner/pull/2520/commits/8fe47d656111386d847d06bfc6e09eeb2e87e4e5

npalm (approved):
Thanks, looks good. Test the PR on top of the default example with the code below:

```hcl

module "runners" {
  ...
  ami_id_ssm_parameter_name = aws_ssm_parameter.runner_enable_cloudwatch.name
}

data "aws_ami" "runner" {
  most_recent = "true"

  filter {
      name   = "name"
      values = ["amzn2-ami-kernel-5.*-hvm-2.0.202209*-x86_64-gp*"] # some older AMI
  }

  owners = ["amazon"]
}

resource "aws_ssm_parameter" "runner_enable_cloudwatch" {
  name  = "/${local.environment}/runners/ami"
  type  = "String"
  value =  data.aws_ami.runner.image_id
}
```

modules/runners/scale-up.tf
```diff
@@ -110,3 +111,25 @@ resource "aws_iam_role_policy_attachment" "scale_up_vpc_execution_role" {
...
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Action": [
+            "ssm:GetParameter"
+          ],
+          "Resource": [
+            "arn:aws:ssm:${var.aws_region}:${data.aws_caller_identity.current.account_id}:parameter/${trimprefix(var.ami_id_ssm_parameter_name, "/")}"
```
jokreliable, line 129: the partition might need to be parameterized here, so that it can run in the various govcloud implementations
(I believe it's already part of some of the options passed in to various modules, maybe not this one)

PR #2521, opened 2022-10-14
feat: Remove support check_run

npalm:
The event `check_run` was used in the module for scaling before GitHub introduced the event `workflow_job`. The workflow job event was introduced in release [0.17.0](https://github.com/philips-labs/terraform-aws-github-runner/compare/v0.16.0...v0.17.0) (2021-08-06).

## Migration
Updatte the webhook / app to subscribe to workflow_job instead of check_run. In case using GHES and not having support for the workflow_job yu have to upgrade to a newer version (3.3.0+ / Dec 2021).

- [ ] verify with example

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,15 +46,14 @@ A logical question would be, why not Kubernetes? In the current approach, we sta
 
 ## Overview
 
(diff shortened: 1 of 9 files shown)


ScottGuymer (approved):

GuptaNavdeep1983 (approved):

npalm:
> looks good.
> 
> Do we want to leave something in the notes or changelog about deprecating it and the migration path>

Therore I gathering all changes in an issue. And write / update the changelog during th erelease / merg back of the next branch

Issue #2449, opened 2022-09-22
Ephemeral autoscaling does not work, idle runners always at maximum


arthur-telia:
We are trying to setup autoscaling using Ephemeral runners flag, It seems that the lambda function never scales runners down and always keeps the runners at the value of `runners_maximum_count` that we have set.

for example, if we set `runners_maximum_count` to `20`, then exactly `20` will be idle all the time, no matter if we have any pending jobs or not

version is `v1.8.1`, 
using prebuilt AWS AMI `ubuntu-jammy-22.04-amd64-server`

related configuration below
```
  # Uncomment to enable ephemeral runners
  delay_webhook_event      = 0
  enable_ephemeral_runners = true
  enabled_userdata         = false
  minimum_running_time_in_minutes = 10
  runners_maximum_count = 3
  scale_down_schedule_expression = "cron(* * * * ? *)"
  enable_job_queued_check = true
  idle_config = [{
    cron      = "* * 9-17 * * 1-5"
    timeZone  = "Europe/Amsterdam"
    idleCount = 3
  }]
```
could you please advice if this is a bug or misconfiguration from our side?
let me know if more information is required

arthur-telia:
@npalm I've tried the pool block as well, but it always stays at the `runners_maximum_count`
```
  pool_runner_owner = "company"                  # Org to which the runners are added
  pool_config = [{
    size                = 5                    # size of the pool
    schedule_expression = "cron(* * * * ? *)"   # cron expression to trigger the adjustment of the pool
  }]
```
how do I scale the runners when I have no pending jobs in the queue?

npalm:
Creating a pool will create every time the function is triggered based on the schedule runners. It first looks up the number of idle runners, and next top up in your case to 5. The scale down lambda can shutdown you pool runners. Based on an minimal running time.

npalm:
[M1kep](https://github.com/M1kep) Would be great if you have time to submit a PR. 

I thought the issue is the pool lambda is not seeing instances that are created but not ready yet. So assuming you running the pool lambda every minute, and creating an instance takes 2 minutes you can easy end up with 2 or 3 times the size of the pool

piscue:
@M1kep @npalm faced the same issue. I had a monitor that controls the amount of EC2 spawn and never honored the `runners_maximum_count`. I did check the code, I was about to open the issue, but I found this one.

[...]

PR #2590, opened 2022-10-28
fix: Honor runner boot time when topping up pools

M1kep:
Relates to #2449 

This change updates the pool lambda so that instances aren't ignored while they boot

### DIFF
--- a/.github/workflows/lambda-runner-binaries-syncer.yml
+++ b/.github/workflows/lambda-runner-binaries-syncer.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - develop
+      - next
   pull_request:
     paths:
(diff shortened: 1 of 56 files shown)


M1kep:
@npalm This seems to be working for us so far. There's still some odd scaling behavior that we haven't pinned down, but this specific change has fixed the primary cause of over-scaling that we've seen.

Not sure if you'd like me to update the conflicts or remove the draft status, but if so let me know

npalm:
Thanks for the update, will try to check the PR tomorrow.

npalm:
Seems something went wrong with rebasing. Please can you chck? I assume the intention is merging back into develop. Seems you rebased to next.

PR #2592, opened 2022-10-29
feat: Allow to toggle access to EC2 instance tags

MGSousa:
Added variable option for toggle (`enabled`/`disabled`) access to EC2 instance tags through metadata service.


npalm (commented):
@MGSousa looks good, but curious in what case do you disable the metadaservice_tags? The instance start-up requires on tags. Or do you use an alternative user_data script to start your runners in such way you are not requiring the tags.

MGSousa:
> @MGSousa looks good, but curious in what case do you disable the metadaservice_tags? The instance start-up requires on tags. Or do you use an alternative user_data script to start your runners in such way you are not requiring the tags.

For this case, I need to disable this metadata service since this restricts "spacing" in tags (https://github.com/aws/aws-cli/issues/6914), and for now, it's the only solution.
Also, I am using a custom user_data.sh script for starting the runners that call the `describe-tags` for each instance.

variables.tf
```diff
@@ -571,6 +571,7 @@ variable "runner_metadata_options" {
   description = "Metadata options for the ec2 runner instances."
```
npalm, line 571:
```suggestion
  description = "Metadata options for the ec2 runner instances. By default, the module uses metadata tags for bootstrapping the runner, only disable `instance_metadata_tags` when using custom scripts for starting the runner."
```

modules/runners/variables.tf
```diff
@@ -499,6 +499,7 @@ variable "metadata_options" {
   description = "Metadata options for the ec2 runner instances."
```
npalm, line 499:
```suggestion
  description = "Metadata options for the ec2 runner instances. By default, the module uses metadata tags for bootstrapping the runner, only disable `instance_metadata_tags` when using custom scripts for starting the runner."
```

npalm (changes requested):
Looks good, need to run a quick deployment test. Added a small suggestion

npalm (approved):

PR #2664, opened 2022-11-15
refactor: use optional in variable block_device_mappings

jpalomaki:
Should make usage a little more straightforward when one only wants to override some attributes, e.g volume size

### DIFF
--- a/modules/runners/variables.tf
+++ b/modules/runners/variables.tf
@@ -58,26 +58,18 @@ variable "s3_runner_binaries" {
 variable "block_device_mappings" {
   description = "The EC2 instance block device configuration. Takes the following keys: `device_name`, `delete_on_termination`, `volume_type`, `volume_size`, `encrypted`, `iops`, `throughput`, `kms_key_id`, `snapshot_id`."
   type = list(object({
-    delete_on_termination = bool
-    device_name           = string
-    encrypted             = bool
-    iops                  = number
-    kms_key_id            = string
-    snapshot_id           = string
-    throughput            = number
+    delete_on_termination = optional(bool, true)
+    device_name           = optional(string, "/dev/xvda")
+    encrypted             = optional(bool, true)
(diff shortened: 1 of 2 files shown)


modules/runners/variables.tf
```diff
@@ -58,26 +58,18 @@ variable "s3_runner_binaries" {
...
-    snapshot_id           = string
-    throughput            = number
+    delete_on_termination = optional(bool, true)
+    device_name           = optional(string, "/dev/xvda")
+    encrypted             = optional(bool, true)
+    iops                  = optional(number)
+    kms_key_id            = optional(string)
+    snapshot_id           = optional(string)
+    throughput            = optional(number)
     volume_size           = number
```
npalm, line 68: Why not setting the volume size as optional as well?

npalm (changes requested):

modules/runners/variables.tf
jpalomaki: Good question. I was not sure about this one, but in the end I reasoned about it like so: volume size should be a conscious choice, if one chooses to override EBS settings. The default value for this _variable_ still provides a default value of 30GiB (to retain backwards-compatibility), so you'd only need to set `volume_size` if you override EBS settings, and size is IMO the most probable thing for users to be adjusting.

jpalomaki: That said, I will happily adjust this, if you think it's more important to align the usage of optional here (provide all defaults in the same way).

npalm: Like your reasing, agree. let's go forward with your proposal

npalm:
Please can you rebase the PR, or maybe easier re-created. I rebased earlier today next with develop. (sorry) / thanks any ways!

jpalomaki:
@npalm PR recreated. Untested atm though.

npalm (approved):

PR #2688, opened 2022-11-25
chore(images): Add ability to specify iam instance role

mattgodbolt:


npalm (approved):

npalm (changes requested):
Sorry for the late reply, please can you format the packer code?

mattgodbolt:
> Sorry for the late reply, please can you format the packer code?

Done! Sorry for _my_ late reply :)

images/ubuntu-focal/github_agent.ubuntu.pkr.hcl
```diff
@@ -78,6 +78,12 @@ variable "custom_shell_commands" {
   default     = []
 }
 
+variable "instance_profile" {
+  description = "IAM instance profile for the builder to run as"
+  type        = string
+  default     = ""
```
npalm, line 84: Maybe better to set the default to null

PR #2704, opened 2022-11-30
fix: SSM prefix

npalm:
SSM prefix not applied for path. 

- Use prefix for all SSM paramaters
- Moved prefix to path instead of name for mult-runner
- Add optionals for ssm path variable


modules/multi-runner/variables.tf
```diff
@@ -515,13 +515,9 @@ variable "workflow_job_queue_configuration" {
...
+    root    = optional(string, "github-action-runners")
+    app     = optional(string, "app")
+    runners = optional(string, "runners")
   })
-  default = {
-    root    = "github-action-runners"
-    runners = "runners"
-    app     = "app"
-  }
+  default = {}
```
GuptaNavdeep1983, line 522: I am not sure what is the use of this?

variables.tf
```diff
@@ -756,15 +756,10 @@ variable "enable_user_data_debug_logging_runner" {
...
+    runners    = optional(string, "runners")
+    use_prefix = optional(bool, true)
   })
-  default = {
-    root       = "github-action-runners"
-    use_prefix = true
-    runners    = "runners"
-    app        = "app"
-  }
+  default = {}
```
GuptaNavdeep1983, line 764: Same as above.

modules/multi-runner/variables.tf
npalm: Is not used, leaving the default let it fail.

npalm: With this example you can try it your self,

```ncl
variable "test" {
  type = object({
    name = optional(string, "jonh")
    age  = optional(number, 20)
  })
  default = {}
}

output "test" {
  value = var.test
}
```

By setting the default to `{}` the object get a required value, the optionals ensures if object values are not set, they set with the default.

GuptaNavdeep1983 (approved):

PR #2733, opened 2022-12-07
fix: Honor runner boot time when topping up pools

mckern:
This patch is rebased off of the patch provided by @M1kep in philips-labs/terraform-aws-github-runner#2590; that PR has gone stale and kept it stuck in draft status. I've re-targeted the patch against the develop branch, and I'm willing to do some work to get this mergable if more work is required. This should take care of #2449 and supersede #2590.

### DIFF
--- a/modules/runners/lambdas/runners/src/pool/pool.ts
+++ b/modules/runners/lambdas/runners/src/pool/pool.ts
@@ -1,9 +1,10 @@
 import yn from 'yn';
 
-import { listEC2Runners } from '../aws/runners';
+import { listEC2Runners, RunnerList } from '../aws/runners';
 import { createGithubAppAuth, createGithubInstallationAuth, createOctoClient } from '../gh-auth/gh-auth';
 import { logger as rootLogger } from '../logger';
 import { createRunners } from '../scale-runners/scale-up';
+import moment from "moment";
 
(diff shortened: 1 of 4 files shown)


modules/runners/lambdas/runners/src/pool/pool.ts
```diff
@@ -99,3 +104,9 @@ async function getInstallationId(ghesApiUrl: string, org: string): Promise<numbe
     })
   ).data.id;
 }
+
+function bootTimeExceeded(ec2Runner: RunnerList): boolean {
```
npalm, line 108: Exact the same function is defined in `scale-down.ts`. Can you move the function to either a common file same dir. Or to the runners.ts and use the function via an import. This avoid duplication.

npalm (changes requested):
@mckern thanks for taking the initiative to got the PR merged! Made a small remark. Would it be possible to check if you als can add a unit test ot check the pool scaling works properly?

mckern:
@npalm 
> Would it be possible to check if you als can add a unit test ot check the pool scaling works properly?

Good question; let's find out 😂.

npalm (commented):
@mckern I have created and fixed the runit test via PR
https://github.com/philips-labs/terraform-aws-github-runner/pull/2801

Issue #2732, opened 2022-12-06
Issue with 1.17.0 when upgrading


jkritzen:
If we upgrade to 1.17.0 we get the following error:

**Error**: error modifying Lambda Function (atc-cgbp-syncer) configuration : InvalidParameterValueException: The provided execution role does not have permissions to call CreateNetworkInterface on EC2

Is there anything missing?

jkritzen:
Is seems a issue within the module, we do not use custom roles in the deployment.

The full error message:

module.github-runner.module.runner_binaries[0].aws_lambda_function.syncer: Still modifying... [id=atc-cgbp-syncer, 4m50s elapsed]
╷
│Error: error modifying Lambda Function (atc-cgbp-syncer) configuration : InvalidParameterValueException: The provided execution role does not have permissions to call CreateNetworkInterface on EC2
│{
│  RespMetadata: {
│    StatusCode: 400,
│    RequestID: "44d753f7-a90e-4ce4-8eed-3c93a4406fa0"
│  },
│  Message_: "The provided execution role does not have permissions to call CreateNetworkInterface on EC2",
│  Type: "User"
│}
│
│  with module.github-runner.module.runner_binaries[0].aws_lambda_function.syncer,
│  on .terraform/modules/github-runner/modules/runner-binaries-syncer/runner-binaries-syncer.tf line 10, in resource "aws_lambda_function" "syncer":
│  10: resource "aws_lambda_function" "syncer" {
│
╵

npalm:
Frow which version to which version are you upgrading?

GuptaNavdeep1983:
@jkritzen , the error is quite confusing. Based on the error message, its not clear why lambda function would try to create ec2 network interface but regardless, can you check if these permissions (ec2:CreateNetworkInterface) are missing to the role which is used to execute this terraform code?

GuptaNavdeep1983:
@jkritzen, this issue has been addressed as per latest release v1.18.0, can you please check and revert.

GuptaNavdeep1983:
solved, feel free to re-open if any need

jkritzen:
@GuptaNavdeep1983 : Thanks for fixing, with v.1.18.0 it works as expected:
![image](https://user-images.githubusercontent.com/43033899/207058783-bcd594ac-c243-468a-81b2-5862f7fd066b.png)

PR #2734, opened 2022-12-08
fix: added permissions for lambda to attach lambda to the VPC.

GuptaNavdeep1983:
#2732 
Updated the code to add permissions for lambda required in case the lambda needs to be attached to the VPC. The issue was reported only for runner_syncer lambda but this issue will be seen for all lambda functions which needs to be attached to VPC.


modules/runner-binaries-syncer/runner-binaries-syncer.tf
```diff
@@ -100,6 +100,13 @@ resource "aws_iam_role_policy" "lambda_logging" {
   })
 }
 
+resource "aws_iam_role_policy" "lambda_syncer_vpc" {
+  name = "${var.prefix}-lambda-syncer-vpc"
```
npalm, line 104: Can you add a conditional and only activate the policy if the lambda is attached to the vpc aka `lambda_subnet_ids` has more than zero elements.

npalm (changes requested):
@GuptaNavdeep1983 thanks looks good, one small suggestion

modules/runner-binaries-syncer/runner-binaries-syncer.tf
GuptaNavdeep1983: Addressed.

modules/runner-binaries-syncer/runner-binaries-syncer.tf
```diff
@@ -100,6 +100,14 @@ resource "aws_iam_role_policy" "lambda_logging" {
   })
 }
 
+resource "aws_iam_role_policy" "lambda_syncer_vpc" {
+  count = var.lambda_subnet_ids != null && var.lambda_security_group_ids != null ? 1 : 0
```
npalm, line 104: This won't work would, default for both `[]`. So would check for a non empty list.

GuptaNavdeep1983: I have tested it and it worked just fine!

npalm: I testing the default example and it tries to create the policies, which should not be created



``` 
  # module.runners.module.runners.aws_iam_role_policy.lambda_scale_up_vpc[0] will be created
  + resource "aws_iam_role_policy" "lambda_scale_up_vpc" {
      + id     = (known after apply)
      + name   = "npalm-lambda-scale-up-vpc"
      + policy = jsonencode(
            {
              + Statement = [
                  + {
                      + Action   = [
                          + "ec2:CreateNetworkInterface",
                          + "ec2:DescribeNetworkInterfaces",
                          + "ec2:DeleteNetworkInterface",
                        ]
                      + Effect   = "Allow"
                      + Resource = "*"
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
      + role   = "npalm-action-scale-up-lambda-role"
    }
```

npalm (approved):
Thanks @navdeepg2021 looks good

npalm:
close #2732

Issue #2715, opened 2022-12-02
[multi-runner] - Add support for multiple label arrays to matcher


npalm:
Currently a multi runner is implement, see next branch (#2517 ) The configuration supports matching based on a single array of labels.

```
      matcherConfig : {
        labelMatchers = ["self-hosted", "linux", "x64", "ubuntu-latest"]
        exactMatch    = true
      }
```

But it will give a bit more flexicility to also allow to match on multiple arrays of runner labels like:

```
      matcherConfig : {
        labelMatchers = [["self-hosted", "linux", "x64", "ubuntu-latest"], ["self-hosted", "linux", "x64", "ubuntu-2204"]]
        exactMatch    = true
      }
```

mariusfilipowski:
Looks like this feature is already available, right?

npalm:
yup, so closing. Thanks @mariusfilipowski

PR #2736, opened 2022-12-08
feat(webhook): Support multiple arrays of tags is matchers.

npalm:
Refactored the matcher to support multiple arrays of tags. This allow to match multiple set of tags on a single runner configuration. Useful to match for example `[self-hosted, ubuntu-latest]` and `[self-hosted, ubuntu-2204]` on the same runner.


npalm:
close #2715

GuptaNavdeep1983:
@npalm, not sure if this PR gives an error on applying without commit https://github.com/philips-labs/terraform-aws-github-runner/pull/2736/commits/2ab5f465796c66bff1ccddb83471d62b79c1ae27#diff-69a0e0ca008de526fb8fe2c1d1995d6632c16a9cd1bc58c9f58eec9ce1676f30!

npalm:
Tested the branch with the example for multi-runners, all works fine.

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
```diff
@@ -418,6 +418,48 @@ describe('handler', () => {
     });
   });
 
+  it('Check webhook will accept jobs when matchers accepts multiple labels.', async () => {
+    process.env.RUNNER_CONFIG = JSON.stringify([
+      {
+        ...queuesConfig[0],
+        matcherConfig: {
+          labelMatchers: [['self-hosted', 'arm64', 'linux', 'ubuntu-latest']],
```
GuptaNavdeep1983, line 426: should we have multiple labels supported by this runner to let the test be more effective?

modules/webhook/lambdas/webhook/src/webhook/handler.ts
```diff
@@ -161,18 +161,22 @@ function isRepoNotAllowed(repoFullName: string, repositoryWhiteList: string[]):
...
+  runnerLabelsMatchers: string[][],
+  workflowLabelCheckAll: boolean,
+): boolean {
+  runnerLabelsMatchers = runnerLabelsMatchers.map((runnerLabel) => {
+    return runnerLabel.map((label) => label.toLowerCase());
   });
   const match = workflowLabelCheckAll
-    ? workflowJobLabels.every((l) => runnerLabels.includes(l.toLowerCase()))
-    : workflowJobLabels.some((l) => runnerLabels.includes(l.toLowerCase()));
+    ? workflowJobLabels.every((wl) => runnerLabelsMatchers.some((rl) => rl.includes(wl.toLowerCase())))
```
GuptaNavdeep1983, line 173: Quite complicated query.

modules/webhook/lambdas/webhook/src/webhook/handler.test.ts
npalm: Good catch

modules/webhook/lambdas/webhook/src/webhook/handler.ts
npalm: I think we need this, to be able to support cases like ubuntu-latest and ubuntu-2204. Alternative is deploying another stack, wich is even worse in my point of view

GuptaNavdeep1983 (approved):

PR #2784, opened 2022-12-21
refactor(runners): rename variable enabled_userdata -> enable_userdata

jpalomaki:
For consistency with other `enable_*` variables


npalm:
thx, all good

PR #2801, opened 2022-12-28
fix: Honnor booting instance in runner pool

npalm:
@mckern @M1kep Thanks for the work done in #2590 and #2733 

When digging in today for review I found out the test are not running as expected. Bu fixing th test I decided also to take the changes you proposed in your PR into account. I would appreciate when you have a bit of time to check this PR.

### DIFF
--- a/modules/runners/lambdas/runners/src/pool/pool.test.ts
+++ b/modules/runners/lambdas/runners/src/pool/pool.test.ts
@@ -149,7 +149,11 @@ describe('Test simple pool.', () => {
     it('Top up pool with pool size 2 registered.', async () => {
       await expect(await adjust({ poolSize: 3 })).resolves;
       expect(createRunners).toHaveBeenCalledTimes(1);
-      expect(createRunners).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({numberOfRunners: 2}), expect.anything());
+      expect(createRunners).toHaveBeenCalledWith(
+        expect.anything(),
+        expect.objectContaining({ numberOfRunners: 2 }),
+        expect.anything(),
+      );
     });
 
     it('Should not top up if pool size is reached.', async () => {
@@ -180,7 +184,11 @@ describe('Test simple pool.', () => {
 
       // 1 idle + 1 booting = 2, top up with 3 to match a pool of 5
       await expect(await adjust({ poolSize: 5 })).resolves;
-      expect(createRunners).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({numberOfRunners: 3}), expect.anything());
+      expect(createRunners).toHaveBeenCalledWith(
+        expect.anything(),
+        expect.objectContaining({ numberOfRunners: 3 }),
+        expect.anything(),
+      );
     });
 
     it('Should not top up if pool size is reached including a booting instance.', async () => {
@@ -217,7 +225,11 @@ describe('Test simple pool.', () => {
     it('Top up if the pool size is set to 5', async () => {
       await expect(await adjust({ poolSize: 5 })).resolves;
       // 1 idle, top up with 4 to match a pool of 5
-      expect(createRunners).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({numberOfRunners: 4}), expect.anything());
+      expect(createRunners).toHaveBeenCalledWith(
+        expect.anything(),
+        expect.objectContaining({ numberOfRunners: 4 }),
+        expect.anything(),
+      );
     });
   });
 });
--- a/modules/runners/lambdas/runners/src/pool/pool.test.ts
+++ b/modules/runners/lambdas/runners/src/pool/pool.test.ts
@@ -1,10 +1,11 @@
 import { Octokit } from '@octokit/rest';
 import { mocked } from 'jest-mock';
+import moment from 'moment-timezone';
 import nock from 'nock';
 
 import { listEC2Runners } from '../aws/runners';
 import * as ghAuth from '../gh-auth/gh-auth';
-import * as scale from '../scale-runners/scale-up';
+import { createRunners } from '../scale-runners/scale-up';
 import { adjust } from './pool';
 
 const mockOctokit = {
@@ -24,6 +25,7 @@ jest.mock('@octokit/rest', () => ({
 
 jest.mock('./../aws/runners');
 jest.mock('./../gh-auth/gh-auth');
+jest.mock('./../scale-runners/scale-up');
 
 const mocktokit = Octokit as jest.MockedClass<typeof Octokit>;
 const mockedAppAuth = mocked(ghAuth.createGithubAppAuth, {
@@ -36,6 +38,28 @@ const mockListRunners = mocked(listEC2Runners);
 const cleanEnv = process.env;
 
 const ORG = 'my-org';
+const MINIMUM_TIME_RUNNING = 15;
+
+const ec2InstancesRegistered = [
+  {
+    instanceId: 'i-1',
+    launchTime: new Date(),
+    type: 'Org',
+    owner: ORG,
+  },
+  {
+    instanceId: 'i-2',
+    launchTime: new Date(),
+    type: 'Org',
+    owner: ORG,
+  },
+  {
+    instanceId: 'i-3',
+    launchTime: new Date(),
+    type: 'Org',
+    owner: ORG,
+  },
+];
 
 beforeEach(() => {
   nock.disableNetConnect();
@@ -55,6 +79,7 @@ beforeEach(() => {
   process.env.INSTANCE_TYPES = 'm5.large';
   process.env.INSTANCE_TARGET_CAPACITY_TYPE = 'spot';
   process.env.RUNNER_OWNER = ORG;
+  process.env.RUNNER_BOOT_TIME_IN_MINUTES = MINIMUM_TIME_RUNNING.toString();
 
   const mockTokenReturnValue = {
     data: {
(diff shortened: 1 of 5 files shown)


npalm:
@mckern @M1kep please can you have a look on the PR. I added tests and refactory the implementaiton so both registered instances as still booting instances are considered.

mckern:
Oh, thank you! I have been working this on the clock (and therefore away from my work computer for the last week and a half); I didn't mean to leave it fallow as long as I did but the end of the year is always kind of a liminal time at work. I'll take a look when I'm on the clock tomorrow (Wednesday) -- this was much appreciated.

npalm:
@mckern @M1kep please can you have a look on the PR. I added tests and refactory the implementaiton so both registered instances as still booting instances are considered.

> Oh, thank you! I have been working this on the clock (and therefore away from my work computer for the last week and a half); I didn't mean to leave it fallow as long as I did but the end of the year is always kind of a liminal time at work. I'll take a look when I'm on the clock tomorrow (Wednesday) -- this was much appreciated.

No problem, had a bit of time and thought let's dig in. I hope this helps. I will leave the PR open. Hopefully you have a bit of time early next year to check the implementation.

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -87,44 +112,9 @@ beforeEach(() => {
       busy: false,
       labels: [],
     },
-    {
-      id: 11,
-      name: 'j-1', // some runner of another env
```
npalm, line 92: Removed the non env runners, since the the function listEc2Runners should only return valid runners for the environment

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -155,16 +145,56 @@ beforeEach(() => {
 
 describe('Test simple pool.', () => {
   describe('With GitHub Cloud', () => {
-    it('Top up pool with pool size 2.', async () => {
-      const spy = jest.spyOn(scale, 'createRunners');
-      await expect(adjust({ poolSize: 2 })).resolves;
-      expect(spy).toBeCalled;
```
npalm, line 161: Seems this have never worked, for some reason the spy is not picking up. Rpleaces the spy by a mock.

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -155,16 +145,56 @@ beforeEach(() => {
...
 
     it('Should not top up if pool size is reached.', async () => {
-      const spy = jest.spyOn(scale, 'createRunners');
-      await expect(adjust({ poolSize: 1 })).resolves;
-      expect(spy).not.toHaveBeenCalled;
+      await expect(await adjust({ poolSize: 1 })).resolves;
+      expect(createRunners).not.toHaveBeenCalled();
+    });
+
+    it('Should top up if pool size is not reached including a booting instance.', async () => {
```
npalm, line 158: Added tests to ensure booted and orphan instances are honnored.

modules/runners/lambdas/runners/src/pool/pool.ts
```diff
@@ -44,20 +50,41 @@ export async function adjust(event: PoolEvent): Promise<void> {
       per_page: 100,
     },
   );
-  const idleRunners = runners.filter((r) => !r.busy && r.status === 'online').map((r) => r.name);
+  const runnerStatus = new Map<string, RunnerStatus>();
+  for (const runner of runners) {
```
npalm, line 54: Build a map to quickly look up a runner status

modules/runners/lambdas/runners/src/pool/pool.ts
```diff
@@ -99,3 +126,9 @@ async function getInstallationId(ghesApiUrl: string, org: string): Promise<numbe
     })
   ).data.id;
 }
+
+function bootTimeExceeded(ec2Runner: RunnerList): boolean {
```
npalm, line 130: This function is almost identical to the on for scale-down. But for some reason when centralizing them tests keep failing on scale down.

mckern (dismissed):

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -175,9 +213,8 @@ describe('Test simple pool.', () => {
     });
 
     it('Top up if the pool size is set to 5', async () => {
-      const spy = jest.spyOn(scale, 'createRunners');
-      await expect(adjust({ poolSize: 5 })).resolves;
-      expect(spy).toBeCalled;
+      await expect(await adjust({ poolSize: 3 })).resolves;
+      expect(createRunners).toHaveBeenCalledTimes(1);
```
M1kep, line 217: The check name doesn't match the poolSize

modules/runners/lambdas/runners/src/pool/pool.ts
M1kep: Both of the below implementations work locally for me, not too sure where the definition should live though.
One thing I do wonder about, is in what scenarios launchTime would be undefined.

(Both are identical except for the parameter ec2Runner parameter type)
```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: Pick<RunnerList, "launchTime"> | Pick<RunnerInfo, "launchTime">): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: {launchTime?: Date}): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -156,16 +146,64 @@ beforeEach(() => {
 
 describe('Test simple pool.', () => {
   describe('With GitHub Cloud', () => {
-    it('Top up pool with pool size 2.', async () => {
-      const spy = jest.spyOn(scale, 'createRunners');
-      await expect(adjust({ poolSize: 2 })).resolves;
-      expect(spy).toBeCalled;
+    it('Top up pool with pool size 2 registered.', async () => {
+      await expect(await adjust({ poolSize: 3 })).resolves;
+      expect(createRunners).toHaveBeenCalledTimes(1);
```
M1kep, line 151: Thoughts on checking the numberOfRunners is correct via the below. The same check could also be applied to the other tests, with a different value

```typescript
expect(createRunners).toHaveBeenCalledWith(expect.anything(), expect.objectContaining({numberOfRunners: 2}), expect.anything());
```

modules/runners/lambdas/runners/src/pool/pool.test.ts
```diff
@@ -36,6 +38,28 @@ const mockListRunners = mocked(listEC2Runners);
 const cleanEnv = process.env;
 
 const ORG = 'my-org';
+const MINIMUM_TIME_RUNNING = 15;
+
+const ec2InstancesRegistered = [
+  {
+    instanceId: 'i-1',
+    launchTime: new Date(),
```
M1kep, line 46: I'm wondering if there are, or could be, any implications with the tests only running with instances that have start times within the MINIMUM_TIME_RUNNING window.

modules/runners/lambdas/runners/src/pool/pool.ts
M1kep: Both of the below implementations work locally for me, not too sure where the definition should live though.
One thing I do wonder about, is in what scenarios launchTime would be undefined.

(Both are identical except for the parameter ec2Runner parameter type)
```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: Pick<RunnerList, "launchTime"> | Pick<RunnerInfo, "launchTime">): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: {launchTime?: Date}): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

npalm:
@M1kep thanks for the comments, will check asap

modules/runners/lambdas/runners/src/pool/pool.ts
npalm: For some reason the scale-down tests remain faling. Will create an issue to solve this later.

npalm: See #2878

ScottGuymer (approved):

PR #2802, opened 2022-12-28
fix: Main module broken after supporting multiple labels 

npalm:
PR #2736 broken the default module by not repecting the change label type


npalm:
checked all good, thx

PR #2846, opened 2023-01-03
chore: Prefix production updates with fix to trigger releases

npalm:
Set different commit prefix for development and production updates
- fix(component) for production
- chore(component) for dev dependencies

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,10 @@ updates:
     schedule:
       interval: "weekly"
 
+
+  # NPM production dependencies are part of the generated Lambda JavaScript.
+  # Therefore updates on production are prefixed with fix(component) to trigger releases.
+  # Development updates are prefixed with chore, and not triggering a release.
   - package-ecosystem: "npm"
     directory: "/modules/runner-binaries-syncer/lambdas/runner-binaries-syncer"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -16,13 +16,19 @@ updates:
...
     schedule:
       interval: "weekly"
+      prefix: "fix(syncer)"
+      prefix-development: "chore(syncer)"
 
   - package-ecosystem: "npm"
     directory: "/modules/webhook/lambdas/webhook"
     schedule:
       interval: "weekly"
+      prefix: "fix(webhook)"
```
ScottGuymer, line 26: Should a bump to versions not be a `chore` in both cases?

npalm: No changed this to bump production changes with fix. Only dev with chor. This seems the only way to trigger new releases for security updates as well.

npalm: `build` would be better maybe but it does not trigger a release. And the artifact is changed, since the derived code from the dependency is part of the bundled js file

GuptaNavdeep1983 (approved):

.github/dependabot.yml
npalm: Added comment to the yaml

ScottGuymer (approved):

PR #2858, opened 2023-01-09
feat: Add runner logfiles to output

sanketrathi:
This PR adds logfile information for runners to the module output

With this PR, we add propagate `runners.logfiles` out through the module
output. This would potentially enable module users to create CloudWatch
queries / metric filters (eg: runner startup time)


modules/multi-runner/outputs.tf
```diff
@@ -10,6 +10,7 @@ output "runners" {
     role_scale_up           = runner.role_scale_up
     role_scale_down         = runner.role_scale_down
     role_pool               = runner.role_pool
+    logfiles                = module.runners.logfiles
```
npalm, line 13: This won't work, you need to refer to the runner object.
```suggestion
    logfiles                = runner.logfiles
```

npalm (changes requested):
Thanks for your contribution. Please can yo shortly describe the need for the output, would like to understand the problem you solving. In general ok, one small remark.

modules/multi-runner/outputs.tf
sanketrathi: Thanks for the review @npalm . The PR got closed with an accidental force push, sincere apologies

sanketrathi: I've also updated the PR description with the motivating use case(s) I'm trying to solve with this change

sanketrathi: Another quick question: I was also planning on modifying the `multi-runner` module to allow `ami_id_ssm_parameter_name` as a parameter with `runner_config`. Should I raise a separate PR for it?

npalm: Yes please create a separate PR

modules/multi-runner/outputs.tf
```diff
@@ -10,6 +10,7 @@ output "runners" {
     role_scale_up           = runner.role_scale_up
     role_scale_down         = runner.role_scale_down
     role_pool               = runner.role_pool
+    logfiles                = runners.logfiles
```
npalm, line 13:
```suggestion
    logfiles                = runner.logfiles
```

npalm (changes requested):

npalm (approved):

npalm (approved):

PR #2880, opened 2023-01-12
refactor(runners): Move bootTimeExceeded to runners module

### LINKED ISSUE #2878: remove duplicate function from runners lambda's
Introduced via #2801 both [`pool.ts`](../blob/main/modules/runners/lambdas/runners/src/pool/pool.ts) adn [`scale-down.ts`](../blob/main/modules/runners/lambdas/runners/src/scale-runners/scale-down.ts) contains the same function `bootTimeExceeded`. Moving the function to a central place, fore exampler `runners.ts` let the unit tests for `scale-down` fail. 

Suggestion by @M1kep
Both of the below implementations work locally for me, not too sure where the definition should live though.
One thing I do wonder about, is in what scenarios launchTime would be undefined.

(Both are identical except for the parameter ec2Runner parameter type)
```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: Pick<RunnerList, "launchTime"> | Pick<RunnerInfo, "launchTime">): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

```typescript
// If launchTime is undefined, this will return false
export function bootTimeExceeded(ec2Runner: {launchTime?: Date}): boolean {
  const runnerBootTimeInMinutes = process.env.RUNNER_BOOT_TIME_IN_MINUTES;
  const launchTimePlusBootTime = moment(ec2Runner.launchTime).utc().add(runnerBootTimeInMinutes, 'minutes');
  return launchTimePlusBootTime < moment(new Date()).utc();
}
```

M1kep:
(If preferred, this can be re-targeted to main once #2801 is merged)
It appears that Jest was replacing the `bootTimeExceeded`, possibly with a `jest.Fn()`, due to the `jest.mock('./../aws/runners')`, causing it to return undefined when called under test. The introduced changes should keep the underlying module function implementations, while mocking the functions that are needed.

Ref: https://stackoverflow.com/a/53402206

Closes #2878

### DIFF
--- a/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
+++ b/modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
@@ -27,7 +27,11 @@ jest.mock('@octokit/rest', () => ({
   Octokit: jest.fn().mockImplementation(() => mockOctokit),
 }));
 
-jest.mock('./../aws/runners');
+jest.mock('./../aws/runners', () => ({
+  ...jest.requireActual('./../aws/runners'),
+  terminateRunner: jest.fn(),
+  listEC2Runners: jest.fn(),
+}));
 jest.mock('./../gh-auth/gh-auth');
 jest.mock('./cache');
 
--- a/modules/runners/lambdas/runners/src/aws/runners.ts
+++ b/modules/runners/lambdas/runners/src/aws/runners.ts
@@ -1,4 +1,5 @@
 import { EC2, SSM } from 'aws-sdk';
+import moment from 'moment';
 
 import { LogFields, logger as rootLogger } from '../logger';
 import ScaleError from './../scale-runners/ScaleError';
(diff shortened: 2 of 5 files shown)


npalm:
recovered the branch

npalm:
@M1kep there are some formatting issues in the CI, can you run a `yarn format`. Can you update the PR as well with a rebase on main and change the PR branch to main. Sorry I missed that you was planning to merge in the fix branch.

M1kep:
@npalm Should be all set now

modules/runners/lambdas/runners/src/scale-runners/scale-down.test.ts
```diff
@@ -27,7 +27,11 @@ jest.mock('@octokit/rest', () => ({
   Octokit: jest.fn().mockImplementation(() => mockOctokit),
 }));
 
-jest.mock('./../aws/runners');
+jest.mock('./../aws/runners', () => ({
```
npalm, line 30: this is what I missed! Big thanks!

npalm (approved):

npalm:
Strange, seems the CI is not triggerd on the runner.

npalm:
Checkedn locally. All good. The Ci seems not to be triggered after change the base branch in the PR.

Issue #2861, opened 2023-01-09
Event match runner labels even though they're just a subset


ahamez:
When using the principal module (so, not the multi-runners one), I added the label `on-demand`:

```HCL
module "on_demand" {
  source  = "philips-labs/github-runner/aws"
  version = "v2.1.0"
  ...
  runner_extra_labels = "on-demand"
}
```

The Github workflow DOES NOT specify this label:
```yaml
...
jobs:
  build:
    runs-on: [self-hosted, Linux, X64]
...
```

However, even though the workflow doesn't specify the `on-demand` label, a runner is started. When setting the log level to `debug`, we can see that the event does match the runner labels:

```
Received workflow job event with labels: '["self-hosted","Linux","X64"]'.
The event does match the runner labels: 'self-hosted,linux,x64,on-demand'
```

Is it an intended behaviour? 

If so, how can I restrict only some workflows to run the runners deployed by this module?
For instance, a use-case would be to run only some workflows on instances with a GPU.

If it matters, I use the org level setup. Also, I didn't change the value of `enable_runner_workflow_job_labels_check_all`, which defaults to `true`.

npalm:
What you described in the issue should work, what means the job should not be accepted, And you are correct, the check label is by default true. The single runner is using internally the same matcher as the multi-runner. I have added a [test](https://github.com/philips-labs/terraform-aws-github-runner/pull/2881) to check the Issue. But the test is running fine. 

It could be still possible that there is a bug in the terraform config, since this is not checked in the unit test.

ahamez:
If it helps, here's the full terraform definition:

```
# ----------------------------------------------------------------------------------------------- #

resource "random_id" "webhook_secret" {
  byte_length = 20
}

# ----------------------------------------------------------------------------------------------- #

module "on_demand" {
  source  = "philips-labs/github-runner/aws"
  version = "v2.1.0"

  aws_region = "eu-west-3"
  vpc_id     = var.vpcs["it-1"].vpc_id
  subnet_ids = var.vpcs["it-1"].private_subnets

  prefix = "github-runner-on-demand"

  ami_filter = {
    name = ["github-runner-on-demand-nogpu*"]
  }
  ami_owners = ["11111111111111"]

  github_app = {
    key_base64     = var.github_actions_on_demand_runner_pem
    id             = "XXXXXX"
    webhook_secret = random_id.webhook_secret.hex
  }

  instance_types = ["m5ad.2xlarge"]

[...]

PR #2881, opened 2023-01-12
chore(webhook): Add test to check issue #2861

npalm:
The default module (single runner) is using internally the same matcher as multi-runner. But to be sure it all works added a unit test for a single matcher rule.


npalm:
@ScottGuymer please can you review this PR?

ScottGuymer (approved):

PR #2901, opened 2023-01-18
fix: update the order of some of the setup files

christophersavory:
Currently this project only works for AMIs that have an `ec2-user` that is setup.  This works fine for all the AMIs that are provided by AWS. However, companies that are using their own AMIs will usually have a separate user that is preferred and the `ec2-user` is often disabled due to security concerns.  

In the current files, `user-data.sh` will add `ec2-user` to do the docker group and it will also do a variable assignment
```
    user_name=ec2-user
```
After that, install-runner.sh and start-runner-sh are called where the `run_as` varable is setup.
```
    run_as=$(echo "$parameters" | jq -r '.[] | select(.Name == "'$ssm_config_path'/run_as") | .Value')
```

What needs to happen is the Setup should happen first (new script created called setup-runner.sh). And then the install/run scrips called.


GuptaNavdeep1983:
@christophersavory,Thanks for the PR, can you please add the description for this PR explaining the rationale for creating this PR?

npalm:
@christophersavory please can you add a description to the PR. It also looks like you fix the user and refactoring the structure. I would suggest we split thoses changes in 2 PR's. First fix the user that is passed. Next restructure.

christophersavory:
@npalm sorry for the delay.  I updated the description as requested. 

I felt that the restructure needed to happen because 
```
    run_as=$(echo "$parameters" | jq -r '.[] | select(.Name == "'$ssm_config_path'/run_as") | .Value')
```
needs to be called early on (in the setup) to get the the proper user to do assignments. 

Is there another way you would suggest?

npalm:
Thansk, will check asap the PR

npalm:
@christophersavory
Still had no time to go in detail in the script changes, but can you re-base the branch please?

npalm:
@christophersavory The PR is still on my radar. But since the impact could be high I need a bit more time to dig in detail and run some deployment test. In case you have time can you have a look on the merge conflict.

christophersavory:
@npalm we can close this PR as I opened a new one. https://github.com/philips-labs/terraform-aws-github-runner/pull/3070

Issue #2874, opened 2023-01-10
DLQ is created although it is not configured


mariusfilipowski:
Using the multirunners, I noticed that we have now corresponding dead letter queues. I did use the following config taken from the multi runner example.

```
      redrive_build_queue = {
        enabled         = false
        maxReceiveCount = null
      }
```
Now I have 2 queue for each runner config. One with the suffix `-queued-builds` and another one with `-queued-builds_dead_letter`
Is this intended?

npalm:
Should not be intended. @GuptaNavdeep1983  ?

mariusfilipowski:
It looks like here a check for `var.redrive_build_queue.enabled` is missing:

https://github.com/philips-labs/terraform-aws-github-runner/blob/fca2ddfd545c2b670adcecf4e7d3107b9f4006b0/modules/multi-runner/queues.tf#L58

and here:

https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/multi-runner/queues.tf#L69

Similar to:
https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/multi-runner/queues.tf#L39

This check is also part of the "normal" runner module:

https://github.com/philips-labs/terraform-aws-github-runner/blob/main/main.tf#L106

npalm:
Would you have time to create a PR?

mariusfilipowski:
Currently not (sadly). Perhaps next week.

mariusfilipowski:
Please check: https://github.com/philips-labs/terraform-aws-github-runner/pull/2903

mariusfilipowski:
Hi! Just asking if the PR could be reviewed or is sth missing from my side?

GuptaNavdeep1983:
PR is already approved and merged.

PR #2903, opened 2023-01-19
fix(multi-runner): Create DLQ only if requested

mariusfilipowski:
Fix the problem that the setting `redrive_build_queue.enabled` was ignored.

Tested with the multirunner-example and terraform plan output.


mariusfilipowski:
Solves: https://github.com/philips-labs/terraform-aws-github-runner/issues/2874

npalm (approved):
Thx, sorry for the late review. Looks good. Tested with the default multi runner example

Issue #2883, opened 2023-01-13
Missing ami_id_ssm_parameter_name parameter from multi-runner


scatybeneva:
The optional parameter ami_id_ssm_parameter_name that can be used in a single runner is not defined with the multi-runner module so it has no effect. Would be very useful to override ami_filter in the same way with multi-runner.

npalm:
Here you can find an example how to overwrite the AMI including a custom user_data script

https://github.com/philips-labs/terraform-aws-github-runner/blob/a1013e91f0fe380ff7988e20e9efea78e73acea6/examples/multi-runner/main.tf#L45-L93

scatybeneva:
Yeah that works fine, but I'm talking about not using an ami_filter but to specify an SSM parameter store path to fetch the AMI ID built from an external process. This was added by [this](https://github.com/philips-labs/terraform-aws-github-runner/pull/2520).

Looking at the code, `ami_id_ssm_parameter_name` isn't set in the multi-runner module. So the scale-up lambda module won't set the corresponding env var either nor the policy to allow access to the SSM parameter.

PR #2911, opened 2023-01-24
fix: Missing ami_id_ssm_parameter_name parameter from multi-runner #2883

ido-goldshtein-orca:


npalm (approved):
@ido-goldshtein-orca thanks, looks all good. Sorry for the late reply

PR #2914, opened 2023-01-25
feat(2913): Add support for webhook lambda memory size

### LINKED ISSUE #2913: Unable to set memory size of lambdas
### Overview

When submitting GitHub workflow requests with large workflows and large user objects occasionally the webhook for processing the Github webhook can fail due to exhausting all memory.  

### Motivation

This issue is a core functionality blocker and prevents standard usage of the entire process due to the memory limitations of the lambda. 

### Reproduction 

Configure a large Github workflow and create a Github application that listens to the workflow on workflow run.  

### Related

I was unable to find any related issues.

devyncunningham:
This adds variables and defaults, with descriptions, for memory size attributions to the webhook lambda used to process GH Actions requests.  

Closes  #2913

### DIFF
--- a/modules/webhook/webhook.tf
+++ b/modules/webhook/webhook.tf
@@ -6,6 +6,7 @@ resource "aws_lambda_function" "webhook" {
   source_code_hash  = var.lambda_s3_bucket == null ? filebase64sha256(local.lambda_zip) : null
   function_name     = "${var.prefix}-webhook"
   role              = aws_iam_role.webhook_lambda.arn
+  memory            = var.lambda_memory_size
   handler           = "index.githubWebhook"
   runtime           = var.lambda_runtime
   timeout           = var.lambda_timeout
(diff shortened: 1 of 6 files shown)


modules/webhook/webhook.tf
```diff
@@ -6,6 +6,7 @@ resource "aws_lambda_function" "webhook" {
   source_code_hash  = var.lambda_s3_bucket == null ? filebase64sha256(local.lambda_zip) : null
   function_name     = "${var.prefix}-webhook"
   role              = aws_iam_role.webhook_lambda.arn
+  memory            = var.lambda_memory_size
```
npalm, line 9: The argument should be memory_size instead of memory.
```suggestion
  memory_size       = var.lambda_memory_size
```

npalm (changes requested):
Thanks for your contribution. The name of the attribute is wrong, please can you update your change baded on the suggestion. Just wondering why you adding more memory, does it speed up the lambda a lot?

npalm:
@devyncunningham did you had time to check my comments?

PR #2922, opened 2023-02-02
fix(runners): Fix path to SSM cloudwatch config parameter on Windows

kring:
`start-runner.ps1` was attempting to read the `cloudwatch_agent_config_runner` parameter from the wrong place, which of course failed. This PR fixes it by making it consistent with `start-runner.sh`.


npalm (approved):
@kring thanks!

Issue #2925, opened 2023-02-06
ScaleUp losing events when no capacity available


tcsc:
We are using a ephemeral `arm64` builders and are intermittently having builds stall due to a lack of capacity, even with `on-demand instances`. 

Every time this happens, the build permanently stalls and has to be manually cancelled.

From a bit of digging it looks like adding `InsufficientInstanceCapacity` to [the list of what's considered a "Scaling Error"](https://github.com/philips-labs/terraform-aws-github-runner/blob/3247f8d197910c4e7f0cd3cb6ab9d91544bcae58/modules/runners/lambdas/runners/src/aws/runners.ts#L238) should fix this. 

Redacted CloudWatch log: 
```
2023-02-06T10:27:56.604+11:00	2023-02-05 23:27:56.494 WARN [runners:34ecb39a-ae35-5506-b158-efc0938129a8 index.js:120365 createRunner] No instances created by fleet request. Check configuration! Response:
2023-02-06T10:27:56.604+11:00	{
2023-02-06T10:27:56.604+11:00	FleetId: 'fleet-92368284-5b0d-44bc-0e18-af80f19be5e5',
2023-02-06T10:27:56.604+11:00	Errors: [
2023-02-06T10:27:56.604+11:00	{
2023-02-06T10:27:56.604+11:00	LaunchTemplateAndOverrides: {
2023-02-06T10:27:56.604+11:00	LaunchTemplateSpecification: {
2023-02-06T10:27:56.604+11:00	LaunchTemplateId: 'lt-REDACTED',
2023-02-06T10:27:56.604+11:00	Version: '4'
2023-02-06T10:27:56.604+11:00	},
2023-02-06T10:27:56.604+11:00	Overrides: {
2023-02-06T10:27:56.604+11:00	InstanceType: 'c6gd.8xlarge',
2023-02-06T10:27:56.604+11:00	SubnetId: 'subnet-REDACTED'
2023-02-06T10:27:56.604+11:00	}
2023-02-06T10:27:56.604+11:00	},
2023-02-06T10:27:56.604+11:00	Lifecycle: 'on-demand',
2023-02-06T10:27:56.604+11:00	ErrorCode: 'InsufficientInstanceCapacity',
[...]

PR #2926, opened 2023-02-06
fix(runner): Adds InsufficientInstanceCapacity to list of scaling errors

tcsc:
The ScaleUp lambda is dropping events when there is no capacity in the zone. This results in jobs being stalled and requiring manual cancellation.

This patch adds the `InsufficientInstanceCapacity` error code to the list of scale-up errors, allowing the built-in retry mechanism to pick up the error and not lose the event

See-Also: philips-labs/terraform-aws-github-runner#2925


npalm (approved):

npalm:
fix #2925

PR #2938, opened 2023-02-09
fix: Allow lambda inside VPC

jsalcedo1987:
Hi! 👋

This PR contains the following changes:
-  It allows webhook lambda to run inside a VPC 
-  fix a minor issue in the syncer module, when the lambda is running inside vpc AWSLambdaVPCAccessExecutionRole needs to be attached.

Updated doc.

Do not hesitate to ask if you need further details,
Thanks!

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -306,6 +306,8 @@ module "runner_binaries" {
 
   lambda_subnet_ids         = var.lambda_subnet_ids
   lambda_security_group_ids = var.lambda_security_group_ids
+  aws_partition             = var.aws_partition
+  
 
   lambda_principals = var.lambda_principals
 }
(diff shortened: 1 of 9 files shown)


main.tf
```diff
@@ -161,6 +161,10 @@ module "webhook" {
   role_permissions_boundary = var.role_permissions_boundary
   repository_white_list     = var.repository_white_list
 
+  lambda_subnet_ids         = var.lambda_subnet_ids
```
npalm, line 164: Can you also update the file `modules/multi-runner/webhook.tf` and pass the new varabiles down to the webhook module.

npalm (changes requested):
@jsalcedo1987 Thanks for your contribution. A small remark. We basides the top level main module also the variables needs to be added to the mult-runner module.

main.tf
jsalcedo1987: Thanks for the heads up! 😅 I'm not using the new multi-runner module so I missed that bit. I've now added the missing variables to the file 👍

npalm (approved):
@jsalcedo1987 thanks looks all good now

PR #2986, opened 2023-02-22
feat: Add option to disable Runner Syncer Lambda trigger

anjosluc:
Currently the module allows us to use Lambda functions to sync GitHub runner binaries on S3 automatically to the latest release available on actions/runners repository, but sometimes a release can have issues in their functionality that were not caught by Github prior to release it, so we need to rollback the binary distribution on S3 to a previous version and disable Lambda Syncer to stop checking for updates. This option allows us to quickly solve problems in runners version like we had in https://github.com/actions/runner/issues/2435 by freezing automatic updates and still being able to use Runner Syncer.


npalm (changes requested):
@anjosluc looks good, can you add the new variable also to the multi-runner module? See https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/multi-runner/runner-binaries.tf

variables.tf
```diff
@@ -707,6 +707,12 @@ variable "enable_runner_binaries_syncer" {
   default     = true
 }
 
+variable "enable_event_rule_binaries_syncer" {
+  type        = bool
+  default     = true
+  description = "Option to disable EventBridge Lambda trigger, useful to stop automatic updates of binary distribution"
```
npalm, line 713:
```suggestion
  description = "Option to disable EventBridge Lambda trigger for the binary syncer, useful to stop automatic updates of binary distribution"
```

anjosluc: done

npalm (approved):

npalm:
@anjosluc thanks for your contribution!

PR #2997, opened 2023-02-26
chore(images): add ubuntu arm64 runner image

azarudeena:
This feature will add the option to have the Ubuntu ARM64 Packer image and it's adjustments needed for the template scripts.


npalm:
@azarudeena I have updated the install script, test the packer build. And test the module as is by running the multi architecture examples.

npalm (approved):

PR #2999, opened 2023-02-27
fix: enforce overwrite to ssm parameter

idog-orca:
fix : SSM parameter "cloudwatch_agent_config_runner" not updating (add overwrite to ssm parameter)

### DIFF
--- a/modules/runners/logging.tf
+++ b/modules/runners/logging.tf
@@ -47,7 +47,8 @@ resource "aws_ssm_parameter" "cloudwatch_agent_config_runner" {
   value = var.cloudwatch_config != null ? var.cloudwatch_config : templatefile("${path.module}/templates/cloudwatch_config.json", {
     logfiles = jsonencode(local.logfiles)
   })
-  tags = local.tags
+  overwrite = true
+  tags      = local.tags
 }
 
 resource "aws_cloudwatch_log_group" "gh_runners" {
--- a/modules/multi-runner/runners.tf
+++ b/modules/multi-runner/runners.tf
@@ -64,7 +64,7 @@ module "runners" {
   logging_retention_in_days        = var.logging_retention_in_days
   logging_kms_key_id               = var.logging_kms_key_id
   enable_cloudwatch_agent          = each.value.runner_config.enable_cloudwatch_agent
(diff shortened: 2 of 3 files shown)


modules/runners/logging.tf
```diff
@@ -47,7 +47,8 @@ resource "aws_ssm_parameter" "cloudwatch_agent_config_runner" {
   value = var.cloudwatch_config != null ? var.cloudwatch_config : templatefile("${path.module}/templates/cloudwatch_config.json", {
     logfiles = jsonencode(local.logfiles)
   })
-  tags = local.tags
+  overwrite = true
```
npalm, line 50: Why did you only update the overwrite here, and not on the other places?

idog-orca: This is the only parameter I encounter a problem with.

npalm: Can you give an example (scenario) that cause the issue please?

npalm: I just checks the docs, the overwrite force terraform to woverwrite existing parameters. This should mean the module will overwrite parameters not managed by the module. Or a left over from on incomplete destroy. Please can you provide some more detail, I think enable this by default could be harmfull.

idog-orca: The agent configuration requirements have changed since the model was deployed. 
Now the terraform ignores my changes to the configuration and doesn't update the SSM parameter.
The overwrite flag fixes the problem.

npalm: I tried to reproduce the issue, but did not succeed.

1. Deployed the module based on the example default
2. Checked the configuration in the parameter store, configuration based on the module defaults available.
3. Next added a custom configuration, e.g. `cloudwatch_config = templatefile("cloudwatch_config.json", {})` with changed template.
4. Ran terraformorm apply
5. Checked the config in the parameter store. Changed are successfully applied.

As per Terraform docs this should work as long the attribute is managed by Terraform. Only in case you attribute had become orphan Terraform won't overwrite it. Since the parameter store is potentially also used by others I think it is risky to overwrite by default.

Is the scenario I describe, correct? Can you reproduce the problem?

PR #3000, opened 2023-02-27
feat: Lambda for webhook authorization

NSpenlihauer:
# Security UseCase 
The purpose of this PR is to propose the ability to add a Labmda authorizer at webhook level.  
As _aws_apigatewayv2_route_ is built internally in the module, caller do not have the capacity to attach lambda directly from terraform deployment.
Lambda Authorizer can be leveraged to put in place CIDR based ACL.

# Additional information
The need of having a specific _aws_apigatewayv2_route_ (**webhook** and **webhook_with_authorizer**) came from the impossibility to gracefully manage the migration of an already deployed infrastructure **with webhook_authorizer** to a state **without webhook_authorizer**. Without these two resources, such migration will fail with this error :

`Error: deleting API Gateway v2 authorizer: ConflictException: Cannot delete authorizer 'multi-runner-github-action-webhook-authorizer', is referenced in route: POST /webhook
`

Thank you in advance for your review (and for this awesome repository) !


modules/webhook/main.tf
```diff
@@ -10,12 +10,51 @@ resource "aws_apigatewayv2_api" "webhook" {
...
+  authorization_type = "CUSTOM"
+  authorizer_id      = aws_apigatewayv2_authorizer.webhook_authorizer[0].id
+}
+
+data "aws_lambda_function" "authorization_function" {
+  count         = var.lambda_webhook_authorizer_id != null ? 1 : 0
+  function_name = var.lambda_webhook_authorizer_id
+}
+
+resource "aws_lambda_permission" "my_authorizer_lambda_permission" {
```
npalm, line 47: Can you update the name of rhe resource?

modules/webhook/main.tf
```diff
@@ -10,12 +10,51 @@ resource "aws_apigatewayv2_api" "webhook" {
...
+}
+
 resource "aws_apigatewayv2_route" "webhook" {
+  count     = var.lambda_webhook_authorizer_id != null ? 0 : 1
   api_id    = aws_apigatewayv2_api.webhook.id
   route_key = "POST /${local.webhook_endpoint}"
   target    = "integrations/${aws_apigatewayv2_integration.webhook.id}"
 }
 
+resource "aws_apigatewayv2_route" "webhook_with_authorizer" {
```
npalm, line 33: Would it not be possible to combine the two aws_apigatewayv2_route by setting the target based on a conditional. And set the euthorizer_id to null if not required?

npalm (changes requested):
@NSpenlihauer thank you for creating the PR. Can you also share the Lambda? Makes it a bit easier to test. Maybe we can add the lambda even to an example.

I made some small comments, but need a bit of more time to process the PR. Lets also see what we can do about making the migration as simple as possible.

Another idea could be we split even the webhook in 2 lamba's.
- One to Authorize (secret, CIDR). And give module users the option to overwrite the auzhorizer by an own lambda (including handling the secret.
- Second to handle the event.

I also looking at the moment at the EventBride. Splitting the lambda could also empower to allow other ways of directing events into the runner module. Asuming authorization is already done.

modules/webhook/main.tf
NSpenlihauer: Unfortunately no, that is what I tried to example in my comment.

We could have a single _aws_apigatewayv2_route_ ressource with ternary operators on _authorization_type_ and _authorizer_id_ (set to null if not required), but that may harden some migrations.
This will work fine until you try to migrate your infrastructure from a state in which it (_aws_apigatewayv2_route_) have an authorizer attached to a state in which you remove it. Here is the problematic workflow : 
- Deploy your infrastructure with lambda_authorization (_lambda_webhook_authorizer_id_ set while calling the module) 
- Update your infrastructure to remove it (remove _lambda_webhook_authorizer_id_ from the call to module)
  - Terraform apply fail with this error : `Error: deleting API Gateway v2 authorizer: ConflictException: Cannot delete authorizer 'multi-runner-github-action-webhook-authorizer', is referenced in route: POST /webhook`

I guess it's some kind of deadlock in which tf is trying to get rid of the authorizer while it's still attached to the route.

I find it more convinient to "duplicate" the ressource rather than forcing users to manually detach the authorizer before applying the modification. 

Having a single _aws_apigatewayv2_route_ resource was actually my first attempt, but I was not able to make it work gracefully.

modules/webhook/main.tf
NSpenlihauer: Sure ! I update it to _authorizer_lambda_permission_

NSpenlihauer:
Thank you @npalm for your review ! 

> Can you also share the Lambda? Makes it a bit easier to test. Maybe we can add the lambda even to an example.

Sure thing, it's still manually handled in our infra, but here is the code behind it : 

```
exports.handler = async(event) => {
    let response = {
        "isAuthorized": false
    };
    const ip4ToInt = ip =>
      ip.split('.').reduce((int, oct) => (int << 8) + parseInt(oct, 10), 0) >>> 0;
    
    const isIp4InCidr = ip => cidr => {
      const [range, bits = 32] = cidr.split('/');
      const mask = ~(2 ** (32 - bits) - 1);
      return (ip4ToInt(ip) & mask) === (ip4ToInt(range) & mask);
    };
    
    const isIp4InCidrs = (ip, cidrs) => cidrs.some(isIp4InCidr(ip));
    let listCidrGh = [
        "192.30.252.0/22",
        "185.199.108.0/22",
        "140.82.112.0/20",
        "143.55.64.0/20"
    ];
    
    let callerIp = null;
    if (event['headers'] && event['headers']['x-forwarded-for'])
    {
        callerIp = event['headers']['x-forwarded-for'];
        if (isIp4InCidrs(callerIp, listCidrGh)) {
            response['isAuthorized'] = true;
        }
    }
    return response;
};
```

Once deployed, a simple `curl -XPOST https://<ID>.execute-api.<REGION>.amazonaws.com/webhook` from your laptop should result in a `{"message":"Forbidden"}` (while webhook events from repositories are correctly accepted)

I tired to make it as compact and fast as possible. Possible, we could add some part to automatically fetch [GitHub CIDR](https://api.github.com/meta), but I'm unsure about this approach for runners infrastructure which may have a lot of events going through the webhook.

Feel free to indicate me in which of the current examples you are eager to see this example, I can add it to this PR.

> One to Authorize (secret, CIDR). And give module users the option to overwrite the auzhorizer by an own lambda (including handling the secret

Purely a personal preference here, but having the secret handled "by design" and intrinsically in the module feel more reassuring from a client point of view.

> I also looking at the moment at the EventBride. Splitting the lambda could also empower to allow other ways of directing events into the runner module. Asuming authorization is already done.

Unfortunately, I never had the opportunity to work on EventBridge, so I'm not knowledgable on this !

Feel free to ask if you have other requests/insights!

Regards,
Nicolas

npalm:
The PR is still in my scope, but have to dig in what direction we will take in the module. Would you problem also solved if we add an option to the webhook to check based on an allow list?

NSpenlihauer:
Hello @npalm , sorry for the delay ! 

> Would you problem also solved if we add an option to the webhook to check based on an allow list

Sure, as soon as we have a way to harden webhook access based on GitHub's CIDRs it will be fine for us ! 

About the short therm solution (the one without eventbridge), do you have any insights/clues about the way users may overload the authorizer lambda ? 

I guess we could:
- Create a new kind of lambda in [modules/webhook/lambdas/webhook](https://github.com/philips-labs/terraform-aws-github-runner/tree/main/modules/webhook/lambdas/webhook/src) named **webhook_authorizer**
- Migrate [verifySignature](https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/webhook/lambdas/webhook/src/webhook/handler.ts#L90) function in this newly created lambda
- Add CIDR based validation in webhook_authorizer. CIDR list could be retrieved from AWS parameter store (parameter which could be generated as code and possibly overloaded by the used) ? 

Having parameters (CIDR) dynamically retrieved from the environment could grant customization without having to create/maintain a customized authorizer on user side.

Are you aligned with this approach ? 

Have a great day ! 
Nicolas

PR #3040, opened 2023-03-08
fix(pool): ensure pool top up respects var.ami_id_ssm_parameter_name

### LINKED ISSUE #3038: AMI ID SSM parameter name is not respected when using pool
Hello,

@npalm We are testing ephemeral runners + pool + prebuilt AMIs (no userdata script), and it looks like the `ami_id_ssm_parameter_name` is not taken into account with pooled instances.

This means that instances are booted up with the default amazon linux AMI, which then of course does not work as it does not include the runner software, and hence the instance does not register itself to GitHub.

Looking at the code, the pool call to `createRunners` does in fact omit this parameter: https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/runners/lambdas/runners/src/pool/pool.ts#L90

I can try to whip up a PR to fix this.

Cheers,
Jukka

jpalomaki:
Fixes #3038

### DIFF
--- a/modules/runners/policies-lambda-common.tf
+++ b/modules/runners/policies-lambda-common.tf
@@ -8,3 +8,26 @@ data "aws_iam_policy_document" "lambda_assume_role_policy" {
     }
   }
 }
+
+resource "aws_iam_policy" "ami_id_ssm_parameter_read" {
+  count       = var.ami_id_ssm_parameter_name != null ? 1 : 0
+  name        = "${var.prefix}-ami-id-ssm-parameter-read"
+  path        = local.role_path
+  description = "Allows for reading ${var.prefix} GitHub runner AMI ID from an SSM parameter"
+  policy      = <<-JSON
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Action": [
+            "ssm:GetParameter"
+          ],
+          "Resource": [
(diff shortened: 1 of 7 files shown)


modules/runners/policies-lambda-common.tf
```diff
@@ -8,3 +8,26 @@ data "aws_iam_policy_document" "lambda_assume_role_policy" {
...
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Action": [
+            "ssm:GetParameter"
+          ],
+          "Resource": [
+            "arn:${var.aws_partition}:ssm:${var.aws_region}:${data.aws_caller_identity.current.account_id}:parameter/${trimprefix(var.ami_id_ssm_parameter_name, "/")}"
```
npalm, line 27: Note to myself, create issue to refactor the parameter path, should inline with the other parameters.

npalm (commented):
Looks good, need to run a quick test

jpalomaki:
FWIW, I have been testing this branch in our AWS environment, specifying `var.ami_id_ssm_parameter_name`, and it seems to work fine (both just-in-time provisioning and pool top up).

What I have not tested, however, is _not_ specifying `var.ami_id_ssm_parameter_name`, so that should probably be tested as well, for possible regressions.

npalm:
Tested, looks good thanks @jpalomaki

npalm (approved):

PR #3042, opened 2023-03-09
refactor(runners): tests for runners

npalm:
However all test are running find there where some small mistakes in checks and test data. 

- Added overview of tests instances to understand easier what the test set is
- Fixed seval issues in test like checks and test data

The tests for scale down are now working, no effect on production code. To get the tests better under control we should do more work in the future to clean them up and clarify the test data.


bartgolsteijn (approved):
LGTM

PR #3043, opened 2023-03-09
feat(runners): add option to prefix registered runners in GitHub

### LINKED ISSUE #3034: Add prefix to runner name
## Problem
When GitHub sent the`workflow_job.in_progress` event, the vent contains information about the runner, runner group and workflow job event. With only this information it is hard to connect with a runner deployment / configuration. The GitHub runner created by the module is named equal to the AWS instance id. This makes it not easy to make the correaltion, especially weh multiple runner configuration are active to connect a GitHub runner back to a configuration. It is certainly possible to find the match based on the instance ID. This always requires API calles. 

## Solution
To make the correlation easier a logical name for the runner / runner configuration could be added as prefix or post fix to the GitHub runner name. 

## Implementation 
- Making this change should be done with some care. Since the module is currently relying on that the name is prefixed with the instance-id to make matches for scale down and pool decicsions.
- The change should be done preferable in such way there is no (mandatory) impact on users having theri own start scripts.
- The current mechanism is used from day one of the module, which means there could some unexpected impact. 
- By adding a configurable part to the name we should check what the max length of the runner name is. Max length of runner name is 64, max instance name is 17 + two chars for `i-`.
- Suggest by default use the prefix to postfix the runner name. 
- The instances Id should remain in the runner name, this is the only way to correlate the ec2 instance to a github runner.

cc: @GuptaNavdeep1983

npalm:
Fixes #3034

## Description
This PR add the option to prefix the name registered in GitHub. By default no prefix will be used.

- The prefix max length is check to ensure the max of 64 chars is not met.
- For scale down the startWith check is replaced by endWith check to match the instance-id
- Choose prefix over postfix since it will make more sense with sorting

### Now tags

- EC2 Tag: `ghr:runner_name_prefix` added 
- EC2 Tag: `ghr:created_by` added values either `scale-up-lambda` or `pool-lambda`

## Start scripts

For custom AMI's scripts can be updated to retrieve the AWS EC2 Tag `RunnerNamePrefix`
```
runner_name_prefix=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:runner_name_prefix) || true
```
Next the $runner_name_prefix can be combined with the instanceId the name. When you would like to add a seperator, add this to the terraform variable `runner_name_prefix`.

Example of runners based on multi-runner example:
image

## Verificaton

Deployed versions main, trigger builds. Next update to branch (rebuild lambda, and ensure runners are active). Next upgrade environment. Checked that old build runners are removed by scale-down, checked new build runners are removed by scale down. 

Verififed for:
- [x] mult-runner example
- [x] default example


modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -214,6 +215,7 @@ export async function createRunner(runnerParameters: RunnerInputParameters): Pro
             ResourceType: 'instance',
             Tags: [
               { Key: 'ghr:Application', Value: 'github-action-runner' },
+              { Key: 'ghr:created_by', Value: numberOfRunners === 1 ? 'scale-up-lambda' : 'pool-lambda' },
```
GuptaNavdeep1983, line 218: did not add the tag runner_name_prefix here?

modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -44,6 +44,7 @@ export interface RunnerInputParameters {
     instanceAllocationStrategy: EC2.SpotAllocationStrategy;
   };
   numberOfRunners?: number;
+  runnerNamePrefix: string;
```
GuptaNavdeep1983, line 47: Is this variable used?

npalm: Not any more, should be removed

modules/runners/lambdas/runners/src/aws/runners.ts
npalm: That tag is defined in the the launch template. Like others.

modules/runners/lambdas/runners/src/aws/runners.ts
npalm: Also cleanup the pool lambda

modules/runners/templates/start-runner.sh
```diff
@@ -16,17 +16,21 @@ echo "Retrieved INSTANCE_ID from AWS API ($instance_id)"
 %{ if metadata_tags == "enabled" }
 environment=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:environment)
 ssm_config_path=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:ssm_config_path)
+runner_name_prefix=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:runner_name_prefix) || true
```
GuptaNavdeep1983, line 19: is true a good enough value for this variable?

npalm: true is not a value, only ensure the script not crashing case of an error. But can remove it

npalm: Set the default to an empty string explicit in case of an error

GuptaNavdeep1983 (approved):

PR #3044, opened 2023-03-10
feat: added outputs for runner log groups.

GuptaNavdeep1983:
Runner level log groups can be very useful to create a metric around the errors generated during runner execution. Generated these outputs from runners module.


modules/runners/outputs.tf
```diff
@@ -42,6 +42,10 @@ output "role_pool" {
   value = try(module.pool[0].role_pool, null)
 }
 
+output "runners_log_groups" {
```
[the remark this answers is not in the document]
GuptaNavdeep1983: Its a combination of multiple log groups.

modules/runners/outputs.tf
```diff
@@ -42,6 +42,11 @@ output "role_pool" {
...
+  description = "List of log groups from different log files of runner machine."
+  value = aws_cloudwatch_log_group.gh_runners
```
npalm: maybe better with a try, potentially the value can be null

GuptaNavdeep1983: fixed null check, not sure why terraform throws error for null output!

npalm (approved):

PR #3046, opened 2023-03-10
fix(runners): increase the log level to WARN when using the enable_job_queued_check parameter

Fgerthoffert:
The readme section detailing the `enable_job_queued_check` makes it an appealing function, but it was unclear to me that it could cause some issues when used with ephemeral runners and matrix workflows. 

It is particularly challenging to notice for organizations with a very large number of repositories, at times with low activity the issue is not present, but when matrix jobs are started or when multiple repositories are triggering jobs around the same time window (such as cron'd workflow), this issue is more likely to be faced.

For example, on the execution that allowed me to find the cause, when starting a matrix of 9 jobs, 4 of them were actually triggering this log message:
```
2023-03-10 15:29:09.728  INFO  [scale-up:a0502c68-f9b9-5d2d-afd6-70edb7486316 index.js:135249  isJobQueued] Job not queued 
```
This log message, aside from being one amongst many other `INFO` messages, does not provide many details about the consequences.

I suggest bumping the level to `WARN` and adding more details in the message itself.

Related discussion: https://github.com/philips-labs/terraform-aws-github-runner/discussions/2931


npalm (approved):
LGTM

npalm:
I think the logging messages we have can be improved. Thanks for your contribution. I have also created a PR to replace the current framework (which is out-dated). The replacement will not direct change the log messages. But provides with each messages the context. And we drop support for line based logging, moving to json which is much easier to parse. See #3037

PR #3067, opened 2023-03-15
fix: allow the instances to send metrics

rgoomar:
This is needed in order to be able to send metrics around the disk and memory. Those were getting a 403 until this update was made.

Example error in the Cloudwatch agent logs:
```
not authorized to perform: cloudwatch:PutMetricData because no permissions boundary allows the cloudwatch:PutMetricData action
        status code: 403, request id: 
```

### DIFF
--- a/modules/setup-iam-permissions/policies/boundary.json
+++ b/modules/setup-iam-permissions/policies/boundary.json
@@ -13,7 +13,8 @@
         "resource-groups:*",
         "ssm:*",
         "ssmmessages:*",
-        "ec2messages:*"
+        "ec2messages:*",
+        "cloudwatch:*"
       ],
       "Resource": "*"
     },


npalm (approved):

npalm:
Just wondering, the change is only adding the permissions to the boundary condition. Assume yout attach a policy to the instance profile as well? Can you share the script you use to push memory metrics?

rgoomar:
> Just wondering, the change is only adding the permissions to the boundary condition. Assume yout attach a policy to the instance profile as well? Can you share the script you use to push memory metrics?

This is actually just fixing the behavior that should happen. The runners add a [policy](https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/runners/policies/instance-cloudwatch-policy.json#L7) that has `cloudwatch:PutMetricsData` which is helpful if you set `enable_cloudwatch_agent` flag. So, it will have the correct IAM policy in place along with the CW agent on the runner although if you look at the logs just from those default values, it fails to actually send over metrics due to a 403 error from the permissions boundaries. 

All I did was go into the console and manually update the role with this permission boundary change and metrics like memory usage on the host starting sending properly.

Issue #2782, opened 2022-12-19
Runner not starting when using user other than ec2-user


christophersavory:
I am using v1.18.0 and trying to use an AMI user other than ec2-user in the runner.  Our origanization publishes its own AMIs and they remove the ec2-user user for security reasons.

It looks like the support is halfway there for this.

I can see in start-runner.sh there is the ability to pull the `run_as` user from SSM with these lines
```
run_as=$(echo "$parameters" | jq --arg environment "$environment" -r '.[] | select(.Name == "/\($environment)/runner/run-as") | .Value')
echo "Retrieved /$environment/runner/run-as parameter - ($run_as)"
```
That part works fine. 

The problem is the outer script, user-data.sh doesn't have that same capability and will fail on `usermod -a -G docker ec2-user` if `ec2-user` does not exist. 

It looks like install-runner.sh is also using `user_name`, which is defaulting to `ec2-user` on this line from user-data.sh:
```
user_name=ec2-user
```

christophersavory:
I might have been using a version that is newer than v1.18.0 since I cloned the github repo directly.

npalm:
Thanks for creating the issue, indead a bug in https://github.com/philips-labs/terraform-aws-github-runner/blob/801e01f290407975aea11b85e44ac9743f9173cb/modules/runners/templates/user-data.sh#L23

Feel free to reais a PR to solve the issue.

christophersavory:
@npalm I'm trying to create a PR for this, but I don't have permissions to push a branch to your repo.

npalm:
@christophersavory that is correct and normal in Open Source, please check the CONTRIBUTING.md. But  in general the way of working is. For the repo (own copy). Here you can create branches. Push your change and create a PR back to the source reposiotry.

npalm:
Yes indeed, create a fork. In you fork it is up to you to create branch or not. I would suggest create a branch. Commit and push you changes to the branch (in your fork). And next you can make a PR.

christophersavory:
@npalm writing/refactoring script files isn't really my forte. But I think I got it working at least for my use case of running as a different user other than `ec2-user`.
I am not able to test the Windows PowerShell scripts though.  

Here is the PR for you to look at. https://github.com/philips-labs/terraform-aws-github-runner/pull/2900

christophersavory:
@npalm can you reopen? I will rebase the PR as you asked.

christophersavory:
[...]

PR #3070, opened 2023-03-16
fix: #2782 update the order of some of the metadata during the runner setup

christophersavory:
Currently this project only works for AMIs that have an `ec2-user` that is setup.  This works fine for all the AMIs that are provided by AWS. However, companies that are using their own AMIs will usually have a separate user that is preferred and the `ec2-user` is often disabled due to security concerns.  

In the current files, `user-data.sh` will add `ec2-user` to do the docker group and it will also do a variable assignment
```
    user_name=ec2-user
```
After that, install-runner.sh and start-runner-sh are called where the `run_as` varable is setup.
```
    run_as=$(echo "$parameters" | jq -r '.[] | select(.Name == "'$ssm_config_path'/run_as") | .Value')
```

What needs to happen is the Setup should happen first (new script created called setup-runner.sh). And then the install/run scrips called.


npalm:
Thanks for taking the time to create a PR, please can you have a look on the failing CI?

christophersavory:
@npalm CI is passing now

PR #3075, opened 2023-03-19
fix(syncer): replaced aws-sdk v2 by aws-sdk v3

npalm:
## Description

AWS SDK v2 is now in maintenance mode and suggest to update to v3, see https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/


npalm:
Checked with a deployment with the default example:

image

ScottGuymer (approved):

PR #3108, opened 2023-03-24
feat(syncer): enable S3 bucket versioning for syncer S3 bucket

reedloden:
Many compliance obligations require S3 bucket versioning be enabled in order to meet specific integrity and availability controls.

Considering the limited objects in this bucket and minimal cost associated with keeping any versions, enabling it by default.


modules/runner-binaries-syncer/main.tf
```diff
@@ -70,6 +70,13 @@ resource "aws_s3_bucket_logging" "action_dist_logging" {
   target_prefix = var.s3_logging_bucket_prefix != null ? var.s3_logging_bucket_prefix : var.distribution_bucket_name
 }
 
+resource "aws_s3_bucket_versioning" "action_dist" {
+  bucket = aws_s3_bucket.action_dist.id
+  versioning_configuration {
+    status = "Enabled"
```
npalm, line 76: I don't see any good readon to version this bucket. It is just a cache. Version the bucket will result in the runner wil use the latest available version. Unless you update the runner start script. If you are subject to regulation it might be better to build a custom AMI to control the version.

But supporting versioning for this bucket is fine. Would suggest that you a variable to allow users to enable versioning.

This requires 
- Adding a variable support the values of the field Status, with default `Disabled`
- Add the variable to the root, runner-binary-syncer and multi runner.
- Naming suggestion: top-level and multi-runner runner_binaries_s3_versioning, and on the module it self  s3_versioning

npalm (changes requested):

modules/runner-binaries-syncer/main.tf
reedloden: I implemented the ability to configure S3 bucket versioning as you requested. Let me know if that works for you.

variables.tf
```diff
@@ -141,6 +141,12 @@ variable "runner_binaries_s3_sse_configuration" {
   }
 }
 
+variable "runner_binaries_s3_versioning" {
+  description = "Status of S3 versioning for runner-binaries S3 bucket."
```
npalm, line 145:
```suggestion
  description = "Status of S3 versioning for runner-binaries S3 bucket. Once set to Enabled the change cannot be reverted via Terraform!"
```

modules/multi-runner/variables.tf
```diff
@@ -324,6 +324,12 @@ variable "runner_binaries_s3_sse_configuration" {
   default     = {}
 }
 
+variable "runner_binaries_s3_versioning" {
+  description = "Status of S3 versioning for runner-binaries S3 bucket."
```
npalm, line 328:
```suggestion
  description = "Status of S3 versioning for runner-binaries S3 bucket. Once set to Enabled the change cannot be reverted via Terraform!"
```

npalm (changes requested):
@reedloden thanks the PR is good. But added two small suggestion to the docs to indicate the change cannot be reverted via Terraform. I also fixed some conflicts by main. Removed the terraform docs from the readme, they will be generated once on the release branch. You can also enable the CI on your fork (even better), this will ensure the docs will be re-created.

GuptaNavdeep1983 (approved):

npalm (approved):

PR #3129, opened 2023-03-29
feat(images): automatically find latest GitHub Runner version when building images

gautamg795:
This requires using a newer version of `packer` to get the `http` data source, but this now allows an operator to omit the `runner_version` variable and still ensure that the built image is using the newest runner, removing the need for an autoupdate when the runner starts up.

Previous behavior of overriding the variable when desired is maintained.


npalm (commented):
@gautamg795 looks good to me. Please go ahead with updating the other image examples

gautamg795:
fyi @npalm, I've made the requested updates and this should be ready for review

npalm (approved):

PR #3131, opened 2023-03-29
SET msbuild path

MariusKorzeniauskas:
After msbuild install, run commands:

  $msBuildPath = "C:\VisualStudio\MSBuild\Current\bin"
  $registryPath = "Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment"
  $oldPath = (Get-ItemProperty -Path $registryPath -Name PATH).path
  $newPath = "$oldpath;$msBuildPath"
  Set-ItemProperty -Path "$registryPath" -Name PATH -Value "$newPath"


npalm:
Please can you describe why raising this PR? What problem is getting resolved.

MariusKorzeniauskas:
In every action we have to call these commands:

$msBuildPath = "C:\VisualStudio\MSBuild\Current\bin"
$registryPath = "Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment"
$oldPath = (Get-ItemProperty -Path $registryPath -Name PATH).path
$newPath = "$oldpath;$msBuildPath"
Set-ItemProperty -Path "$registryPath" -Name PATH -Value "$newPath"

because system variables not set correctly.

PR #3133, opened 2023-03-30
docs: removing duplicate terraform-docs sections

tobiasehlert:
The tool terraform-docs is using `BEGIN_TF_DOCS` and `END_TF_DOCS` as markers for docs generations.
So with this the duplicate terraform-docs sections are removed from the README files.


npalm:
closing, adapted changes in #3140.

Thanks for your contribution!

PR #3138, opened 2023-03-30
fix(runners): upgrade aws sdk v2 to v3

npalm:
Updates the AWS sdk for the runners lambda's from v2 to v3

Tasks:

- [x] - update code
- [x] - run test with actual deployed runners

## Notes

- Refactored types, moved types for runners to a separate file
- Tests, mocks replaces, test should be unchanged


modules/runners/lambdas/runners/src/aws/runners.ts
```diff
@@ -84,22 +96,12 @@ function constructFilters(filters?: ListRunnerFilters): Ec2Filter[][] {
 }
 
 async function getRunners(ec2Filters: Ec2Filter[]): Promise<RunnerList[]> {
-  const ec2 = new EC2();
-  const runners: RunnerList[] = [];
-  let nextToken;
-  let hasNext = true;
-  while (hasNext) {
```
npalm, line 91: As far I can figure out the aws v3 sdk will by default fetch all pages if the token is not set. @ScottGuymer @GuptaNavdeep1983 can you do a double check?

PR #3139, opened 2023-03-31
chore(examples): multi-runner example with separate configuration files

koendelaat:
The example shows how runner configurations can be configured in Yaml files.
This has the following advantages:
- Compacter files for better readability
- Possibility to use CODEOWNERS file on the different configurations.


npalm (approved):
LGTM, thanks @koendelaat

PR #3142, opened 2023-03-31
feat(lambda): Add support for X-Ray tracing

mcaulifn:
Add support for x-ray


mcaulifn:
@npalm Not sure what is up with the tfdocs workflow

koendelaat:
> @npalm Not sure what is up with the tfdocs workflow

I was facing the same issue.
I've resolved it here: https://github.com/philips-labs/terraform-aws-github-runner/pull/3143/commits/e70bf2ff109ad068d1d80553be1a5537f888c6b3

npalm:
I recently move the update of the tfdocs to the release branch. But this conflicts with the pre-commit hook with tfdocs. So therefore I removed the pre-commit. And run the workflow on the PR. But that seems not working. Will dig in next week.

.github/workflows/update-docs.yml
```diff
@@ -18,13 +18,15 @@ jobs:
...
         with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
           ref: ${{ github.event.pull_request.head.ref }}
 
       - if: github.event_name == 'push'
         name: Checkout branch
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # ratchet:actions/checkout@v3
 
       - name: Generate TF docs
+        if: github.event.pull_request.head.repo.full_name == github.repository
```
npalm, line 29: This will avoid to generate docs for PR's at all. Which means it does almost make no sense to run the workflow on the pull_request event.

Just created a PR to run the docs action on branches. For forks that have actions enabled it will update the docs. If not enabled no impact. And docs will updated once released (or via PR to main).

See https://github.com/philips-labs/terraform-aws-github-runner/pull/3144

What do you think?

mcaulifn: Removing these fixes in favor of yours.

npalm (commented):
Looks good, will check in detail later. But it seems you missed the pool lambda, which is in as submodule of the runners.

npalm (commented):
@mcaulifn thx, I ran a quick test and only see tracing for the scale-down, not for the scale-up. Need to dig in a bit more next wee. 

Besides that can you also add the tracing variable to multi-runner module.

mcaulifn:
Added to multi- and pool.

npalm (changes requested):
Tested and got a working with some tweaks:

- multi-runner module file runner-binaries.tf add `lambda_tracing_mode = var.lambda_tracing_mode`
- multi-runner module file webhook.tf add `lambda_tracing_mode = var.lambda_tracing_mode`
- missed a policy for the pool see inlince comment.

changes also available on branch https://github.com/philips-labs/terraform-aws-github-runner/tree/mcaulifn/xray

npalm:
> Tested and got a working with some tweaks:
> 
> * multi-runner module file runner-binaries.tf add `lambda_tracing_mode = var.lambda_tracing_mode`
> * multi-runner module file webhook.tf add `lambda_tracing_mode = var.lambda_tracing_mode`
> * missed a policy for the pool see inlince comment.
> 
> changes also available on branch https://github.com/philips-labs/terraform-aws-github-runner/tree/mcaulifn/xray

@mcaulifn just pushed my addition to your branch

npalm (approved):

PR #3144, opened 2023-04-01
chore(ci): fix tf-docs on PR

npalm:
We moved recently to generate Terrafom docs on the release branch. However it would be nice to have up-to-date docs on the main branch. In the past we used pre-commit. But requires a contributor have local tools setup. Keeping pre-commit and the tf-dcos action creates a conflict.

This PR will address the issues and introducee a sub optimal solution. I could not find an option to update from a Pull Request trigger the source repo or suggest the change via a PR. 

The solution in this PR
- Run the workflow on every branch
- For the main branch (protected) create a PR

Looking up if a branch is protected is possible via the API, but this API call is not allowed by the GitHub action token. It requires a personal token with more permission. 

To run the wofklow on forks, the forked repo needs to enable action. If not enabled nothing will stop conribution. Once merged. Either via a PR the main will be updated with the docs. Or via the release branch.

@koendelaat @mcaulifn any suggestion are welcome

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,9 +51,9 @@ If your issue appears to be a bug, and hasn't been reported, open a new issue. H
 
 **If you get help, help others. Good karma rulez!**
 
-### Submitting a Merge Request
+### Submitting a Pull Request
 
-Before you submit your merge request consider the following guidelines:
+Before you submit your pull request consider the following guidelines:
 
 * Make your changes in a new git branch:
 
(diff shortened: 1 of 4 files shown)


mcaulifn (approved):
I like ! :shipit:

CONTRIBUTING.md
```diff
@@ -63,7 +63,8 @@ Before you submit your merge request consider the following guidelines:
 
 * Create your patch, **including appropriate test cases**.
 * Install [Terraform](https://www.terraform.io/). We lock the version with [tvenv](https://github.com/tfutils/tfenv), check `required_version` in `versions.tf` for the current development version of the module.
```
GuptaNavdeep1983, line 65:
```suggestion
* Install [Terraform](https://www.terraform.io/). We lock the version with [tfenv](https://github.com/tfutils/tfenv), check `required_version` in `versions.tf` for the current development version of the module.
```

GuptaNavdeep1983 (changes requested):

GuptaNavdeep1983 (approved):

PR #3194, opened 2023-04-18
lowercase distribution_bucket_name

dmitrykiselev27:
Prefix may contain uppercase characters that will
break the distribution bucket name validation


npalm (commented):
LGTM, thx!

PR #3196, opened 2023-04-19
fix(multi-runner): VPC and Subnet ID configurable for multi runner

prashanthbgoud:


modules/multi-runner/runners.tf
```diff
@@ -3,8 +3,8 @@ module "runners" {
   for_each      = local.runner_config
   aws_region    = var.aws_region
   aws_partition = var.aws_partition
-  vpc_id        = var.vpc_id
-  subnet_ids    = var.subnet_ids
+  vpc_id        = each.value.runner_config.vpc_id
```
npalm, line 6: Could you try someting like this. If `each.value.runner_config.vpc_id` is not set set var.vpc_id. An dsimilar for the subnets

By doing this we can keep the change non breaking. And it avoid in many cases extra configuration.

prashanthbgoud: I will have a look today

prashanthbgoud:
@npalm can you have a look?

GuptaNavdeep1983 (approved):

npalm (changes requested):
some small suggestions

npalm:
@prashanthbgoud is the PR still relevant?

PR #3219, opened 2023-04-25
fix(multi-runner):  convertdistribution_bucket_name to lowercase

spcaipers-arm:
This is a mandatory spec.

At my company, we have some forcing capital letter prefix conventions and we are using the var.prefix for that. But the bucket name you can not use capital letters. I had to patch your module like that in order to work. I am assuming this is a harmless change that you can take advantage of that.


npalm (approved):
@spcaipers-arm thanks!

PR #3220, opened 2023-04-25
fix: Changed the ami filters to ensure that AMI is available before its used in launch template.

GuptaNavdeep1983:
This PR addresses the need to make the AMI's used by the launch template as available which is not working with current set of filters, at the same time we want to allow the users to specify their own filters.

Added a default entry in the map to only find the AMI's with state as available and also added a validation to ensure that the entry is also present event after the override.


modules/runners/variables.tf
```diff
@@ -124,20 +124,27 @@ variable "instance_types" {
...
 
-variable "ami_filter" {
-  description = "Map of lists used to create the AMI filter for the action runner AMI."
+variable "ami_filters_default" {
+  description = "Default Map of lists used to create the AMI filter for the action runner AMI."
+  type        = map(list(string))
+  default     = { state = ["available"] }
+}
+
+variable "ami_filters_custom" {
```
npalm, line 133: Not clear to me why a new variable is introduced. Would prefer if we can keep it simple and have only one variable for the filter.

npalm (approved):
Looks good to me

PR #3221, opened 2023-04-25
fix(multi-runner): allow runner_additional_security_group_ids to apply to multi_runner_config

darklab-sh:
Currently, when specifying the runner_additional_security_group_ids input within the multi_runner_config block, it is ignored. It has to be specified outside the multi_runner_config block, which means all runner configs will receive the same additional security groups.  In our use-case, we need to specify different security groups for each config.

This PR supports adding the runner_additional_security_group_ids input either inside or outside of the multi_runner_config block. If added outside, the behavior does not change. All runner configs will receive the security groups. If added inside the multi_runner_config block, the specified security group(s) will only be added to the specified runner config.

I've tested all use cases I can think of, ran the pre-commit checks and updated the documentation. Please let me know if anything else is needed.


modules/multi-runner/variables.tf
```diff
@@ -63,6 +63,7 @@ variable "multi_runner_config" {
       runner_name_prefix                      = optional(string, "")
       runner_run_as                           = optional(string, "ec2-user")
       runners_maximum_count                   = number
+      runner_additional_security_group_ids    = optional(list(string), [])
```
npalm, line 66: Can you add the variable to description below, so it will be added to the docs. Docs are generated based on terraform descriptions

darklab-sh: Got it! I've added the variable with description below.

modules/multi-runner/runners.tf
```diff
@@ -49,7 +49,7 @@ module "runners" {
   idle_config                          = each.value.runner_config.idle_config
   enable_ssm_on_runners                = each.value.runner_config.enable_ssm_on_runners
   egress_rules                         = var.runner_egress_rules
-  runner_additional_security_group_ids = var.runner_additional_security_group_ids
+  runner_additional_security_group_ids = concat(each.value.runner_config.runner_additional_security_group_ids, var.runner_additional_security_group_ids)
```
npalm, line 52: I think it is better if we tak here a slightly different approach. Not merge both groups, but only use one. And the moste specific take precedence.

```suggestion
  runner_additional_security_group_ids = coalesce(each.value.runner_config.runner_additional_security_group_ids, var.runner_additional_security_group_ids)
```

npalm: @darklab-sh can you have a look on my suggestions?

npalm (dismissed):

npalm (approved):

PR #3224, opened 2023-04-26
refactor: structure of lambda code

npalm:
- moved lambda code out of terraform modules
- use yarn workspaces with lerna to avoid duplication code


.github/workflows/release.yml
```diff
@@ -26,43 +27,44 @@ jobs:
...
-        run: echo "name=${GITHUB_REF#refs/heads/}" >> $GITHUB_OUTPUT
-      - name: Release
-        id: release
-        uses: google-github-actions/release-please-action@c078ea33917ab8cfa5300e48f4b7e6b16606aede # ratchet:google-github-actions/release-please-action@v3
-        with:
-          default-branch: ${{ steps.branch.outputs.name }}
-          release-type: terraform-module
-          token: ${{ steps.token.outputs.token }}
+
+  # release:
```
npalm, line 38: Disabled release part to test workflow as much as possible

lambdas/functions/control-plane/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "github-runner-lambda-scale-runners",
+  "name": "@terraform-aws-github-runner/control-plane",
```
npalm, line 2: renamed module from runners to control-plan. Will rework the naming further in future PRs

PatrickMennen (approved):
Overall looks quite good, you can move the eslint configuration to the highest level and extend it from there in your sub-packages, just as you do with your typescript config

for the near future:
if your build and test steps will remain the same, you could integrate an nx config, which in turn will be picked up by lerna to build and test your modules

npalm:
> Overall looks quite good, you can move the eslint configuration to the highest level and extend it from there in your sub-packages, just as you do with your typescript config
> 
> for the near future: if your build and test steps will remain the same, you could integrate an nx config, which in turn will be picked up by lerna to build and test your modules

moved eslint as well

.ci/Dockerfile
```diff
@@ -1,26 +1,16 @@
...
+FROM build as lambdas
+COPY lambdas /lambdas
+RUN --mount=type=cache,target=/work/node_modules,id=lambdas \
+        yarn install --frozen-lockfile && yarn build
 
 FROM scratch as final
-COPY --from=runner-binaries-syncer /lambda/runner-binaries-syncer.zip /runner-binaries-syncer.zip
-COPY --from=runners                /lambda/runners.zip                /runners.zip
-COPY --from=webhook                /lambda/webhook.zip                /webhook.zip
+COPY --from=lambdas /lambdas/functions/gh-agent-syncer/runner-binaries-syncer.zip /runner-binaries-syncer.zip
```
ScottGuymer, line 14: This is likely to trip up someone in future if a new function is added..

A better way may be to run a command in the lambdas stage to bring these zips together in a folder or even zip them..

and then copy that single file/folder here

ScottGuymer (approved):
Looks good. only one comment.

.ci/Dockerfile
npalm: This Dockerfile is only for convience to let users build the lambda's locally without Docker. Agree could be improved. But OK if we do this later. And create an issue?

npalm: fixed

PatrickMennen (approved):

PR #3244, opened 2023-05-02
feat(runners): allow explicitly standard or unlimited

alexjurkiewicz:
Allow configuring T instance types as explicitly standard or unlimited
credit model.

This can help to control costs when using T instances as a "cpu-limited"
runner pool.


GuptaNavdeep1983 (dismissed):

modules/runners/variables.tf
```diff
@@ -594,3 +594,14 @@ variable "lambda_tracing_mode" {
   type        = string
   default     = null
 }
+
+variable "credit_specification" {
```
npalm, line 598: Please can you add the variable also to the module multi-runner. Please define the variable on the level of the runner_config https://github.com/philips-labs/terraform-aws-github-runner/blob/97c4ee7d1145c5f10ea01bbe4a81e1e6e827cef9/modules/multi-runner/variables.tf#L30

Would set the default to null, and ensure you pass the variable in the module via runners.tf in this module

npalm (changes requested):
Thanks looks good in general, one change request. The runners can be created from the root and the submodule multi-runners

modules/runners/variables.tf
alexjurkiewicz: done

GuptaNavdeep1983: @alexjurkiewicz, I dont think that this is what Niek mentioned in his comment. In this case, same config will apply 

You have defined the variable at the level of the multi runner module but inside multi runner module, there is a variable multi_runner_config which defines the config for all individual runner configs supported by this module. Its this location under which he requested you to create this variable from where it can be controlled for individual runner configs. As of now the same credit specification would apply to all runner configs.

alexjurkiewicz: my bad, I understand now! I've made the requested change

GuptaNavdeep1983:
Tested the changes and they look good. Thanks.

npalm (approved):

PR #3246, opened 2023-05-02
feat: added the option to generate outputs from packer builds.

GuptaNavdeep1983:
Added option in all packer configurations to generate output manifest file to easily get the generated artifact(AMI-ID) details.


npalm (approved):
LGTM

PR #3251, opened 2023-05-04
fix(syncer): S3 bucket logging prefix variable condition

wzyboy:
The "bucket prefix" variable should have a different validation regex than "bucket name" variable but the code uses the same. The PR fixes the issue.

Note that in root module, the regex is correct, but in "runner-binaries-syncer" sub-module, the regex is incorrect.


npalm:
Thanks for taking time to raise a PR. Will check asap

npalm (approved):
@wzyboy thx

PR #3283, opened 2023-05-22
fix: added additional outputs to multi runner module.

GuptaNavdeep1983:


modules/multi-runner/outputs.tf
```diff
@@ -42,3 +42,8 @@ output "webhook" {
 output "ssm_parameters" {
   value = module.ssm.parameters
 }
+
+output "workflow_job_queue" {
```
npalm, line 46: Why not add this one as part of the webhook? Like it is done in the root module?

npalm (changes requested):
Small question. Also still thinking that we have to get rid of this queue and move to the eventbridge (eventually)

modules/multi-runner/outputs.tf
npalm: Maybe better to wrap in queues like the root module

npalm (approved):

Issue #2848, opened 2023-01-03
Support Amazon Linux 2022 AMI


markandrus:
First, thanks for this great project!

Second, I was able to get version 2.0.0 of this project working with [Amazon Linux 2022](https://aws.amazon.com/linux/amazon-linux-2022) on arm64 by applying the following changes:

<details><summary>user-data.sh</summary>

```diff
--- a/modules/runners/templates/user-data.sh	2023-01-03 11:04:23
+++ b/modules/runners/templates/user-data.sh	2023-01-03 14:20:55
@@ -18,11 +18,11 @@
 yum update -y
 
 # Install docker
-amazon-linux-extras install docker
+yum install -y docker
 service docker start
 usermod -a -G docker ec2-user
 
-yum install -y amazon-cloudwatch-agent curl jq git
+yum install -y amazon-cloudwatch-agent jq git
 
 user_name=ec2-user

```

</details>

<details>
<summary>install-runner.sh</summary>

```diff
--- a/modules/runners/templates/install-runner.sh
+++ b/modules/runners/templates/install-runner.sh
@@ -43,7 +43,7 @@ echo "Delete tar file"
 rm -rf $file_name
 
 if [[ "$architecture" == "arm64" ]]; then
-  yum install -y libicu60
+  yum install -y libicu
 fi
 
 os_id=$(awk -F= '/^ID/{print $2}' /etc/os-release)
```
</details>

 I applied these changes in a hacky way by overriding `userdata_template` and abusing `userdata_pre_install`. I'll spare the details on what I did in `userdata_pre_install`, but here is how I set them:

```tf
  runner_architecture  = "arm64"
  ami_filter           = { name = ["al2022-ami-2022.*-kernel-5.*-arm64"] }
  userdata_template    = "user-data.sh"
  userdata_pre_install = file("pre-install.sh")
```

Could this project could support Amazon Linux 2022 AMI natively by tweaking `user-data.sh` and `install-runner.sh`?

npalm:
I last checked Amazon 2022 a few weeks ago (back in 2022). At that time it was still not GA. We have some troubles with supporting node18 on Amazon Linux 2 as well. Would be great to support or move to 2022 in case it is stable. But for maintenace purpose we don't want to support only a few version by default. 

Looking on your changes, the work to support Amazon 2022 should be minimal. So I would prefer a way we can support both for a short time. Or support it in such way that they work with the same config. The last one I prefer.

[...]

PR #3293, opened 2023-05-25
feat: Add Support for Amazon Linux 2023

eabrouwer3:
First attempt at resolving issue #2848. However, because the latest supported Amazon Linux version is now 2023, this PR has been updated to use `al2023` by default.


npalm:
I think it would be better that we replace Amazon Linux 2 by default to 2022

eabrouwer3:
@npalm I agree, if we're ok with that. I'll update my PR to do that.

eabrouwer3:
@npalm I've checked my PR according to the `CONTRIBUTING.md` guidelines, squashed all my commits into one, and also updated the PR to bump up to al2023 instead of al2022 because that's the most current supported version. I couldn't find all the necessary details for the packer files for the al2022 versions. Let me know what you think and if I need to change/update anything else!

GuptaNavdeep1983:
@eabrouwer3 All the files relevant for packer are under images folder and the install/start scripts are under modules folder.

eabrouwer3:
@GuptaNavdeep1983, what I mean is that I couldn't find the `al2022` AMI when searching in the EC2 console. And I needed to find that to double check the image name format, and get the `owner` for the packer file. I figured everything out for the `al2023` image though.

eabrouwer3:
Closing this PR in favor of making it on a new branch (made the PR off of my `main`, making it hard to sync with the forked repo's `main`).

PR #3303, opened 2023-05-31
fix: Fix pool logic with runner name prefix

pbarnajc:
The logic for determining the current pool runner count was using a map that uses runner names as the key. It would then attempt to get from that map using EC2 instance IDs, however this would never find the runners if there was a runner prefix used.

### DIFF
--- a/lambdas/functions/control-plane/src/aws/runners.d.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.d.ts
@@ -9,6 +9,7 @@ export interface RunnerList {
   type?: string;
   repo?: string;
   org?: string;
+  runnerName: string;
 }
 
 export interface RunnerInfo {
@@ -16,6 +17,7 @@ export interface RunnerInfo {
   launchTime?: Date;
   owner: string;
   type: string;
+  runnerName: string;
 }
 
 export interface ListRunnerFilters {
--- a/lambdas/functions/control-plane/src/aws/runners.test.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.test.ts
@@ -153,6 +153,7 @@ describe('terminate runner', () => {
       instanceId: 'instance-2',
       owner: 'owner-2',
       type: 'Repo',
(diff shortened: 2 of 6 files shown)


GuptaNavdeep1983:
@pbarnajc Thanks for submitting the PR. I think it will be easier to grab the RUNNER_NAME_PREFIX environment variable and use it to remove prefix from runner name before adding it to the map.

It could be expensive to fetch the tags from the aws api for each metric and is done in a loop.

pbarnajc:
> @pbarnajc Thanks for submitting the PR and is a good find. I think it will be easier to grab the RUNNER_NAME_PREFIX environment variable and use it to remove prefix from runner name before adding it to the map.
> 
> ```
> const runnerNamePrefix = process.env.RUNNER_NAME_PREFIX;
> runner.name = runnerNamePrefix ? runner.name.replace(runnerNamePrefix, ''): runner.name;
> ```
> 
> It could be expensive to fetch the tags from the aws api for each metric and is done in a loop.

Thanks for the review. That is much cleaner!

GuptaNavdeep1983:
> > @pbarnajc Thanks for submitting the PR and is a good find. I think it will be easier to grab the RUNNER_NAME_PREFIX environment variable and use it to remove prefix from runner name before adding it to the map.
> > ```
> > const runnerNamePrefix = process.env.RUNNER_NAME_PREFIX;
> > runner.name = runnerNamePrefix ? runner.name.replace(runnerNamePrefix, ''): runner.name;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It could be expensive to fetch the tags from the aws api for each metric and is done in a loop.
> 
> Thanks for the review. That is much cleaner!

Can you please format the code using yarn format to fix the build?

GuptaNavdeep1983:
Tested pool with this PR and found out that the logic to find the existing idle runner is restored in case runner names are prefixed.

Issue #3277, opened 2023-05-17
Require SSL Interaction with S3 Bucket


sgtoj:
Can the S3 bucket policy be updated to restrict access to the bucket to be via SSL only? A sample of the policy is below. I can submit PR.

```tf
data "aws_iam_policy_document" "bucket_policy" {
  sid       = "ForceSSLOnlyAccess"
  effect    = "Deny"
  actions   = ["s3:*"]
  resources = [local.bucket_arn, "${local.bucket_arn}/*"]

  principals {
    identifiers = ["*"]
    type        = "*"
  }

  condition {
    test     = "Bool"
    values   = ["false"]
    variable = "aws:SecureTransport"
  }
}
```

GuptaNavdeep1983:
@sgtoj, Thanks for reporting this issue. 

Can you please create a PR?

sgtoj:
PR created: https://github.com/philips-labs/terraform-aws-github-runner/pull/3342

ryanoolala:
thanks i'm looking for this!

sgtoj:
I am closing because #3342 was merged.

PR #3342, opened 2023-07-03
feat(syncer): update bucket policy to require ssl

sgtoj:
This change will update bucket policy to always require SSL interaction, as reported in #3277. With this change, the module will always create and attach a bucket policy. However, the previous optional force SSE policy still remains optional.


GuptaNavdeep1983:
@sgtoj Thanks for submitting this PR. I have tested the changes with default example.

GuptaNavdeep1983 (approved):

PR #3351, opened 2023-07-05
feat(images): Allow specifying temporary security group source IP for packer builds

rslotte:
Currently when packer runs an image build the security group attached to the ec2 instance has an ingress rule that allows access on port 22 from 0.0.0.0/0. In our enterprise environment this is immediately flagged in Security Hub as a critical security finding.

This change will allow packer to restrict the ingress rules of the security group it creates to the IP of the machine running packer (in our case the IP of the Github runner doing our AMI builds).

### DIFF
--- a/images/linux-amzn2/github_agent.linux.pkr.hcl
+++ b/images/linux-amzn2/github_agent.linux.pkr.hcl
@@ -78,7 +78,7 @@ variable "custom_shell_commands" {
 }
 
 variable "temporary_security_group_source_public_ip" {
-  description = "When enabled, use public IP of the host (obtained from https://checkip.amazonaws.com) as CIDR block to be authorized access to the instance, when packer is creating a temporary security group."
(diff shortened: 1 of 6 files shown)


GuptaNavdeep1983:
@rslotte Thanks for submitting the PR. I see that we also support variable `security_group_id` in the variables of packer and as per packer documentation if `security_group_id` is specified, `temporary_security_group_source_public_ip` would be ignored. 

Wouldn't it make sense to add some information about this in the description of the variable  `temporary_security_group_source_public_ip`

rslotte:
> @rslotte Thanks for submitting the PR. I see that we also support variable `security_group_id` in the variables of packer and as per packer documentation if `security_group_id` is specified, `temporary_security_group_source_public_ip` would be ignored.
> 
> Wouldn't it make sense to add some information about this in the description of the variable `temporary_security_group_source_public_ip`

Thanks @GuptaNavdeep1983 I've updated the description.

GuptaNavdeep1983 (approved):

PR #3354, opened 2023-07-10
feat: tag runner volumes with the same tags as the instance

galargh:
#### Description

Currently, volumes attached to runners are only tagged with tags specified by the AWS Launch Template. In particular, they're missing `Owner` tag, which can only be applied during "runtime".

This PR changes the way instances are created from the launch template. It adds tag specifications for the `volume` resource type to the instance creation request. With that input, the created volumes will have the same tags as the instance itself.

During the creation of this PR I noticed I had to perform a few more actions to be able to test it. I added the new instructions to the contribution guide.

I also added terraform lock file to git ignore because `terraform init` creates it and I didn't see it being committed before.

#### Testing

- [x] Ran `yarn test` in `lambdas` directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,4 +24,3 @@ secrets.auto.tfvars
 
 node_modules/
 
-.terraform.lock.hcl
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ secrets.auto.tfvars
 **/coverage/*
 
 node_modules/
+
+.terraform.lock.hcl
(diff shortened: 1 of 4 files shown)


GuptaNavdeep1983:
@galargh, Thanks for submitting this PR. All the examples already have .lock.hcl files and it was not required at the root of the repository.

GuptaNavdeep1983 (approved):

galargh:
Thank you 🙇

PR #3366, opened 2023-07-18
feat(lambdas): Add branch white list capability to webhook lambda

rslotte:
This PR adds a new branch white list functionality to the webhook Lambda.

In our use case we have certain workflows that need to access our production network. If we simply add a repository white list then developers would still be able to open pull requests adding arbitrary workflows that can execute code within the production network. By limiting the branch that can utilize the prod specific runner group we can rely on branch protection rules to ensure to un-reviewed code is not executed in our production network.

The workflow_job payload includes the branch name under workflow_job object as documented here:
https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=queued#workflow_job

and I have verified it under a runner Github app:

workflow_job


npalm:
Thanks for taking the time to submit a PR. I have some doubts about supporting this feature. Mayve even having the allow list for repo's was not a good choice. But this was introduced in the early days of Actions.

WIth this PR you not protect for using the runners for certain branches, you only protect from scaling. As soon a runner becomes avaialble the job will start. 

Maybe it is an idea that you have a look on template wofflows for dedicated runner groups, see https://github.blog/2022-03-23-github-actions-secure-self-hosted-runners-specific-workflows/ This still does not protect you for your branches. But that you can maybe solve be defining in you re-usable workflows in the job with a condition.

rslotte:
> Thanks for taking the time to submit a PR. I have some doubts about supporting this feature. Mayve even having the allow list for repo's was not a good choice. But this was introduced in the early days of Actions.
> 
> WIth this PR you not protect for using the runners for certain branches, you only protect from scaling. As soon a runner becomes avaialble the job will start.
> 
> Maybe it is an idea that you have a look on template wofflows for dedicated runner groups, see https://github.blog/2022-03-23-github-actions-secure-self-hosted-runners-specific-workflows/ This still does not protect you for your branches. But that you can maybe solve be defining in you re-usable workflows in the job with a condition.

Very interesting, thanks for the information.

I think we will then explore using Terraform to configure the Github app and runner group repository whitelists directly, then explore forking the runner binary to achieve the branch whitelisting we are looking for.

PR #3370, opened 2023-07-18
fix(lambda): added variable validation for variable lambda_tracing_mode.

GuptaNavdeep1983:


npalm (approved):
@navdeepg2021 I have added a fix to the validation condition. Validation was failing for null values.

PR #3371, opened 2023-07-19
fix: broken AMI fileter

GuptaNavdeep1983:
After introducing the default value of `{state = ["Available"]}`, we missed to update the logic around the use of default filters and the custom filters.
With this change:

 - If `ami_filter` is NOT changed by the root module user, it will be merged with default value of `local.default_ami` and the output will have combination of default_ami and `{state = ["Available"]}`.
 - If `ami_filter` is CHANGED by the root module user with a new value of `name`, the default value will be overriden as a result of the merge operation.
 - if `ami_filter` is explicitly set to null, variable validation will fail and not proceed.

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -38,7 +38,7 @@ locals {
   }
 
   ami_kms_key_arn = var.ami_kms_key_arn != null ? var.ami_kms_key_arn : ""
-  ami_filter      = coalesce(var.ami_filter, local.default_ami[var.runner_os])
+  ami_filter      = merge(local.default_ami[var.runner_os], var.ami_filter)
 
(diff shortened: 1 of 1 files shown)


npalm (approved):

npalm:
@navdeepg2021 thanx for the quick fix!

kkorotkov:
@navdeepg2021 , @npalm This fix is not backwards compatible and introduces an unwanted limitation. It took me quite some time to pinpoint the issue. The logic fails if the root module user applies `ami_filter` without including a `name` filter. In this scenario, it combines the `name` from `local.default_ami` with the rest of the user filter, almost certainly resulting in a query that returns nothing.

PR #3373, opened 2023-07-20
fix: add state for multi-runner default ami-filter

npalm:
After changed to defaults for AMI filter, changes were not applied to the multi-runners defaults


npalm:
Tested with the multi-runners example

PR #3375, opened 2023-07-20
feat(runners): add configurable eviction strategy to idle config

maschwenk:
We do some on-instance caching so when we scale down we'd prefer to keep the older instances around instead of the new ones (because they will have a hotter cache). This adds a configurable setting to the idleConfig to pick a sorting strategy. Never contributed to this repo, so please tell me if I'm doing something wrong!


lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -148,10 +148,12 @@ async function evaluateAndRemoveRunners(
   scaleDownConfigs: ScalingDownConfig[],
 ): Promise<void> {
   let idleCounter = getIdleRunnerCount(scaleDownConfigs);
+  const evictionStrategy = getEvicationStrategy(scaleDownConfigs);
   const ownerTags = new Set(ec2Runners.map((runner) => runner.owner));
 
   for (const ownerTag of ownerTags) {
-    const ec2RunnersFiltered = ec2Runners.filter((runner) => runner.owner === ownerTag);
+    const ec2RunnersFiltered = ec2Runners.filter((runner) => runner.owner === ownerTag)
+      .sort(evictionStrategy === 'oldest_first' ? oldestFirstStrategy : newestFirstStrategy)
```
maschwenk, line 156: We default to the existing strategy of `oldest_first` by default

maschwenk:
@npalm I rebased and fixed the prettier failures as well as one typecheck failure. I'm running into failures for tests because of missing type coverage. Will wait to make sure you approve of the approach before proceeding to write tests.

npalm:
> @npalm I rebased and fixed the prettier failures as well as one typecheck failure. I'm running into failures for tests because of missing type coverage. Will wait to make sure you approve of the approach before proceeding to write tests.

Thanks for the PR. WIll check the approach ASAP.

GuptaNavdeep1983:
@maschwenk one of the test case is failing for this PR. Can you please fix that?

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -191,17 +194,25 @@ async function terminateOrphan(instanceId: string): Promise<void> {
...
+function oldestFirstStrategy(a: RunnerInfo, b: RunnerInfo): number {
+  if (a.launchTime === undefined) return 1;
+  if (b.launchTime === undefined) return 1;
+  if (a.launchTime < b.launchTime) return 1;
+  if (a.launchTime > b.launchTime) return -1;
+  return 0;
+}
+
+function newestFirstStrategy(a: RunnerInfo, b: RunnerInfo): number {
+  if (a.launchTime === undefined) return 1;
```
npalm, line 206: You can use here oldestFirstStrategy, for example

```return oldestFirstStrategy(a, b) * -1;```

**Review by npalm — COMMENTED** [review]
The approach looks good to me. I understand the idea of keeping caches aroudn. But we see this as a risk as well. Supporting this eviction strategy is fine as long you esnure you add a test case.

**Comment by maschwenk** [comment]
@npalm 👍🏼 sounds good. Will fix.

**Comment by maschwenk** [comment]
@npalm I've added a test to make sure the default config is extracted correctly

@GuptaNavdeep1983 took on your suggestion as well 👍🏼

File: lambdas/functions/control-plane/src/scale-runners/scale-down-config.ts [hunk 3]
```diff
@@ -2,10 +2,12 @@ import parser from 'cron-parser';
 import moment from 'moment';
 
 export type ScalingDownConfigList = ScalingDownConfig[];
+export type EvictionStrategy = 'newest_first' | 'oldest_first';
 export interface ScalingDownConfig {
   cron: string;
   idleCount: number;
   timeZone: string;
+  evictionStrategy?: EvictionStrategy;
```
maschwenk, line 10: I changed this to be an optional type so that nobody needs to mess with their existing TF config and make sure that the types are accurate (because we do a cast)

lambdas/functions/control-plane/src/scale-runners/scale-down-config.ts
```diff
@@ -25,3 +27,12 @@ export function getIdleRunnerCount(scalingDownConfigs: ScalingDownConfigList): n
   }
   return 0;
 }
+
+export function getEvictionStrategy(scalingDownConfigs: ScalingDownConfigList): EvictionStrategy {
+  for (const scalingDownConfig of scalingDownConfigs) {
+    if (inPeriod(scalingDownConfig)) {
+      return scalingDownConfig.evictionStrategy ?? 'oldest_first';
```
maschwenk, line 34: This ensures we always default to `oldest_first` in both the case of not configuring and also being outside of the cron window

maschwenk:
@npalm Ahh I just realized it's failing due to the test coverage? I couldn't figure out why before. Happy to add a bit of coverage.

maschwenk:
@npalm Added a test in `lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts` to get the coverage level to acceptable

maschwenk:
@npalm 😅 the Code health thing did not like some duplication I had, cleaned that up. This should be ready for review! 🙏🏼

maschwenk:
@npalm should be ready for another look 🙏🏼 sorry for the churn

lambdas/functions/control-plane/src/scale-runners/scale-down-config.ts
```diff
@@ -25,3 +27,12 @@ export function getIdleRunnerCount(scalingDownConfigs: ScalingDownConfigList): n
   }
   return 0;
 }
+
+export function getEvictionStrategy(scalingDownConfigs: ScalingDownConfigList): EvictionStrategy {
+  for (const scalingDownConfig of scalingDownConfigs) {
```
npalm, line 32: Can you add a debug logger line here that writes which strategy will be applied?

npalm (changes requested):
Changes to the Lambda looks good. But also the Terraform code needs to be adjusted to pass the variable to the lambda.

Can you update:
- variables files in root module, runner module and multi-runner module
- Suggest to add an `optional(string)` or `optional(string, "oldest_first)`.
- The README in the root explains the idle_config, can you update this section as well

maschwenk:
@npalm Added all that here I think [3c7d840](https://github.com/philips-labs/terraform-aws-github-runner/pull/3375/commits/3c7d8404157b219881ee79f68012904c05354007)

Unsure how best to edit the READMEs with the HTML. Formatting there is a bit rough. Let me know if there's anything I missed. I had missed the note in the CONTRIBUTING.md to allow the Actions to run on my fork, now a bit easier to see how things are working.

README.md
```diff
@@ -294,9 +294,12 @@ The module will scale down to zero runners by default. By specifying a `idle_con
...
 ```hcl
 idle_config = [{
-   cron      = "* * 9-17 * * 1-5"
-   timeZone  = "Europe/Amsterdam"
-   idleCount = 2
+   cron             = "* * 9-17 * * 1-5"
+   timeZone         = "Europe/Amsterdam"
+   idleCount        = 2
+   # Defaults to 'oldest_first', use 'newest_first' if you want to evict newest runners first,
+   # for example if the oldest runners have a long-living cache
```
npalm, line 301: Looks like in that case you would remove the newest first

README.md
```diff
@@ -294,9 +294,12 @@ The module will scale down to zero runners by default. By specifying a `idle_con
```
npalm, line 294: Can you add here a sentence like:

By default the oldest istances are removed. This helps to keep your environment up-to-date and reduce problems as full disks or memory issues. You can overwrite this behavior by setting the `evictionStrategy`.

npalm (commented):
Looks in gneral good, a few remarks. Will test asap

maschwenk:
@npalm I think my comment above the eviction strategy was more confusing so I just moved it up into the paragraph above to give a more thoughtful explanation. Completely aside from the point of this PR, your note of:

> This helps keep your environment up-to-date and reduce problems like running out of disk space or RAM

The disk space clause makes sense to me, but I'd be curious how often your instances are carrying around "dead" RAM after jobs run on them? We definitely have RAM issues but because we don't run things containerized it often just toasts the box completely and we take it out of rotation versus leaking over time. Just curious about your experience there. Thanks for reviewing!

maschwenk:
@npalm were you able to take a look?

npalm:
> @npalm were you able to take a look?

Sorry not yet, seems also the coerage is dropped slightly. I wil have a look early next week

npalm:
Just ran the test suite locally, but see no coverage error. Will have a check later.

npalm:
> Just ran the test suite locally, but see no coverage error. Will have a check later.

Re-ran the build, seems no fine.

npalm (approved):
@maschwenk thx for your contribution

maschwenk:
@npalm I'm seeing those failures you are getting on master, I think I have an idea of why they are happening. Will upstream a change if I figure it out.

npalm:
Do you mean the release build on main? For some reason the build is randomly failing. I have no clue why (yet). The documentation build is another problem. And a less a big issue. Do you have any clue why the release build is failing when building all lambda's?

maschwenk:
> Do you mean the release build on main? For some reason the build is randomly failing. I have no clue why (yet). The documentation build is another problem. And a less a big issue. Do you have any clue why the release build is failing when building all lambda's?

@npalm I don't yet, but my test was mutating `process.env`, so I was thinking perhaps the mutation was polluting global state. Haven't been able to repro locally. It's also odd because the stack-trace is pretty useless, not clear what the actual error is, just shows where it's coming from.

lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
```diff
@@ -431,6 +430,19 @@ describe('scaleDown', () => {
         expect(mockOctokit.apps.getRepoInstallation).toBeCalled();
         expect(terminateRunner).not.toBeCalled();
       });
+
+      describe('With newest_first eviction strategy', () => {
+        beforeEach(() => {
+          process.env.SCALE_DOWN_CONFIG = JSON.stringify([{ ...defaultConfig, evictionStrategy: 'newest_first' }]);
```
npalm, line 436: @maschwenk Your suggestion is right. We need to ensure SCALE_DOWN_CONFIG is alwasy set to the default / undefined before the test. Can you add tot the most outer before the statement process.env.SCALE_DOWN_CONFIG = 'oldest_first'

npalm:
I think I have solved the issue in #3418

npalm:
Not yet, still failing

npalm:
@maschwenk I have tried earlier today to fix the test. I am not able to reproduce the failing test on my local. Assume it has something to do with the order. I added the SCALE_DOWN_CONFIG to other test to ensure the value is set right. At some moment got a error the coverage was not metting our settings. Fixed that as well. THought it was all good. But now the main is failing again. So ondering if you have still any ieda.

maschwenk:
@npalm still at a little bit of a loss. Is there any way we can get better logs out of the failing tests perhaps? Or maybe try forcing the Jest test to run in an order that will deterministically fail?

npalm:
Yes we can set a few things in jest (at least)
- verbose -> true
- silent -> false
- define reporters. and save as artiefact (retention 1 day).

I will also dig in a bit more next week.

npalm:
This PR should fix the issue: #3432

maschwenk:
@npalm ❤️

PR #3377, opened 2023-07-24
fix: Runners module variable name change.

GuptaNavdeep1983:
Name of the variable `runner_extra_labels` changed to `runner_labels` in runners module as now the runners module needs all the labels.


lambdas/libs/aws-powertools-util/src/logger/index.ts
```diff
@@ -4,7 +4,7 @@ import { Context } from 'aws-lambda';
 const childLoggers: Logger[] = [];
 
 const defaultValues = {
-  region: process.env.AWS_REGION || 'N/A',
+  region: process.env.AWS_REGION,
```
GuptaNavdeep1983, line 7: somehow coverage is reduced without this change.

npalm (approved):

lambdas/libs/aws-powertools-util/src/logger/index.ts
npalm: strange, but setting a default is indeed not needed. The value is always set by aws

PR #3393, opened 2023-07-27
fix: allow disable JIT config for ephemeral runners

npalm:
The JIT config introduced for ephemeral runners seems not be supported (yet) by GHES. This PR adds an option to disable jit config, and use registration token instead.

fix #3392

PR #3407, opened 2023-08-02
fix(runners): ensure `/opt/start-runner-service.sh` is run with Bash

mcieno:
#3393 broke (Ubuntu) ephemeral runners when `enable_jit_config = true`.

The introduction of `enable_jit_config` came with an update of script `/opt/start-runner-service.sh` that required it to run in a shell with *double bracket* keyword support:

https://github.com/philips-labs/terraform-aws-github-runner/blob/cfbcc944fc183b481caaee323e7832ec1964eb54/modules/runners/templates/start-runner.sh#L123-L129

Currently, script `/opt/start-runner-service.sh` is chmoded executable but has no shebang.
In Ubuntu, this results in `nohup` running it with plain `sh`, hence the following error log:

```log
Starting the runner as user ubuntu
Starting the runner in ephemeral mode
/opt/start-runner-service.sh: 4: [[: not found
Starting without JIT config
An error occurred: Not configured. Run config.(sh/cmd) to configure the runner.
Runner listener exit with terminated error, stop the service, no retry needed.
Exiting runner...
Runner has finished
Stopping cloudwatch service
Terminating instance
```

These changes ensure that `nohup` runs the script with `bash`.

### DIFF
--- a/modules/runners/templates/start-runner.sh
+++ b/modules/runners/templates/start-runner.sh
@@ -135,9 +135,8 @@ cat >/opt/start-runner-service.sh <<-EOF
   echo "Terminating instance"
   aws ec2 terminate-instances --instance-ids "$instance_id" --region "$region"
 EOF
-  chmod 755 /opt/start-runner-service.sh
   # Starting the runner via a own process to ensure this process terminates
-  nohup /opt/start-runner-service.sh &
+  nohup bash /opt/start-runner-service.sh &
 
 else
   echo "Installing the runner as a service"


npalm:
@GuptaNavdeep1983 maybe similar to the issue you had with the AMI start script.

npalm (approved):
Thanks for you contribution!

Looks good, als ran a quick verification.

npalm:
@GuptaNavdeep1983 maybe similar to the issue you had with the AMI start script.

PR #3409, opened 2023-08-04
feat: add failover to on-demand in case request is failing

npalm:
Adding the option to create on-demand instances in case spot is failing.

## Problem

This module either support spot or on-demand instances. When using spot (default), creation can fail with several reasons. In case there is no sufficient capacity on AWS it makes sens to request on-demand instances. AWS does not support this kind of requests via the fleet API. This PR addresses this problme and add the option (opt-in_ to create on-demand instances in case of Insufficient capacity. 

## Migrations

No migrations required

## Opt-in

Opt in by setting the variable`enable_runner_on_demand_failover`

## Verfication

Not easy to test the failover. But due to changes in multi-runner, runner module as well lambda scale-up and pool.  The following checks are required

- [x] Ephemeral example with pool
- [x] Multi runner example


npalm:
> As far as I can tell, it all looks good.
> 
> Do you want to address the points from CodeScene?

No, and actually the PR decreases the complexity by splitting the method in function. I think we can ignore codescne here

GuptaNavdeep1983 (dismissed):

npalm:
@navdeepg2021 thx, review comments are addressed

lambdas/functions/control-plane/src/aws/runners.ts
```diff
@@ -167,13 +167,11 @@ async function processFleetResult(
...
     if (
       errors.some((e) => runnerParameters.onDemandFailoverOnError?.includes(e)) &&
-      runnerParameters.onDemandFailoverOnError &&
       runnerParameters.ec2instanceCriteria.targetCapacityType === 'spot'
     ) {
       logger.warn(`Create fleet failed, initatiing fall back to on demand instances.`);
       logger.debug('Create fleet failed.', { data: fleet.Errors });
-      const numberOfInstances =
-        (runnerParameters.numberOfRunners ? runnerParameters.numberOfRunners : 1) - instances.length;
+      const numberOfInstances = (runnerParameters.numberOfRunners ?? 1) - instances.length;
```
GuptaNavdeep1983, line 174: i dont think numberOfRunners is nullable anymore!

lambdas/functions/control-plane/src/aws/runners.test.ts
```diff
@@ -140,6 +141,28 @@ describe('list instances', () => {
...
+      Filters: [
+        { Name: 'instance-state-name', Values: ['running'] },
+        { Name: 'tag:ghr:Application', Values: ['github-action-runner'] },
+      ],
+    });
+  });
+
+  it('Filter instances with status undefined, fall back to defaults.', async () => {
+    mockEC2Client.on(DescribeInstancesCommand).resolves(mockRunningInstances);
+    await listEC2Runners({ statuses: undefined });
```
GuptaNavdeep1983, line 158: why this test is required? who is calling listec2runners with that status?

GuptaNavdeep1983 (approved):

PR #3143, opened 2023-03-31
chore!: Change runner labels to list[string] and apply TFLint

koendelaat:
Change variables `runner_labels` and `runner_extra_labels` to type: `list[string]` to better represent the content

Additionally:
- [x] Automatic determine runner labels for multi runner based on label matchers
- [x] Enable Terraform linting on all modules, including naming conventions
- [x] Enable check annotations for linting errors
- [x] Fix existing linting errors
- [ ] Apply naming conventions within objects (not supported by [TFLint](https://github.com/terraform-linters/tflint/issues/1310))

PR #3410, opened 2023-08-05
fix!: Change runner labels to list[string] and apply TFLint

npalm:
Change variables `runner_labels` and `runner_extra_labels` to type: `list[string]` to better represent the content

Additionally:
- [x] Automatic determine runner labels for multi runner based on label matchers
- [x] Enable Terraform linting on all modules, including naming conventions
- [x] Enable check annotations for linting errors
- [x] Fix existing linting errors
- [ ] Apply naming conventions within objects (not supported by [TFLint](https://github.com/terraform-linters/tflint/issues/1310))

This PR is based on #3143 creaed by @koendelaat but now via a dedicate branch in this repository.


npalm:
@koendelaat I have revived your PR in a dedicated branch, added you as contributor as well. Once ready will merge the PR in a branch next. I plannign also to do some major refactoring on variables.

npalm:
@koendelaat  can make a comment what your idea was about naming conventions for objects?

koendelaat:
@npalm some objects are used in submodules in Terraform, other are directly passthrough to software components. The latter need to follow the coding style of that component.

For the first one I would suggest using codestyle of Terraform.

Points to consider:
- this will lead to inconsistent coding styles for objects
- There is currently no automated way to lint these code styles (afaik)

npalm:
> @npalm some objects are used in submodules in Terraform, other are directly passthrough to software components. The latter need to follow the coding style of that component.
> 
> For the first one I would suggest using codestyle of Terraform.
> 
> Points to consider:
> 
> * this will lead to inconsistent coding styles for objects
> * There is currently no automated way to lint these code styles (afaik)

@koendelaat I have addressed the issues in a dedicated issue #3411 Any thoughts on this issue?

.github/workflows/terraform.yml
```diff
@@ -41,26 +41,38 @@ jobs:
...
         name: Setup TFLint
         uses: terraform-linters/setup-tflint@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
       - if: contains(matrix.terraform, '1.3.')
         name: Run TFLint
-        run: tflint -f compact
+        run: |
+          tflint --init -c ${GITHUB_WORKSPACE}/.tflint.hcl
+          tflint -f compact -c ${GITHUB_WORKSPACE}/.tflint.hcl --var-file ${GITHUB_WORKSPACE}/.github/lint/tflint.tfvars
```
GuptaNavdeep1983, line 63: Compact format is already specified under tflint.hcl file!

modules/multi-runner/main.tf
```diff
@@ -8,7 +8,9 @@ locals {
     key_base64 = module.ssm.parameters.github_app_key_base64
   }
 
-  runner_config = { for k, v in var.multi_runner_config : k => merge({ id = aws_sqs_queue.queued_builds[k].id, arn = aws_sqs_queue.queued_builds[k].arn }, v) }
+  runner_extra_labels = { for k, v in var.multi_runner_config : k => sort(setsubtract(setunion(flatten(v.matcherConfig.labelMatchers), compact(v.runner_config.runner_extra_labels)), ["self-hosted", v.runner_config.runner_os, v.runner_config.runner_architecture])) }
```
GuptaNavdeep1983, line 11: We dont want the user to pass runner_extra_labels, so shud we not mark this as deprecated and update the code to remove the user of this variable from individual runner_configs?

sdarwin: "We don't want the user to pass runner_extra_labels" We're using runner_extra_labels to specify and customize labels. It's very important. Would another var take on that same function?

npalm: This allows to set other labels for runners. Let's keep it for as is

npalm:
Ready and tested

PR #3418, opened 2023-08-09
chore: fix test and add vscode workspace config

npalm:
- fix for randomly failing unit test after merging #3375
- add workspace config for vscode
- increase coverage


JeroenKnoops (approved):
LGTM

lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
```diff
@@ -398,6 +396,7 @@ describe('scaleDown', () => {
         idleCount: 3,
         cron: '* * * * * *',
         timeZone: 'Europe/Amsterdam',
+        evictionStrategy: 'oldest_first',
```
maschwenk, line 399: I was specifically trying not to set this because I wanted to ensure that providing no value still defaulted to the right value. Would I make sense to make the `SCALE_DOWN_CONFIG` as injectable at the method level to make it easier to test? Mutating `process.env` in tests is just asking for trouble.

PR #3437, opened 2023-08-16
feat: Add Support for Amazon Linux 2022

eabrouwer3:
Add `runner_amazon_linux_2022` boolean variable

Get rid of amazon linux 2022 config option

Make PR changes even smaller

Update all references to amzn2. Also bump to al2023 (latest supported version)

Add the al2023 packer file


eabrouwer3:
Working right now on addressing the latest comment from #3293.

eabrouwer3:
@GuptaNavdeep1983, I've fixed the `packer build` issue and rebased master onto my branch, so I think I've addressed any issues you had on the old PR.

@npalm, tagging you on this as well, as you were the "reviewer" on the old PR.

modules/runners/templates/install-runner.sh
```diff
@@ -45,9 +45,9 @@ rm -rf $file_name
 os_id=$(awk -F= '/^ID/{print $2}' /etc/os-release)
 echo OS: $os_id
 
-# Install libicu60 for arm64 on non-ubuntu
+# Install libicu for arm64 on non-ubuntu
 if [[ "$architecture" == "arm64" ]] && [[ ! "$os_id" =~ ^ubuntu.* ]]; then
-  yum install -y libicu60
+  yum install -y libicu
```
wangzw, line 50: Also required for x86_64

modules/runners/main.tf
```diff
@@ -19,7 +19,7 @@ locals {
   s3_location_runner_distribution = var.enable_runner_binaries_syncer ? "s3://${var.s3_runner_binaries.id}/${var.s3_runner_binaries.key}" : ""
   default_ami = {
     "windows" = { name = ["Windows_Server-2022-English-Full-ECS_Optimized-*"] }
-    "linux"   = var.runner_architecture == "arm64" ? { name = ["amzn2-ami-kernel-5.*-hvm-*-arm64-gp2"] } : { name = ["amzn2-ami-kernel-5.*-hvm-*-x86_64-gp2"] }
+    "linux"   = var.runner_architecture == "arm64" ? { name = ["al2023-ami-2023.*-kernel-5.*-arm64"] } : { name = ["al2023-ami-2023.*-kernel-5.*-x86_64"] }
```
wangzw, line 22: kernel-6.*  is latest, should we use the latest?

npalm: Not sure. But in case kernel 6 is working. I think we should use the latest.

wangzw: I'm currently using kernel-6 and feeling good. Please update to kernel-6 if no one against it. ^-^

Chobito:
any update on this? seems that packer check failed due rate limit

modules/runners/templates/install-runner.sh
eabrouwer3: Updated - removed the check for `arm64` arch.

modules/runners/main.tf
eabrouwer3: Updated to `kernel-6`.

npalm (commented):
Just checked the PR with mult-runner example. Both the arm64 and amazon linux runners are not coming online. On the x64 runner looks like the curl installation is failing.

wangzw:
> Just checked the PR with mult-runner example. Both the arm64 and amazon linux runners are not coming online. On the x64 runner looks like the curl installation is failing.

You are right, curl-minimal is installed by default, and conflict with curl, I forget to mention that.

@eabrouwer3

eabrouwer3:
> > Just checked the PR with mult-runner example. Both the arm64 and amazon linux runners are not coming online. On the x64 runner looks like the curl installation is failing.
> 
> You are right, curl-minimal is installed by default, and conflict with curl, I forget to mention that.

So should we remove the `curl` install then? In `modules/runners/templates/user-data.sh`. I think we still need it in the `ubuntu` files where we install it, cause this shouldn't affect any of those.

wangzw:
> > > Just checked the PR with mult-runner example. Both the arm64 and amazon linux runners are not coming online. On the x64 runner looks like the curl installation is failing.
> > 
> > 
> > You are right, curl-minimal is installed by default, and conflict with curl, I forget to mention that.
> 
> So should we remove the `curl` install then? In `modules/runners/templates/user-data.sh`. I think we still need it in the `ubuntu` files where we install it, cause this shouldn't affect any of those.

I do not think `yum install -y curl` works for ubuntu, so just remove it. Anyone who use ubuntu has to customized a `user_data.sh`

wangzw:
Looks good to me, but you need to fix the conflicts.

aktur:
@eabrouwer3 I was updating a runner to Linux 2023 recently too and have two comments:
- you should use `dnf` instead of `yum` on Linux 2023;
- to install `curl` without conflicts, use the following line `dnf install curl -y --allowerasing`.

eabrouwer3:
> @eabrouwer3 I was updating a runner to Linux 2023 recently too and have two comments:
> 
> * you should use `dnf` instead of `yum` on Linux 2023;
> * to install `curl` without conflicts, use the following line `dnf install curl -y --allowerasing`.

How do the changes in [this commit](https://github.com/philips-labs/terraform-aws-github-runner/pull/3437/commits/d6b876dbb08dfe8835a4756633cf4ec60592f7c9) look to you? I did a little bit of googling and replaced `yum update -y` with `dnf upgrade-minimal -y` as well. But I'm unsure if I should change those to be `dnf upgrade` instead. Not sure what issues we were having without the `yum update` before.

aktur:
> > @eabrouwer3 I was updating a runner to Linux 2023 recently too and have two comments:
> > 
> > * you should use `dnf` instead of `yum` on Linux 2023;
> > * to install `curl` without conflicts, use the following line `dnf install curl -y --allowerasing`.
> 
> @aktur, How do the changes in [this commit](https://github.com/philips-labs/terraform-aws-github-runner/pull/3437/commits/d6b876dbb08dfe8835a4756633cf4ec60592f7c9) look to you? I did a little bit of googling and replaced `yum update -y` with `dnf upgrade-minimal -y` as well. But I'm unsure if I should change those to be `dnf upgrade` instead. Not sure what issues we were having without the `yum update` before.


In many environments, especially in production, a conservative approach with `dnf upgrade-minimal` might be preferred to minimize potential disruptions from new features. But if you're in a development environment or if you actively want the newest features, `dnf upgrade` might be more appropriate. Anyway, these changes look OK to me, and I would stay with the upgrade-minimal.

olyhuta:
Hi here, is it supposed to be merged in the nearest future?

npalm:
Will test the PR asap. Would like to get the PR merged with other breaking changes planned.

modules/runners/templates/install-runner.sh
```diff
@@ -45,9 +45,9 @@ rm -rf $file_name
 os_id=$(awk -F= '/^ID/{print $2}' /etc/os-release)
 echo OS: $os_id
 
-# Install libicu60 for arm64 on non-ubuntu
-if [[ "$architecture" == "arm64" ]] && [[ ! "$os_id" =~ ^ubuntu.* ]]; then
-  yum install -y libicu60
+# Install libicu on non-ubuntu
+if [[ ! "$os_id" =~ ^ubuntu.* ]]; then
```
npalm, line 49: Why is arm64 removed here?

Chobito: Libicu60 is needed now in all linux architectures, not only arm64

npalm (commented):
Validatinging examples

- [x] default
- [ ] multi runners
- [ ] pre-build ami

aktur (approved):

matthowardcohere (approved):

Mogost (approved):

ivan-klass (approved):

Mogost:
Looks like this is connected to https://github.com/philips-labs/terraform-aws-github-runner/issues/3534

aktur:
> Looks like this is connected to #3534

Yes, it is. I've successfully run the PR's runner, and this error is gone.

olyhuta (approved):

devdavidkarlsson (approved):

Chobito (approved):

garethallen-daemon:
Any update on this?

PR #3447, opened 2023-08-23
feat: support for cloudwatch config on multi-runners

piscue:
## Problem
Cloud config can only set for all multi-runners and not per runner. 

## Solution
Add option to set the cloud config per runner.

### DIFF
--- a/modules/multi-runner/runners.tf
+++ b/modules/multi-runner/runners.tf
@@ -69,7 +69,7 @@ module "runners" {
   logging_retention_in_days        = var.logging_retention_in_days
   logging_kms_key_id               = var.logging_kms_key_id
   enable_cloudwatch_agent          = each.value.runner_config.enable_cloudwatch_agent
-  cloudwatch_config                = each.value.runner_config.cloudwatch_config
+  cloudwatch_config                = coalesce(each.value.runner_config.cloudwatch_config, var.cloudwatch_config)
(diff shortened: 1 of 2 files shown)


piscue:
@npalm could you take a look on this one?

modules/multi-runner/runners.tf
```diff
@@ -69,7 +69,7 @@ module "runners" {
   logging_retention_in_days        = var.logging_retention_in_days
   logging_kms_key_id               = var.logging_kms_key_id
   enable_cloudwatch_agent          = each.value.runner_config.enable_cloudwatch_agent
-  cloudwatch_config                = var.cloudwatch_config
+  cloudwatch_config                = each.value.runner_config.cloudwatch_config
```
npalm, line 72: Can you change this to something like

```coalesce(each.value.runner_config.cloudwatch_config, var.cloudwatch_config)```


And update the docs of both variables/

npalm (changes requested):
Looks good, see comment with a suggestion

modules/multi-runner/runners.tf
piscue: I did the update and the doc I've found. The rest of the documentation seems to exist already.

npalm (commented):
Looks god, wil run asap a quick test

npalm (approved):

PR #3449, opened 2023-08-23
feat: add logging to start runner script.

GuptaNavdeep1983:
To debug the issues around start runner script which is very common during any updates, added a new tag which can be used to print the statements from bash script.

### DIFF
--- a/lambdas/functions/control-plane/src/aws/runners.d.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.d.ts
@@ -39,5 +39,4 @@ export interface RunnerInputParameters {
   };
   numberOfRunners?: number;
   amiIdSsmParameterName?: string;
-  logLevel?: string;
 }
--- a/lambdas/functions/control-plane/src/aws/runners.d.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.d.ts
@@ -39,4 +39,5 @@ export interface RunnerInputParameters {
   };
   numberOfRunners?: number;
   amiIdSsmParameterName?: string;
+  logLevel?: string;
 }
--- a/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
@@ -50,6 +50,7 @@ interface CreateEC2RunnerConfig {
   ec2instanceCriteria: RunnerInputParameters['ec2instanceCriteria'];
   numberOfRunners?: number;
   amiIdSsmParameterName?: string;
+  logLevel?: string;
 }
 
 function generateRunnerServiceConfig(githubRunnerConfig: CreateGitHubRunnerConfig, token: string) {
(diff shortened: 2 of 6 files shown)


lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -205,6 +206,7 @@ export async function createRunners(
     runnerType: githubRunnerConfig.runnerType,
     runnerOwner: githubRunnerConfig.runnerOwner,
     ...ec2RunnerConfig,
+    logLevel: process.env.LOG_LEVEL || 'info',
```
npalm, line 209: I think we should stop with reading prcoess.env enverywhre. This is not helping the complexity and understanding wht is happening where. In. this case it is even not needed to read the value from the environment. You can simply request the level in `runner.ts` bij `logger.getLogLevel()`.

modules/runners/templates/start-runner.sh
```diff
@@ -32,6 +32,7 @@ availability_zone=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.
 environment=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:environment)
 ssm_config_path=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:ssm_config_path)
 runner_name_prefix=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:runner_name_prefix || echo "")
+log_level=$(curl -f -H "X-aws-ec2-metadata-token: $token" -v http://169.254.169.254/latest/meta-data/tags/instance/ghr:log_level || echo "")
```
npalm, line 35: The user-data script already read a variable log level. This code should move to the wrapper. And merged with the flag enableUserdataLogging.

lambdas/functions/control-plane/src/aws/runners.d.ts
```diff
@@ -39,4 +39,5 @@ export interface RunnerInputParameters {
   };
   numberOfRunners?: number;
   amiIdSsmParameterName?: string;
+  logLevel?: string;
```
npalm, line 42: Not needed, loglevel can be obtained via the logger. See other comments

npalm (changes requested):
The idea looks good to add support to enable activating logging in this way. It gives an option to enable debug logging without rebuilding an AMI. Pleas can you update the PR description with better reasoning why adding this feature?

lambdas/functions/control-plane/src/aws/runners.d.ts
GuptaNavdeep1983: removed

npalm:
@GuptaNavdeep1983 what shall we do with this PR?

PR #3453, opened 2023-08-26
feat(webhook): add support of disabling api gateway execute endpoint

jfeddern:
First of all I'd like to thank you for providing this awesome Terraform module. I really appreciate the work you put into this!

Currently we're using GitHub Enterprise and our company policies deny sending webhooks to domains which do not refer to company managed domains (ike the AWS provided domain for the API Gateway). Therefore, we added a custom domain to our api gateway. In order to use custom domains for the API Gateway we need to disable the default execute api endpoint, otherwise custom domains are not going to work.

At the moment Terraform detect the manually applied changes on the API Gateway in regards to disabling the execute endpoint. Hence, at each run Terraform would like to apply the old configuration. This PR adds the possibility to disable the execute endpoint.

If you have any questions to this, I'd be happy to answer it.


jfeddern:
Hi @GuptaNavdeep1983, didn't thought about a solution for custom domain on lambda function yet. As we use the API Gateway at the moment.

npalm:
The PR is still on my radar, will dig in next week

PR #3467, opened 2023-09-06
feat: allow setting VPC and subnets per runner

taharah:


examples/multi-runner/main.tf
```diff
@@ -16,6 +28,13 @@ module "base" {
   aws_region = local.aws_region
 }
 
+module "base_other" {
```
GuptaNavdeep1983, line 31: why did we add a separate base?

taharah: I can revert this change, but I thought it would be good to add to the example deploying runners to different VPCs since that is the main intent of this PR.

GuptaNavdeep1983: No, its good, thanks.

taharah:
Rebased the PR on the latest changes as well as pushed up a fix after testing this change internally.

examples/multi-runner/main.tf
```diff
@@ -3,7 +3,19 @@ locals {
   aws_region  = "eu-west-1"
 
   # Load runner configurations from Yaml files
-  multi_runner_config = { for c in fileset("${path.module}/templates/runner-configs", "*.yaml") : trimsuffix(c, ".yaml") => yamldecode(file("${path.module}/templates/runner-configs/${c}")) }
+  multi_runner_config = {
+    for c in fileset("${path.module}/templates/runner-configs", "*.yaml") :
```
npalm, line 7: This is not working for the templates not have the variables defined. 

Adding the subnet / vpc in one template is indeed a good example. But applying this example is failing. So we have to find another way of ding this.

npalm: I think this will work, which rquires renaming the template files

```
  multi_runner_config_template = {
    for c in fileset("${path.module}/templates/runner-configs", "template_*.yaml") :

    trimsuffix(c, ".yaml") => yamldecode(
      templatefile(
        "${path.module}/templates/runner-configs/${c}",
        {
          vpc_id     = module.base_other.vpc.vpc_id
          subnet_ids = jsonencode(module.base_other.vpc.private_subnets)
        }
      )
    )
  }

  multi_runner_config_static = { for c in fileset("${path.module}/templates/runner-configs", "static_*.yaml") : trimsuffix(c, ".yaml") => yamldecode(file("${path.module}/templates/runner-configs/${c}")) }

  # merge multi_runner_config and multi_runner_config_static
  multi_runner_config = merge(local.multi_runner_config_template, local.multi_runner_config_static)

```

@GuptaNavdeep1983  do you know a more elagnt approach?

npalm (changes requested):

examples/multi-runner/main.tf
maartenvanderhoef: Apologies for breaking into the discussion but I would love this feature.. 

Is it an idea to completely leave out the YAML abstraction ? A single locals structure with a config without for loops which creates 2 or 3 different runner types should be enough as an example for everyone. If people want to abstract that later with anything they can do that ?

Cheers.

npalm: We like the example to split the configuration in a YAML, helps the readability. I agree it is not a necessity. The examples are carefully choosen. This allows us to quickly run tests for PRs. The examples covers all major scenarios (except the custom AMI).

@taharah do you have time to have a look on my comments?

npalm: @taharah the PR is stale, did you had time to check the comments?

npalm:
> Hi, I've noticed that this PR seems ready for merging. Could you kindly share if there's an expected timeline for the merge? Thanks for your great work on this!

Need to check again, thought there was some issue with the examples.

taharah:
> Need to check again, thought there was some issue with the examples.

Examples should be fixed now albeit with a solution that is less ideal than I would like. I'll take a look at fixing the failing tests tomorrow morning.

taharah:
@npalm I've fixed the failing tests as well as rebased my branch.

examples/multi-runner/main.tf
```diff
@@ -3,7 +3,27 @@ locals {
...
+    for k, v in local.multi_runner_config_files :
+
+    k => merge(
+      v,
+      {
+        runner_config = merge(
+          v.runner_config,
+          {
+            subnet_ids = lookup(v.runner_config, "subnet_ids", null) != null ? module.base_other.vpc.private_subnets : null
+            vpc_id     = lookup(v.runner_config, "vpc_id", null) != null ? module.base_other.vpc.vpc_id : null
```
npalm, line 21: Example of a second VPC is nice, but this is our default example to test. Would prefer we can keep it if one vpc. I think the suggestion below hows clearly how to use it. And even test it by injecting a single subnet.

```suggestion
            subnet_ids = lookup(v.runner_config, "subnet_ids", null) != null ? [module.base.vpc.private_subnets[0]] : null
            vpc_id     = lookup(v.runner_config, "vpc_id", null) != null ? module.base.vpc.vpc_id : null
```

modules/runners/.terraform.lock.hcl
npalm, line 1: Can you remove this file?

modules/multi-runner/.terraform.lock.hcl
npalm, line 1: can you remove this file.

examples/multi-runner/main.tf
```diff
@@ -17,6 +37,13 @@ module "base" {
   aws_region = local.aws_region
 }
 
+module "base_other" {
+  source = "../base"
+
+  prefix     = "${local.environment}-other"
+  aws_region = local.aws_region
+}
+
```
npalm, line 46: The second base module can be removed with previous suggestion.
```suggestion
```

npalm (changes requested):
@taharah thanks for all the work. Change looks in general good and is working. However I prefer a single VPC in the example that we use as default for testing. I have provided suggestion to remove the second base.

taharah:
@npalm comments have been addressed and I have rebased to include the latest changes as well.

effata:
@npalm This PR seems to have been ready for merge for a month now, any chance of getting it merged soon? It solves a very concrete use case for us, and I've been running a local fork of this PR for a while and it seems to work great. 
Also, thanks for the great work with this module!

npalm:
> @npalm This PR seems to have been ready for merge for a month now, any chance of getting it merged soon? It solves a very concrete use case for us, and I've been running a local fork of this PR for a while and it seems to work great. Also, thanks for the great work with this module!

Sorry, but last weeks were busy. Will catch up with the PR asap.

npalm (approved):
@taharah thx, checked all good

taharah:
> @taharah thx, checked all good

Thank you!

PR #3491, opened 2023-09-21
feat(runners): add support to disable default labels (Linux)

jgutierrezglez:
In case the pool of runners deployed using this module are fully available for an wide GH org (no repository restrictions) any workflow configured to run in runners that contain just default labels in the runs-on definition:

e.g.

`runs-on: self-hosted`
`runs-on: Linux`

can end-up running in this pool without knowing it. That's why I have decided to remove the default labels from our runners and just rely on unique custom labels, and I believed the best way to do it by adding a runner_enable_default_labels variable - that by default is true (so, it doesn't change the current behavior), but it can help other people to deal with similar issues like the one described above.


npalm:
I think I missed GitHub changes this. In the past the default labels are added by GitHub without control. When we introduced the JIT config the default labels are not added by default. But is this the same for non JIT?

jgutierrezglez:
> I think I missed GitHub changes this. In the past the default labels are added by GitHub without control. When we introduced the JIT config the default labels are not added by default. But is this the same for non JIT?

We're using JIT - and the default labels are added by default, that's why we have created this new variable and set it to false in the cases where we just want to use custom labels

main.tf
```diff
@@ -9,7 +9,7 @@ locals {
   }
 
   default_runner_labels = "self-hosted,${var.runner_os},${var.runner_architecture}"
-  runner_labels         = var.runner_extra_labels != "" ? "${local.default_runner_labels},${var.runner_extra_labels}" : local.default_runner_labels
+  runner_labels         = !var.runner_enable_default_labels ? var.runner_extra_labels : var.runner_extra_labels != "" ? "${local.default_runner_labels},${var.runner_extra_labels}" : ""
```
npalm, line 12: This change should also be reflected to the mult-runners. This module can be used as root module as well.

See https://github.com/philips-labs/terraform-aws-github-runner/blob/35aa73ac1e6df30d4a5c314457da5646a117131b/modules/multi-runner/runners.tf#L47

npalm (commented):
Need to find some time to test.

- Changes impacted the runner module should also reflected to the multi-runner module. See comment
- I think this option requires that JIT config is enabled. Can you hadd this to the docs of the introduced variable.

sdarwin:
For the reasons explained in the original post, this feature would be great, so you can enable runners without repository restrictions.  

`runs-on: self-hosted` won't accidentally trigger a launch if only custom labels have been added.  

Maybe I am mistaken (and I might be!) but here are a few quick comments:

1. This appears to be the pull request adding the feature: https://github.com/actions/runner/pull/2443

2. It doesn't mention JIT config.  Isn't JIT completely orthogonal to default labels? Unrelated.  So remove JIT from this PR.

3.  The feature uses the flag `--no-default-labels`.   Where is the flag `--no-default-labels` in this PR?

npalm:
Sorry we had no time yet to dig in this PR. I do my best to check the PR in the coming week. Sorry for the delay.

jeff-wassouf:
Hi @npalm, any update on this PR? This will help us a lot once merged. Thank you.

iotanum:
hello, any updates on this? would be a huge help!

jgutierrezglez:
Hello @npalm, 

Is there something that we can do to speed-up the review / merge of this change ? 

Regards,

Javier G.

npalm:
sorry, shame on me. But the whole label stuff is become so complex over the years. I am a bit afraid of breaking tings. I need to have some focus time to dig in the topic deep. And maybe we should also think on how we can grow the maintainer pool of this module.

iotanum:
I've been using up-to-date module with these custom label changes for my organization for the past month, didn't encounter any issues (using another module for another AWS organization in the same GitHub org)

sdarwin:
@jgutierrezglez 

- At least in your own opinion, is this PR ready to merge?

- The github feature itself is implemented with the command-line flag `--no-default-labels` .   Otherwise default labels are added.  Please correct me if I am wrong...  How is `--no-default-labels`  not mentioned in the files of this pull request?  Without it, default labels would be applied.

jgutierrezglez:
> At least in your own opinion, is this PR ready to merge?

I really think it's ready for review. We have been using this changes in our own version of the module for more than a year without no issue (ephemeral runners)

> The github feature itself is implemented with the command-line flag --no-default-labels . Otherwise default labels are added. Please correct me if I am wrong... How is --no-default-labels not mentioned in the files of this pull request? Without it, default labels would be applied.

I don't really understand what you mean with this..the way how I implemented is:

runner_enable_default_labels var is set to true by default, so, by default nobody should notice any change. In case you set this var to false, it's when the default labels won't be added to the runner_labels var

sdarwin:
@jgutierrezglez , let's continue from this scenario:  "In case you set this var to false, it's when the default labels won't be added to the runner_labels var"

Perhaps the default labels won't be added to the runner_labels var.  That variable is 'internal' to terraform, at least one some level. It's part of terraform. But we have to ask: what about the actual real runners themselves?  Notice these references:

https://github.com/actions/runner/pull/2443 "Add --no-default-labels config option to self-hosted runners"

https://stackoverflow.com/questions/75330363/remove-default-labels-on-github-self-hosted-runner "Remove default labels on self-hosted runners"
```
The support for not adding default labels to self-hosted runners, was added in this PR:
https://github.com/actions/runner/pull/2443
It should become available in the next stable release that includes this change.
To use it, pass in:
--no-default-labels
``` 

How are you bypassing the usage of "--no-default-labels"?   It's not mentioned in this pull request.   Without applying "--no-default-labels" (which previously was not available), the runner will always apply your custom labels + the default labels, so you would always get the default labels + the custom labels, on the runner.

Investigating more, "--no-default-labels" is a flag on config.sh.  

config.sh is called in `modules/runners/templates/start-runner.sh`.  Rather than applying all the time, it is used "if the runner is non ephemeral or jit config is disabled".

Otherwise, `--jitconfig $${config}` is used.

Ok...  So from this perspective, maybe your code changes are working after all.  Great.   Perhaps they are effective with `--jitconfig $${config}` . But, still broken if `config.sh` is called.   Because `config.sh` requires "--no-default-labels".   Do you agree?

sdarwin:
I don't understand how there is any connection between a new method of terraform variable validation, and what ought to be done here: 

conditionally adding the "--no-default-labels" flag to `config.sh`.

sdarwin:
In the absence of automated CI tests, how about manual testing. 

Run through these cases and report back the findings. Not what might be expected, but what the test show.  Post screen shots perhaps.  

Let's say your chosen labels are [a,b,c]. We'll call them my_custom_labels. You can use any that you prefer. 

Check in the Github control panel UI at https://github.com/_my_org_/_my_repo_/settings/actions/runners to see what labels are ultimately applied by Terraform. Let's call that "real resulting labels".  

Fill in the blanks.  

| runner_enable_default_labels | your labels  | enable_jit_config | enable_ephemeral_runners | real resulting labels |
| ---- | ---- | ---- | ---- | ---- |
| true | my_custom_labels | false | false | |
| true | my_custom_labels | false | true | |
| true | my_custom_labels | true | false | |
| true | my_custom_labels | true | true  | |
| false | my_custom_labels | false | false | |
| false | my_custom_labels | false | true | |
| false | my_custom_labels | true | false | |
| false | my_custom_labels | true | true  | |

If all eight tests are correct, hopefully the maintainers will agree the PR is ready.

sdarwin:
Windows `modules\runners\templates\start-runner.ps1` with `config.cmd` is missing.

sdarwin:
How about this instead?

```
if [[ "$default_labels" == "true" ]]; then
    extra_flags="--no-default-labels"
else
    extra_flags=""
fi

sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh ${extra_flags} --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
```

modules/runners/runner-config.tf
```diff
@@ -12,6 +12,13 @@ resource "aws_ssm_parameter" "runner_agent_mode" {
   tags  = local.tags
 }
 
+resource "aws_ssm_parameter" "default_labels" {
```
npalm, line 15: I think a better name is `disable_default_labels` this indicates that the paramater is a boolean, and is in line with the agent paramater (diable -> no-default).

modules/runners/templates/start-runner.sh
```diff
@@ -216,7 +219,13 @@ echo "Starting the runner as user $run_as"
 # configure the runner if the runner is non ephemeral or jit config is disabled
 if [[ "$enable_jit_config" == "false" || $agent_mode != "ephemeral" ]]; then
   echo "Configure GH Runner as user $run_as"
-  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
+  if [[ "$default_labels" == "true" ]]; then
+      extra_flags="--no-default-labels"
+  else
+      extra_flags=""
+  fi
+  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh ${extra_flags} --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
```
npalm, line 227:
```suggestion
  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh $${extra_flags} --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
```

modules/runners/variables.tf
```diff
@@ -217,9 +217,20 @@ variable "runner_boot_time_in_minutes" {
...
+  description = "Enable default labels for the runners (os, architecture and `self-hosted`). If disabled, the runner will only have the extra labels provided in `runner_extra_labels`."
+  type        = bool
+  default     = true
+}
+
 variable "runner_labels" {
   description = "All the labels for the runners (GitHub) including the default one's(e.g: self-hosted, linux, x64, label1, label2). Separate each label by a comma"
   type        = list(string)
+
+  validation {
```
npalm, line 230: This is breaking, we don't know if someone is using the module with extra labels. Condition cannot be scoped outside the variable to check if default labels are disabled. 

I would remove the condition for now, and create issue to see if we can use pre-conditions checks to validate the labels array is set.

modules/runners/variables.tf
```diff
@@ -217,9 +217,20 @@ variable "runner_boot_time_in_minutes" {
   default     = 5
 }
 
+variable "runner_enable_default_labels" {
```
npalm, line 220: Can you rename this to `runner_disable_default_labels`

variables.tf
```diff
@@ -58,10 +58,21 @@ variable "runner_boot_time_in_minutes" {
   default     = 5
 }
 
+variable "runner_enable_default_labels" {
```
npalm, line 61: Variables introduced on top-level also needs to be defined in the multi-runner module. 

See also previous comment for naming.

npalm (changes requested):
Thx the PR already looks quite good. But made some comments to suggest different names. Please can you also update the following parts.

- Update the windows start script to read the parameter
- Update docs, add the paramater: https://philips-labs.github.io/terraform-aws-github-runner/configuration/#aws-ssm-parameters
- Update the multi-runner module to pass the parameter down.

npalm:
@jgutierrezglez feel free to grant maintainers access to write to the PR branch. There is a checkbox you can tick to that. In that case I can push my updates used for local testing as well.

jgutierrezglez:
> @jgutierrezglez feel free to grant maintainers access to write to the PR branch. There is a checkbox you can tick to that. In that case I can push my updates used for local testing as well.

It's already enabled, you should be able to push your updates

mohramadan911:
Is this merged already ? we are really looking forward for this PR as soon as possible- we need to separate our runners so self-hosted will act for all jobs but only runners with dependabot label

jgutierrezglez:
@npalm do you need any change from my side in order to approve this PR?

mohramadan911:
> @npalm do you need any change from my side in order to approve this PR?



> npalm

Do you think this PR can be released soon ?

sdarwin:
> do you need any change from my side in order to approve this PR?

On Oct 4, he mentioned certain requirements. For example this has not yet been done: "Update the windows start script to read the parameter".    Check the other notes from there.

jgutierrezglez:
Update the windows start script to read the parameter is done already, and I reviewed the rest of them, and normally all the remarks have been already implemented. @npalm could you check again let me know if you have any comment or remark about the last changes?

npalm:
Planned to check the PR tomorrow.

modules/runners/templates/start-runner.ps1
```diff
@@ -134,8 +137,14 @@ if ($agent_mode -eq "ephemeral") {
         Invoke-Expression ".\run.cmd --jitconfig $${config}"
     }
     else {
-        Write-Host "Starting without jit config"
-        Invoke-Expression ".\run.cmd"
+        if ($disable_default_labels -eq "true") {
```
npalm, line 140: I think all cases should support the default labels. I tried testing based on the multi-runner example. Bot got windows not working. 

I have tried some uupdates on the powershell script, but for some reason the runner starts but just applyting the diable label switch:

Screenshot 2024-11-15 at 12 58 06

I changed the Window script as follow:

```powershell
$runnerExtraOptions = ""
if ($disable_default_labels -eq "true") {
    $runnerExtraOptions += "--no-default-labels"
}

if ($agent_mode -eq "ephemeral") {
    if ($enable_jit_config -eq "true") {
        Write-Host "Starting with jit config"
        Invoke-Expression ".\run.cmd --jitconfig $${config}"
    }
    else {
        Write-Host "Starting without jit config and with default labels with runnerExtraOptions: $runnerExtraOptions"
        Invoke-Expression ".\run.cmd $runnerExtraOptions"
    }
    Write-Host "Runner has finished"

    if ($enable_cloudwatch_agent)
    {
        Write-Host "Stopping CloudWatch Agent"
        & 'C:\Program Files\Amazon\AmazonCloudWatchAgent\amazon-cloudwatch-agent-ctl.ps1' -a stop
    }

    Write-Host "Terminating instance"
    aws ec2 terminate-instances --instance-ids "$InstanceId" --region "$Region"
} else {
    if ($runnerExtraOptions) {
        $runnerArguments = "-Argument $runnerExtraOptions"
    }
    Write-Host "Installing the runner as a service with runnerArguments: $runnerArguments"

    $action = New-ScheduledTaskAction -WorkingDirectory "$pwd" -Execute "run.cmd" $runnerArguments
    $trigger = Get-CimClass "MSFT_TaskRegistrationTrigger" -Namespace "Root/Microsoft/Windows/TaskScheduler"
    Register-ScheduledTask -TaskName "runnertask" -Action $action -Trigger $trigger -User $username -Password $password -RunLevel Highest -Force
    Write-Host "Starting runner after $(((get-date) - (gcim Win32_OperatingSystem).LastBootUpTime).tostring("hh':'mm':'ss''"))"
}

```

And test the non ephemeral case, but this was not working
 
Screenshot 2024-11-15 at 12 58 41

npalm (changes requested):
@jgutierrezglez thanks for this great PR. I have tested several cases

- [ ] Default runner - works, requires a suggested change on defaults
- [x] Multi runner (linux) 
- [ ] Multi runner (linux) - with JIT
- [ ] Multi runner (windows)

For the multi runner I needed to make the follwoing adjusments:

1. mutli-runner module runners.tf

Pass parametr into runner module:
```  
runner_disable_default_labels        = each.value.runner_config.runner_disable_default_labels
```

and avoid setting defaults if labels are disabled.
```
  runner_labels                        = each.value.runner_config.runner_disable_default_labels ?  sort(distinct(each.value.runner_config.runner_extra_labels)) :  sort(distinct(concat(["self-hosted", each.value.runner_config.runner_os, each.value.runner_config.runner_architecture], each.value.runner_config.runner_extra_labels)))
```

I would say the PR is getting close to merge. To get it faster passed you could split the PR in two parts. First the linux part adding the feature, next another PR to fix the windows part.

variables.tf
```diff
@@ -58,10 +58,21 @@ variable "runner_boot_time_in_minutes" {
...
+  default     = true
+}
+
 variable "runner_extra_labels" {
-  description = "Extra (custom) labels for the runners (GitHub). Labels checks on the webhook can be enforced by setting `enable_runner_workflow_job_labels_check_all`. GitHub read-only labels should not be provided."
+  description = "Extra (custom) labels for the runners (GitHub). Separate each label by a comma. Labels checks on the webhook can be enforced by setting `enable_workflow_job_labels_check`. GitHub read-only labels should not be provided."
   type        = list(string)
   default     = []
+
+  validation {
```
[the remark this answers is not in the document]
jgutierrezglez: We can do it, but that will require the upgrade of the TF version required: https://www.hashicorp.com/blog/terraform-1-9-enhances-input-variable-validations

modules/runners/templates/start-runner.ps1
jgutierrezglez: As you suggested, I'll address this issue on a new PR dedicated to fix the issue with Windows

jgutierrezglez:
@npalm https://github.com/philips-labs/terraform-aws-github-runner/pull/4261 PR to address the Windows fix

sdarwin:
@jgutierrezglez , this is a question rather than definitely something needs to change, but why does it say `$${extra_flags}` instead of `${extra_flags}` in start-runner.sh?  What effect does that have?  isn't `extra_flags` a regular variable within the bash script that would have one `$`?

variables.tf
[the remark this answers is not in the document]
npalm: let's do that later, don't like to to drop to many versions of Terraform at once.

modules/runners/templates/start-runner.sh
```diff
@@ -216,7 +219,12 @@ echo "Starting the runner as user $run_as"
 # configure the runner if the runner is non ephemeral or jit config is disabled
 if [[ "$enable_jit_config" == "false" || $agent_mode != "ephemeral" ]]; then
   echo "Configure GH Runner as user $run_as"
-  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
+  if [[ "$disable_default_labels" == "true" ]]; then
+      extra_flags="--no-default-labels"
+  else
+      extra_flags=""
+  fi
+  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh ${extra_flags} --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
```
npalm, line 227: extra_flags needs to be escaped otherwise terraform is this expecting as a template input.


```suggestion
  sudo --preserve-env=RUNNER_ALLOW_RUNASROOT -u "$run_as" -- ./config.sh $${extra_flags} --unattended --name "$runner_name_prefix$instance_id" --work "_work" $${config}
```

variables.tf
```diff
@@ -58,10 +58,21 @@ variable "runner_boot_time_in_minutes" {
   default     = 5
 }
 
+variable "runner_disable_default_labels" {
+  description = "Disable default labels for the runners (os, architecture and `self-hosted`). If enabled, the runner will only have the extra labels provided in `runner_extra_labels`."
```
npalm, line 62:
```suggestion
  description = "Disable default labels for the runners (os, architecture and `self-hosted`). If enabled, the runner will only have the extra labels provided in `runner_extra_labels`. In case you on own start script is used, this configuration parameter needs to be parsed via SSM. For Windows no support yet."
```

modules/multi-runner/variables.tf
```diff
@@ -164,6 +165,7 @@ variable "multi_runner_config" {
         runner_additional_security_group_ids: "List of additional security groups IDs to apply to the runner. If added outside the multi_runner_config block, the additional security group(s) will be applied to all runner configs. If added inside the multi_runner_config, the additional security group(s) will be applied to the individual runner."
         runner_as_root: "Run the action runner under the root user. Variable `runner_run_as` will be ignored."
         runner_boot_time_in_minutes: "The minimum time for an EC2 runner to boot and register as a runner."
+        runner_disable_default_labels: "Disable default labels for the runners (os, architecture and `self-hosted`). If enabled, the runner will only have the extra labels provided in `runner_extra_labels`."
```
npalm, line 168:
```suggestion
        runner_disable_default_labels: "Disable default labels for the runners (os, architecture and `self-hosted`). If enabled, the runner will only have the extra labels provided in `runner_extra_labels`. In case you on own start script is used, this configuration parameter needs to be parsed via SSM. For Windows no support yet."
```

npalm (changes requested):
Made two tiny documentation suggestions. Rest looks good and tested for linux (amazon and ubuntu). 

Maybe worth to add a small section in the [docs](https://philips-labs.github.io/terraform-aws-github-runner/configuration/#configuration-considerations) to disuss the choice betwoon using default labels and disable them. 

Besides those small notes good to go 🚀

jgutierrezglez:
> Made two tiny documentation suggestions. Rest looks good and tested for linux (amazon and ubuntu).
> 
> Maybe worth to add a small section in the [docs](https://philips-labs.github.io/terraform-aws-github-runner/configuration/#configuration-considerations) to disuss the choice betwoon using default labels and disable them.
> 
> Besides those small notes good to go 🚀

I added a small note under "Configuration considerations" - let me know what you think about it

npalm (changes requested):
Tested and all good, one tiny change requested

npalm (approved):

Issue #2988, opened 2023-02-22
Option for EBS Optimized?


tedparagon:
I can't seem to find an option to enable EBS Optimized on the runner instance. Is this an oversight on my part or might there be a reason for not including this?

I believe this should just be a matter of setting `ebs_optimized` on the `aws_launch_template`.

Thanks!

jpalomaki:
@tedparagon I cannot see why we'd not want to support this. Can you post a PR to add the flag? See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#ebs_optimized

tedparagon:
@pbarnajc thanks for adding this. I see `ebs_optimized` on the runners and multi-runners modules but don't see a way of setting it from the root module. I also can't find any mention of it in the docs.

AlexShemeshWix:
I can see the ebs_optimized on runners module. But how do i enable it on main module?

AlexShemeshWix:
@pbarnajc?

PR #3492, opened 2023-09-21
feat: Add ebs_optimized option

pbarnajc:
Fixes: https://github.com/philips-labs/terraform-aws-github-runner/issues/2988

Adds ebs_optimized parameter for ec2 instances.

Currently this defaults to false to keep it in line with the previous behavior.


npalm (approved):
@pbarnajc thx

PR #3494, opened 2023-09-22
fix: missing partion in ARN to support China or govCloud

carloscao0928:
Fixed the problem of AWS not being able to run correctly in China due to partition errors


carloscao0928:
Hi, can you help review it and publish a release for that? @npalm @gertjanmaas @GuptaNavdeep1983 , many thanks

PR #3508, opened 2023-09-26
docs: Document using labels with self-hosted runners

sdarwin:
Explanation of issue: There are many popular CI systems that exist out in the world, not only GitHub Actions. It's probably common enough that a developer comes to this project and needs to configure the Terraform runners with GitHub Actions, while also having in their mind how other systems operate, and those may be different. It's easy to imagine all runner labels must match all job labels. Other systems are designed that way. To quickly solve the problem this is a sort of cheatsheet.

### DIFF
--- a/docs/additional_notes.md
+++ b/docs/additional_notes.md
@@ -1,0 +1,34 @@
+
+# Runner Labels
+
+Some CI systems require that all labels match between a job and a runner. In the case of GitHub Actions, workflows will be assigned to runners which have all the labels requested by the workflow, however it is not necessary the workflow mentions all labels.
+
+Labels specify the capabilities the runners have. The labels in the workflow are the capabilities needed. If the capabilities requested by the workflow are provided by the runners, there is match.  
+
+Examples:
+
+| Runner Labels | Workflow runs-on: | Result |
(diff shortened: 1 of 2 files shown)


docs/additional_notes.md
```diff
@@ -0,0 +1,29 @@
+
+# Runner Labels
+
+Some CI systems require that all labels match between a job and a runner. In the case of GitHub Actions, workflows will be assigned to runners which have all the labels requested by the workflow, however it is not necessary the workflow mentions all labels. This may best be explained by example:
```
npalm, line 4: I think expaling labels as capabilites will help as well. See it as follow the labels specifies the capabilites the runner have, and the labels in the workflow the capabilites need. If the capabailites requrest by the workflow are provided by the runnerw have a match. Can in coroprate this?

npalm (changes requested):
@sdarwin thanks for adding this extra explanation. Maybe good to link the doc from in the readme. Would use one of the first places where the readme explain the labels on a high level. Ensure you link relative

docs/additional_notes.md
sdarwin: Text updated

npalm (approved):

PR #3510, opened 2023-09-28
fix: incorrect path to update app script

npalm:


Jegan-Kunniya (approved):
Approving the change, as it looks good to me.

PR #3513, opened 2023-09-29
feat: Use Instance Metadata Service Version 2 by default.

### LINKED ISSUE #3512: IMDSv2 should be required by default
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html

I'd love to see this option enabled by default for additional security:
```
    http_tokens                 = "required"
```

It looks like we can custom-configure the instance metadata options for runners.

 
```
variable "metadata_options" {
  description = "Metadata options for the ec2 runner instances. By default, the module uses metadata tags for bootstrapping the runner, only disable `instance_metadata_tags` when using custom scripts for starting the runner."
  type        = map(any)
  default = {
    instance_metadata_tags      = "enabled"
    http_endpoint               = "enabled"
    http_tokens                 = "optional"
    http_put_response_hop_limit = 1
  }
}
```

Seems like the description indicates that'd break things without some userdata script rework though.

dylanmtaylor:
This closes #3512.


dylanmtaylor:
We are running this module in production with this set and it works perfectly fine.

modules/runners/variables.tf
```diff
@@ -504,7 +504,7 @@ variable "metadata_options" {
   default = {
     instance_metadata_tags      = "enabled"
     http_endpoint               = "enabled"
-    http_tokens                 = "optional"
+    http_tokens                 = "required"
```
npalm, line 507: Can you also apply this change to:
https://github.com/philips-labs/terraform-aws-github-runner/blob/35aa73ac1e6df30d4a5c314457da5646a117131b/modules/multi-runner/variables.tf#L36

npalm (changes requested):
Thanks, can you apply the cahnge to the mult-runner moduel as well

modules/runners/variables.tf
dylanmtaylor: Yes, of course.

dylanmtaylor:
> Thanks, can you apply the cahnge to the mult-runner moduel as well

@npalm This is ready for re-review.

npalm (approved):
@dylanmtaylor Thanks and sorry for the delay. Changes is checked and all good.

Issue #3515, opened 2023-10-02
Allow multi-runner configs to have per runner config subnet definitions


pbarnajc:
The use case here is if you want some runners on a private subnet and some on a public subnet and you want to control that per runner type.

I'll see if I can get a PR up for this.

PR #3516, opened 2023-10-02
feat: Set up per runner type subnet specification for multi-runner

pbarnajc:
Resolves https://github.com/philips-labs/terraform-aws-github-runner/issues/3515
This just adds the ability to specify a specific subnet_id for a specific runner type for the multi-runner module. There should still always be subnet_ids specified at the module level for the default, but if subnet_ids are specified at the runner level they should override the module level values.

This will enable users to have runners in different subnets without needing to call this module multiple times.


pbarnajc:
Whoops, just saw https://github.com/philips-labs/terraform-aws-github-runner/pull/3467 is already doing this

Issue #3527, opened 2023-10-05
Endless loop on runner spin up: An error occurred (ParameterNotFound) when calling the GetParameter operation


npwolf:
I'm trying to convert my existing setup which is working to the multi-runner setup. I'm using the ubuntu example. The agent is coming up in ec2 on the github event, but the user-data.log shows its going into an endless waiting loop.

I added a set -x to the start runner code to get the additional info here:

```
An error occurred (ParameterNotFound) when calling the GetParameter operation:
+ config=
+ [[ -z '' ]]
+ echo 'Waiting for GH Runner config to become available in AWS SSM'
Waiting for GH Runner config to become available in AWS SSM
+ sleep 1
++ aws ssm get-parameter --name /github-action-runners/gh-ci-multi/linux-x64-ubuntu/runners/tokens/i-08e20e2f0e47f277a --with-decryption --region us-east-2
++ jq -r '.Parameter | .Value'
```

I'm using v4.4.1 of the module.

npwolf:
I see all these values in SSM, but the tokens for the machines are not getting created:
<img width="1173" alt="image" src="https://github.com/philips-labs/terraform-aws-github-runner/assets/2318938/bd332ee9-d1b7-467e-9440-edab5584461a">

imphil:
I'm seeing the same issue, and it seems to be a duplicate of https://github.com/philips-labs/terraform-aws-github-runner/issues/3461

npwolf:
I found my scale up lambda was outputting the same thing as https://github.com/philips-labs/terraform-aws-github-runner/issues/3332 "Resource not accessible by integration". 

Adding `  enable_organization_runners: true` under `runner_config: ` did the trick. Seems like it might not be possible to have repo level runners with this?

PR #3529, opened 2023-10-06
fix: Add enable_organization_runners: true to multi example

npwolf:
I tried repo level runners and ran into #3527 . This may be indicative of a bug at a different level, but at least this allows a working example.


npalm:
Thanks for creating the PR. But the example works. But running with repo level runners requires that the app is allowed to create runners on repo level.

Issue #3528, opened 2023-10-06
Is there a way to disable NAT Gateway and deploy directly to public subnets?


guillaume-michel:
We would like to cut down the cost of NAT Gateway and we don't mind for now, having our runners in subnet with direct internet access.

Is there a way to achieve this please?

imishchuk-carbon:
Hello @guillaume-michel 

We've investigated this question too and as far as I can see, currently there is no option to associate Public IPs with runners.

I've submitted this [PR](https://github.com/philips-labs/terraform-aws-github-runner/pull/3547) to add necessary functionality.

For context:
Running runners in private subnet behind NAT gateways incurs additional cost due to NAT gateway usage. In AWS Cost Explorer this can be found under `UsageTypes -> NatGateway-Bytes`.
In our case that is ~13k $/year.

Placing runners in public subnet with public IP associated in theory makes them accessible from internet, but:
1. By default, security group does not allow any ingress traffic
2. In case of ephemeral runners, they are short lived
3. By default runners don't have any SSH key associated

Based on that, risk likelihood is deemed low.

PR #3547, opened 2023-10-17
feat: Add public IP association to github runner

imishchuk-carbon:
### Description

* Add option to associate public IP with runner (disabled by default)

Fixes [3528](https://github.com/philips-labs/terraform-aws-github-runner/issues/3528)

Suggested changes have been used in our env for over a month and it works as expected.

### Checklists

**Development and testing:**
- [x] All tests related to the changed code pass in development
- [x] Pull request is ready for review


imishchuk-carbon:
Pre-commit tests
image

npalm:
@imishchuk-carbon thx looks good. Tested with ip4 did you also check the change with ip6 / dualstack?

imishchuk-carbon:
Hey @npalm 
No, did not test with IPv6/dualstack. Can do that, if it's required.

npalm:
> Hey @npalm No, did not test with IPv6/dualstack. Can do that, if it's required.

In case you have time would be nice. I think the current module is not taking care of ipv6. So assume it is not working. In case you have time to run a quick check that would be nice. If not please can you update the variable description with only tested with ipv4?

imishchuk-carbon:
Hey @npalm 
Tested with dualstack and it did not work for me, in the end. Issues encountered:
1. Using IPv6/Dualstack makes calls to dualstack endpoint of S3, which [requires](https://docs.aws.amazon.com/AmazonS3/latest/userguide/dual-stack-endpoints.html#dual-stack-endpoints-cli) additional configuration, such as 

```
ipv6=$(TOKEN=`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` \
&& curl -IH "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/ipv6 2>/dev/null | head -n 1 | cut -d$' ' -f2)

if [[ $ipv6 == '200' ]]; then
  aws configure set default.s3.use_dualstack_endpoint true
  aws configure set default.s3.addressing_style virtual
fi
```

2. Connection to `https://pipelinesghubeus22.actions.githubusercontent.com/..` time outs. As far as I see, Github [does not support](https://github.com/orgs/community/discussions/10539) IPv6 for some endpoints. This probably can be worked around by using Egress-Only gateway in AWS, but requires further testing and that is getting out-of-scope for this case.
3. Runner startup is much slower for some reason. While on IPv4-only, it takes less than 2 minutes from instance start to job being picked up (We are using ephemeral runners and execute config/install scripts from scratch). On IPv6/Dualstack it took ~6 minutes to start the runner. Now sure where this slowdown is coming from...

----

That being said, I've renamed variable to explicitly state that it is for `IPv4` and added `Only tested with IPv4` to variable description.

image

npalm (approved):

imishchuk-carbon:
Thanks @npalm

ay0o:
@imishchuk-carbon you can forget about my message. There was some misunderstanding with the AWS documentation, because it led to think that `AssociatePublicIpAddress` could be `true`, `false` or not specified at all. However, that's not true. The property is only added to the request when it's `true`, if it's `false`, it is not, so my point of always forcing the `network_interfaces` block to either `true` or `false` doesn't apply.

PR #3549, opened 2023-10-19
fix: add tags to aws resources

gmcoringa:
Add tags to AWS resources that do not have any tags.


npalm (approved):
@gmcoringa thx

We use tag_all to ensure tags are set.

PR #3564, opened 2023-10-24
fix: restrict runner security group to only ingress

ThomasLachaux:
Runner security group is currently open.

This patch removes the default ingress rule security group to deny everything.

More info on [this documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group#removing-all-ingress-and-egress-rules).


npalm (approved):
@ThomasLachaux I think the security group is not openm now rules are defined. But with your additon potential drift will get detected. Which means added rules (mannually) will be removed.

PR #3572, opened 2023-10-27
fix: allow use of associate_public_ipv4_address from top-level module

skottler:


npalm (approved):
@skottler seems to be missed in https://github.com/philips-labs/terraform-aws-github-runner/pull/3547

skottler:
> @skottler seems to be missed in https://github.com/philips-labs/terraform-aws-github-runner/pull/3547

Yep! Thanks for the merge 😄

Issue #1808, opened 2022-03-01
SSM parameters not cleaned up


samuelb:
Hello

I found today that we have hundreds of AWS System Manager Parameters defined. The lambda is creating those when starting a new runner instance to pass the authentication token and other parameters to the runner agent software.

It's so many, that `aws ssm describe-parameters` runs into rate limits. It's not only about keeping things clean, but also it has functional impact. 

Can a cleanup of unneeded parameters get implemented?

samuelb:
To clean up the parameters, I ran `aws ssm describe-parameters --filter "Key=Name,Values=action-runner-" | jq -r ".Parameters[].Name" | xargs -n 10  aws ssm delete-parameters --names`

tonyc-blastpoint:
Still an issue.

tonyc-blastpoint:
Still an issue.

tonyc-blastpoint:
Still an issue.

tonyc-blastpoint:
Still an issue!

tonyc-blastpoint:
...still an issue.

samuelb:
Still an issue, but a PR is on the way

chasemgray:
@samuelb did your PR get merged to fix this?

samuelb:
I meant PR #2587. I'm not the author of it and no, it didn't got merged till now.

chasemgray:
Is anyone going to reopen and merge the PR? We hit a 10,000 SSM parameter limit due to old parameters not being cleaned up.

PR #3577, opened 2023-10-30
feat: SSM housekeeper

npalm:
# Description
The runner module uses SSM to provide the JIT config or token to the runner. In case the runner does not start healthy the SSM parameter is not deleted. This PR adds a Lambda to remove by default SSM paramaters in the token path that are older then a day. 

The lambda will be deployed by default as part of the control plane and manage the tokens in the path used by the scale-up runner function.


ScottGuymer (dismissed):

lambdas/functions/control-plane/jest.config.ts
```diff
@@ -6,10 +6,10 @@ const config: Config = {
   ...defaultConfig,
   coverageThreshold: {
     global: {
-      statements: 97.6,
-      branches: 94.6,
-      functions: 97,
-      lines: 98,
+      statements: 97.89,
```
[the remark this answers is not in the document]
npalm: I set them just equal to the actuals. This will show a delcine directly. And if needed they can be updated.

npalm:
fix #1808

examples/multi-runner/main.tf
```diff
@@ -68,7 +68,7 @@ module "runners" {
   # override delay of events in seconds
 
   # Enable debug logging for the lambda functions
-  # log_level = "debug"
+  log_level = "debug"
```
GuptaNavdeep1983, line 71: do we need this?

lambdas/functions/control-plane/package.json
```diff
@@ -8,7 +8,7 @@
     "test": "NODE_ENV=test jest",
     "test:watch": "NODE_ENV=test jest --watch",
     "lint": "yarn eslint src",
-    "watch": "ts-node-dev --respawn --exit-child src/local.ts",
+    "watch": "ts-node-dev --respawn --exit-child src/local-ssm-housekeeper.ts",
```
GuptaNavdeep1983, line 11: do you want to keep it like this?

GuptaNavdeep1983 (approved):

lambdas/functions/control-plane/package.json
npalm: does not really matter.

GuptaNavdeep1983 (approved):

PR #3615, opened 2023-11-16
refactor: reduce verbosity when customizing some EBS settings for multi-runner

jpalomaki:
Copied from here: https://github.com/philips-labs/terraform-aws-github-runner/blob/main/variables.tf#L267-L280

Should be backwards compatible as the defaults remain the same.


jpalomaki:
@npalm Looks like checks are failing on an unrelated variable not passing linting

npalm:
> @npalm Looks like checks are failing on an unrelated variable not passing linting

fixed in #3616

npalm:
> @npalm Looks like checks are failing on an unrelated variable not passing linting

can you rebase with main, problem should be resolved

jpalomaki:
@npalm Updated

npalm (approved):

PR #3617, opened 2023-11-17
chore: avoid update-docs workflow runs on pushes to main

npalm:
Docs should be updated on branches, in case of the locked main branch a PR is created. This to ensure the main is up-to-date.


koendelaat (changes requested):
I think there are two options to handle `update-docs` on the `main` branch

1. Don't update docs on the `main` branch
2. Create a PR when docs need to be updated on the `main` branch

Current suggestion seems to implement both options at the same time, which makes no sense to me. So we either have to skip checking on the main branch (as suggested), but then we should also remove the conditional PR creation step.

Other option could be to fix the PR creation step.
Suggestion:

```yaml
      - name: Create Pull Request (main branch only)
        if: github.ref == 'refs/heads/main'
        uses: peter-evans/create-pull-request@153407881ec5c347639a548ade7d8ad1d6740e38 # ratchet:peter-evans/create-pull-request@v5.0.2
        with:
          token: ${{ steps.token.outputs.token || secrets.GITHUB_TOKEN }}
          commit-message: "Update Terraform docs"
          title: "docs: Update Terraform docs"
          branch: update-docs
          base: ${{ github.ref }}
```

npalm:
@koendelaat thx. Updated the workflow. The workflow should now run on any branch in any repo and commit changes. With only an exception for the main branch in this repo. In that case a PR should be opened.

.github/workflows/update-docs.yml
```diff
@@ -16,34 +16,18 @@ jobs:
     name: Auto update terraform docs
     runs-on: ubuntu-latest
     steps:
-      - name: Get installation token
```
npalm, line 19: Removed the app. This is a left over from the time we were making changes to protected branches.

examples/default/variables.tf
```diff
@@ -8,7 +8,7 @@ variable "github_app" {
 }
 
 variable "environment" {
-  description = "Environment name, used as prefix"
+  description = "Environment name, used as prefix."
```
npalm, line 11: Test workflow by making a tiny change.

.github/workflows/update-docs.yml
```diff
@@ -16,34 +16,18 @@ jobs:
...
           git-commit-message: "docs: auto update terraform docs"
-          git-push: ${{ github.ref != 'refs/heads/main' }}
+          git-push: ${{ github.ref != 'refs/heads/main' || github.repository_owner != 'philips-labs' }}
 
       - name: Create Pull Request (main branch only)
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && github.repository_owner == 'philips-labs'
         uses: peter-evans/create-pull-request@153407881ec5c347639a548ade7d8ad1d6740e38 # ratchet:peter-evans/create-pull-request@v5.0.2
         with:
           token: ${{ steps.token.outputs.token || secrets.GITHUB_TOKEN }}
```
koendelaat, line 33: This token doesn't exist anymore

.github/workflows/update-docs.yml
```diff
@@ -16,34 +16,18 @@ jobs:
...
         with:
           find-dir: .
           git-commit-message: "docs: auto update terraform docs"
-          git-push: ${{ github.ref != 'refs/heads/main' }}
+          git-push: ${{ github.ref != 'refs/heads/main' || github.repository_owner != 'philips-labs' }}
 
       - name: Create Pull Request (main branch only)
-        if: github.ref == 'refs/heads/main'
+        if: github.ref == 'refs/heads/main' && github.repository_owner == 'philips-labs'
         uses: peter-evans/create-pull-request@153407881ec5c347639a548ade7d8ad1d6740e38 # ratchet:peter-evans/create-pull-request@v5.0.2
```
koendelaat, line 31: The parameters of this action are incorrect, see my previous review for a suggestion

koendelaat (changes requested):

.github/workflows/update-docs.yml
npalm: which token?

.github/workflows/update-docs.yml
npalm: fixing...

npalm: fiexed and testes

koendelaat (approved):

Issue #3594, opened 2023-11-03
Lambda environment paramaters can exceed 4KB limit


rsavage-nozominetworks:
Greetings-

Apparently the latest version (5.3.0) of `terraform-aws-github-runner` is breaking our lambda updates.   This has never been an issue until now.

```
│ Error: updating Lambda Function (REDACTED-webhook) configuration: operation error Lambda: UpdateFunctionConfiguration, https response error StatusCode: 400, RequestID: 20f1f0c3-3326-4a61-895f-fccf2eb4ff64, InvalidParameterValueException: Lambda was unable to configure your environment variables because the environment variables you have provided exceeded the 4KB limit.
```

Were additional labels added thus causing the 4KB limit to be exceeded?

rsavage-nozominetworks:
Apparently changes to the runner labels have been made which seems to add additional labels to the respectful lambda environment variables. Due to this, it exceeded AWS 4kb limit and throws the error.   We don't have many additional labels and they are fairly short.  I can't image we are the only ones getting this error now.

I have since reverted back to 5.2.1 and the problem is gone.

npalm:
Thanks for reporting. But 5.3.0 should not be breaking. It adds a feature for the multi runners to set a priority to the matcher. See here teh diff. Only terraform code https://github.com/philips-labs/terraform-aws-github-runner/commit/1829721cb647b005ca44b5feb8d0de27217e6074

npalm:
Adding ar ref to an AWS post, total size hsould be less than 4K, https://repost.aws/knowledge-center/lambda-environment-variable-size

Which mean we have to re-thing how we configure the lambda's

npalm:
Can you share your configuration, of course without the sensitive part? So we can reproduce the isse.

npalm:
Quickly had a chat with Amazon, there seems no easy way to pro-active warn on the size.

npalm:
Maybe adding this to the lambda got give some ideas about the current usages.

```typescript
logger.debug('the total size of env ' + Buffer.from(process.env.string).length + ' bytes');
```

npalm:
This script should give some insights in the sizes of the environment variables per function.

```bash
#!/bin/bash

functions=$(aws lambda list-functions --query 'Functions[*].FunctionName' --output text)

for function in $functions
do
  env_vars=$(aws lambda get-function-configuration --function-name $function --query 'Environment.Variables' --output text)
[...]

PR #3618, opened 2023-11-19
refactor: restructure webhook lambda

npalm:
Thie PR prepare for changes to easier migrate reading config from SSM instead of environment (#3594), add option to only accept messages of a defined IP list, and to introduce option to connect runners via EventBridg.

- Validate input and throw validation exceptions if event cannot be accepted
- Structure the code that to allow the webhook to be split in acceptiong an event and distribute to a runner (prepare for EventBridge).
- Remove deprecated jest functions.
- THE PR minimized changed on thest, only small structural things. This to ensure the test still validating the implemention


lambdas/functions/webhook/src/ConfigResolver.ts
npalm, line 1: Reading configiguration in a generic way instead of reading environment variable multiple times and everywhere. Next step could be move some or all configuration to SSM for example

GuptaNavdeep1983 (approved):

npalm:
Tested with default and multi-runner example

PR #3619, opened 2023-11-20
feat(main): Make configuration parameters of type SSM SecureString

alpozcan:
It is understandable to keep non-sensitive configuration information as a plain SSM `String`, however for the purposes of SOC2 compliance in our organisation, we need to make all SSM Parameters encrypted.

This PR changes the type for the below SSM parameters so that they are `SecureString`:
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/cloudwatch_agent_config_runner"`
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/run_as"`
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/agent_mode"`
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/enable_jit_config"`
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/enable_cloudwatch"`
- `"${var.ssm_paths.root}/${var.ssm_paths.config}/token_path"`

### DIFF
--- a/modules/runners/logging.tf
+++ b/modules/runners/logging.tf
@@ -43,7 +43,7 @@ locals {
 resource "aws_ssm_parameter" "cloudwatch_agent_config_runner" {
   count = var.enable_cloudwatch_agent ? 1 : 0
   name  = "${var.ssm_paths.root}/${var.ssm_paths.config}/cloudwatch_agent_config_runner"
-  type  = "String"
+  type  = "SecureString"
   value = var.cloudwatch_config != null ? var.cloudwatch_config : templatefile("${path.module}/templates/cloudwatch_config.json", {
     logfiles = jsonencode(local.logfiles)
   })
--- a/modules/runners/runner-config.tf
+++ b/modules/runners/runner-config.tf
@@ -1,34 +1,34 @@
 resource "aws_ssm_parameter" "runner_config_run_as" {
   name  = "${var.ssm_paths.root}/${var.ssm_paths.config}/run_as"
-  type  = "String"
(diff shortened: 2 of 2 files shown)


npalm:
Thank you for taking the time to create a PR. The SSM paramters you are changing are not considers to be secret or senstive and all. For that reason, we use the plain string. Please can you explain why you think those should be sentive?

alpozcan:
> Thank you for taking the time to create a PR. The SSM paramters you are changing are not considers to be secret or senstive and all. For that reason, we use the plain string. Please can you explain why you think those should be sentive? 

Hi,
I believe I already did in the first paragraph above, also acknowledging that they are not sensitive.

npalm:
> > Thank you for taking the time to create a PR. The SSM paramters you are changing are not considers to be secret or senstive and all. For that reason, we use the plain string. Please can you explain why you think those should be sentive?
> 
> Hi, I believe I already did in the first paragraph above, also acknowledging that they are not sensitive.

Sorry mis-read the description. Did you test your change on a actual deployment?

alpozcan:
@npalm Sorry, didn't actually mean to close this.

I have re-[fork](https://github.com/PropellerAero/terraform-aws-github-runner)ed this in our GH organisation, applied the changes above there and also reverted it to v4.6.0. This'll do for our purposes for now.

It deployed successfully on top our v4.1.0 deployment, and I confirmed just now that fresh runners got launched successfully with this change.

If you'd like to upstream this, I'll spend the time to re-open a PR to say 'thanks' for the module :)

Issue #3556, opened 2023-10-23
The bash script in webhook-github-app ignores the -ws and -we parameters and always attempts to fetch from TF state


rgrama:
https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/webhook-github-app/bin/update-app.sh#L33-L43

```shell
while getopts a:f:k:ws:we:h flag
do
    case "${flag}" in
        a) APP_ID=${OPTARG};;
        f) APP_PRIVATE_KEY_FILE=${OPTARG};;
        k) APP_PRIVATE_KEY_BASE64=${OPTARG};;
        we) WEBHOOK_ENDPOINT=${OPTARG};;
        ws) WEBHOOK_SECRET=${OPTARG};;
        h) usages ;;
    esac
done
```
If you echo $flag in that while loop you'll see that each of the letters are printed individually which means we and ws will not be set.

an alternative to that while loop:
```shell
while [[ "$#" -gt 0 ]]; do
    case "$1" in
        -a) APP_ID="$2"; shift ;;
        -f) APP_PRIVATE_KEY_FILE="$2"; shift ;;
        -k) APP_PRIVATE_KEY_BASE64="$2"; shift ;;
        -ws) WEBHOOK_SECRET="$2"; shift ;;
        -we) WEBHOOK_ENDPOINT="$2"; shift ;;
        -h) usages ;;
        *) echo "Unknown option passed: $1"; exit 1 ;;
    esac
    shift
done
```

npalm:
Thanks for taking time to create an issue, feel free to suggest the change in a PR.

PR #3625, opened 2023-11-20
fix(examples): webhook-github-app multiple character args

Benbentwo:
## What
* Change to `webhook-github-app` to use single character options

## Why

The module fails silently, but bash `getopts` does not support multicharacter options this changes that to use single character options instead.

## References
* fixes: #3556
* https://stackoverflow.com/questions/3975004/how-to-make-a-multi-character-parameter-in-unix-using-getopt
* https://unix.stackexchange.com/questions/684717/how-to-use-getopts-in-bash
* https://stackoverflow.com/questions/402377/using-getopts-to-process-long-and-short-command-line-options


npalm (approved):
@Benbentwo thanks!

PR #3633, opened 2023-11-22
docs: moving docs to pages

npalm:
The documentation in our README have become quite extensive over the years. This PR is moving docs including examples and modules to a documentation site. The top-level readme is shrinked down to some basic / essential information.

- New top-level readme: https://github.com/philips-labs/terraform-aws-github-runner/tree/npalm/mkdocs?tab=readme-ov-file
- Documentation site: https://philips-labs.github.io/terraform-aws-github-runner

For examples and module documentation sources in example and module directories are linked.

PR #3669, opened 2023-12-14
docs: exactMatch variable

sdarwin:
Usage of the setting `exactMatch` is unclear.      

There is an issue entitled "What's the point of exactMatch in multi-runner?" And I keep asking myself this question.    

As an attempt to answer, this PR includes documentation which covers multiple cases of both `exactMatch: true` and `exactMatch: false`.  

- please verify the analysis is correct for each case  
- if possible, provide a situation which demonstrates the intended positive use of `exactMatch: false`.  that can be added to the text.   
- if the conclusion is ultimately that `exactMatch` should be enabled, and is recommended, then we might wonder:  why is the default setting `false`

### DIFF
--- a/docs/additional_notes.md
+++ b/docs/additional_notes.md
@@ -32,3 +32,113 @@ If default labels are removed:
 | 'custom5' | Linux | no match |
 | 'custom5' | [ self-hosted, Linux ] | no match |
 | 'custom5' | [ custom5, self-hosted, Linux ] | no match |
+
+# Exact Match
+
+The multi-runner module has a setting `exactMatch` which affects how the matcherConfig functions.  
+
+The module will decide the runner for the workflow job based on the match in the labels defined in the workflow job and runner configuration. The runner configuration allows the match to be exact or non-exact match. We recommend to use only exact matches. For exact match, all the labels defined in the workflow should be present in the runner configuration matchers and for non-exact match, some of the labels in the workflow, when present in runner configuration, shall be enough for the runner configuration to be used for the job. First the exact matchers are applied, next the non exact ones.
+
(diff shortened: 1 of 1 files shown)


sdarwin:
Maybe there is another solution. 

In the root level module it has a variable:  

```
variable "enable_runner_workflow_job_labels_check_all" {
  description = "If set to true all labels in the workflow job must match the GitHub labels (os, architecture and `self-hosted`). When false if __any__ label matches it will trigger the webhook."
  type        = bool
  default     = true
}
```
That's used to populate `exactMatch`.

What could be done in multi-runner is:

- Change all instances of `exactMatch` to be named `enable_runner_workflow_job_labels_check_all`.  In the module, the examples, javascript source code, everywhere.  No `exactMatch`.  
- Set the default to "true".
- Then, remove `enable_runner_workflow_job_labels_check_all` from some of the multi-runner examples.  When multi-runner has dozens of possible variables, they don't all need to be set in an example.  
- Document the purpose of  `enable_runner_workflow_job_labels_check_all` with a concrete example.  "Imagine this scenario, with these labels, and this workflow. You can see how it would be helpful to disable `enable_runner_workflow_job_labels_check_all` in this particular circumstance."

npalm:
@sdarwin thanks for taking the time to improve the docs. Will check your PR in more detail asap. The labels are always confusing. In the meantime I would really appreciate if you have time to have a look at this PR (https://github.com/philips-labs/terraform-aws-github-runner/pull/3633). This PR moves all docs to a docs site. Feel free to make comments.

sdarwin:
This PR should be reconsidered...  See my previous comment, about a suggested strategy.

PR #3703, opened 2024-01-06
feat: Allow to manage github app secret manually

NadavOps:


NadavOps:
This way we can create the parameter manually, and not need a way to pass the secret through env variable or other secret mechanism to avoid from saving it in the version control system.

NadavOps:
It is a valid option, but I like to avoid environment variables, because then if I try to run it locally or someone else he needs to have this variable set up as well in their system. when things are all defined in the deceleration is better IMO.

npalm:
Sorry for the delay. Please can you update the documentation with this option. I think it is good to add some more details to the [AWS paramater section](https://philips-labs.github.io/terraform-aws-github-runner/configuration/#configuration-considerations).

Please can you format the terraform code?

NadavOps:
@npalm Hey, added explanation

modules/ssm/ssm.tf
```diff
@@ -7,13 +7,19 @@ resource "aws_ssm_parameter" "github_app_id" {
 }
 
 resource "aws_ssm_parameter" "github_app_key_base64" {
+  count  = var.github_app.key_base64 == null ? 0 : 1
```
npalm, line 10: How did you test this by setting the variable explicit to null? In the app object?

npalm: @NadavOps is this PR still relevant for you. How to test this?

PR #3708, opened 2024-01-09
fix: Loosen the AWS Terraform provider constraint

sthristov:
The current AWS provider constraint allows only the rightmost version component to increment. [see docs](https://developer.hashicorp.com/terraform/language/expressions/version-constraints#version-constraint-syntax)
This esentially prevents me from updating the provider to the current latest version (5.31.0).
This change set will allow people to update the minor version of the provider.


npalm (approved):
@sthristov thx

PR #3714, opened 2024-01-10
fix(webhook-github-app): Allow new lines in base64 key for module webhook-github-app

imphil:
We store the `github_app.key_base64` configuration settings with
newlines (nothing unusual, actually). This breaks the
`webhook-github-api` module, which then reports the following when
running `terraform apply`:

```
module.webhook-github-app.null_resource.update_app (local-exec): Executing: ["bash" "-c" ".terraform/modules/webhook-github-app/modules/webhook-github-app/bin/update-app.sh -e https://REDACTED/webhook -s REDACTED -a REDACTED -k FIRSTLINE\nSECONDLINE\nAND_SO_ON==\n"]
module.webhook-github-app.null_resource.update_app (local-exec): Could not read private key from /dev/fd/63
module.webhook-github-app.null_resource.update_app (local-exec): 40975344027F0000:error:1608010C:STORE routines:ossl_store_handle_load_result:unsupported:crypto/store/store_result.c:151:
module.webhook-github-app.null_resource.update_app (local-exec): {"message":"A JSON web token could not be decoded","documentation_url":"https://docs.github.com/rest"}gh: A JSON web token could not be decoded (HTTP 401)
```

The root cause is that `base64 -d` cannot decode the argument passed
through `-k`. Fix that by telling `echo` to leave the newlines in place,
so that `base64 -d` can deal with them. Adding `tr -d` is not necessary
in this case.


npalm (approved):
@imphil thanks, LGTM

PR #3715, opened 2024-01-11
fix(multi-runner): remove renundant default labels

npalm:
The multi runner module is adding on two places the default labels like `self-hosted`, `os` and `arch`. The PR ensures the default labels are only added in [`multi-runner/runners.tf`](https://github.com/philips-labs/terraform-aws-github-runner/blob/4f257f9ff643e1f31e597a15dd9a0a65afbdc605/modules/multi-runner/runners.tf#L48)


koendelaat (approved):
I expect no changes, so good to go!

Issue #3700, opened 2024-01-02
User data exceeds 16kb


razor54:
```╷
│ Error: creating EC2 Launch Template (lt-0d19cc63e67d12cbf) Version: InvalidUserData.Malformed: User data is limited to 16384 bytes.
│ 	status code: 400, request id: d433057d-2bfb-4ce5-a121-26ad9048dbd2
│
│   with module.runners.module.runners.aws_launch_template.runner,
│   on .terraform/modules/runners/modules/runners/main.tf line 64, in resource "aws_launch_template" "runner":
│   64: resource "aws_launch_template" "runner" {
│
╵
```
I used to be able to run ~6.8kb of custom init scripts, but now I get the above error after updating from `5.3.0` to `5.6.0`

```
-rw-r--r-- 1 andregaudencio staff 3.1K Feb 11  2023 post-install.sh
-rw-r--r-- 1 andregaudencio staff 3.7K Jan  2 11:50 user-data.sh
```

whilst the final base64 encoded string appears to be ~44kb, 
```
-rw-r--r-- 1 andregaudencio staff 44K Jan  2 11:47 base64-string
```

razor54:
Also tested on `5.4.2` and got the same result. Reverted to `5.3.0`

npalm:
Please can anyone try the fix in PR #3717

razor54:
That worked for me :)

PR #3717, opened 2024-01-12
fix(runners): zip base64 encoded user data to avoid size limits

npalm:
Schrink the size of the userdata buy zipping the content. 

close: #3700


razor54:
I opened the Issue, and upon testing this branch I validated that it works 👍🏼

razor54 (approved):

GuptaNavdeep1983 (approved):

PR #3719, opened 2024-01-13
docs: multi runner does not use enable_runner_workflow_job_labels_check_all

NadavOps:
…ob_labels_check_all


NadavOps:
@npalm  @sdarwin removed the variable from multi runner as I think the only way to pass the behavior `exactMatch` is through the `multi_runner_config` variable

npalm (approved):

PR #3727, opened 2024-01-17
fix: Re-add support for runners without public IP

kristian-lesko:
Hello,

Per our testing, EC2 runner instances are always created with a public IP after #3547, even when this is not requested. This seems to be due to the conditional inclusion of a network interface in the launch template - if public IPv4 is not requested, there is no interface defined in the template, and instances are launched with a default one that includes a public IP.


kristian-lesko:
My mistake, we had a misconfiguration on our side that caused this not to work. Apologies for the confusion.

Issue #3741, opened 2024-01-26
AL2023 Consistently Failing to boot


jkruse14:
I have been trying to upgrade to the latest major version, specifically `5.6.2` from `4.7.0` (lambda versions were updated too); however, the instances consistently fail to boot when using AL2023. The default AMI is used (as of this writing, `ami-0a3c3a20c09d6f377`)

While debugging, I run `tail -f /var/log/user-data.log` to see what information the script is giving. I consistently see RPM failing, but in different spots along the way, but with the same message when installing dependencies, here is an example:

<img width="315" alt="image" src="https://github.com/philips-labs/terraform-aws-github-runner/assets/16331726/1046d0ab-b462-4fe3-90d8-47999cd20116">

When this message is printed, the boot up stalls out and nothing happens. The instances is eventually brought down by the scale down lambda.

I looked for out of memory errors as described [here](https://repost.aws/knowledge-center/ec2-linux-resource-over-utilization) using the following:

```
$ sudo journalctl -p err | grep -i "out of memory"
```
and
```
$ sudo dmesg | grep -i "out of memory"
```

I did not find any messages relating "to out of memory"

In the journalctl, I did find this error:

<img width="389" alt="image" src="https://github.com/philips-labs/terraform-aws-github-runner/assets/16331726/a741c2e4-c869-4178-8cc8-f8ddbc49243a">

I'm not sure if that message is relevant.

I've also found this issue in amazon-linux-2023's repo that seems relevant:
https://github.com/amazonlinux/amazon-linux-2023/issues/397

Please let me know if I can provide any additional information.

Additional Log Information:
```
<13>Jan 26 19:02:10 user-data: Un-tar action runner
<13>Jan 26 19:02:17 user-data: Delete tar file
<13>Jan 26 19:02:17 user-data: OS: "amzn" "fedora"
[   87.371493] systemd-sysv-generator[20854]: SysV service '/etc/rc.d/init.d/cfn-hup' lacks a native systemd unit file. Automatically generating a unit file for compatibility. Please update package to include a native systemd unit file, in order to make it more safe and robust.
[   87.386192] zram_generator::config[20856]: zram0: system has too much memory (7833MB), limit is 800MB, ignoring.
<13>Jan 26 19:02:18 user-data: Last metadata expiration check: 0:00:31 ago on Fri Jan 26 19:01:47 2024.
<13>Jan 26 19:02:19 user-data: Dependencies resolved.
[...]

PR #3748, opened 2024-02-02
feat: add retry logic to libicu download

jkruse14:
## Background
---
There seems to be a race condition some where with the user-data script and the EC2 starting up and locking RPM. This issue is seen elsewhere, not necessarily with this repo's user-data, see the following:

[Amazon Linux 2023 - issue with installing packages with cloud-init](https://repost.aws/questions/QU_tj7NQl6ReKoG53zzEqYOw/amazon-linux-2023-issue-with-installing-packages-with-cloud-init)

[dnf/yum both fails while being executed on instance bootstrap on Amazon Linux 2023](https://repost.aws/questions/QUgNz4VGCFSC2TYekM-6GiDQ/dnf-yum-both-fails-while-being-executed-on-instance-bootstrap-on-amazon-linux-2023)

Also, #3741

## Changes Made
---
Added a loop to retry if the rpm lock file is found which sleeps for 5 seconds then retries again with a total of 5 iterations.

## Testing
---
Spun up an instance and ran the script directly (using sudo the second time because this was not run via user-data)
image


jkruse14:
@npalm - any chance we could get eyes on this?

npalm (commented):
Thanks, looks good. Need to test asap

jkruse14:
> Thanks, looks good. Need to test asap

@npalm  - Anything you need from me for testing?

modules/runners/templates/install-runner.sh
```diff
@@ -47,7 +47,20 @@ echo OS: $os_id
 
 # Install libicu on non-ubuntu
 if [[ ! "$os_id" =~ ^ubuntu.* ]]; then
-  dnf install -y libicu
+  max_attempts=5
+  attempt_count=1
```
npalm, line 51:
```suggestion
  attempt_count=0
```

npalm (changes requested):
Looks good, a small suggestion

npalm (approved):

Issue #3736, opened 2024-01-24
Windows: `User data format: unrecognized`


guicaulada:
When using version 5.6.3 I noticed Windows instances were not executing the User Data script and `C:\UserData.log` was not created when using SSM to log into the machine.

I believe this was caused by: https://github.com/philips-labs/terraform-aws-github-runner/pull/3717

That PR fixed another issue that I had with my Linux runners where my user data would endup too big, however it broke the Windows runners when they stopped executing the user data script.

Looking at `C:\ProgramData\Amazon\EC2Launch\log\agent.log`.

When using 5.6.3 I get the following log message:
```
2024-01-24 16:54:51 Info: Getting user data
2024-01-24 16:54:51 Info: Try parsing user data in YAML format.
2024-01-24 16:54:51 Info: Parsing failed, fall back to XML format.
2024-01-24 16:54:51 Warning: No XML or YAML user data detected. If XML or YAML user data is intended, check for typos or refer to documentation (https://docs
.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2launch-v2-settings.html#ec2launch-v2-example-user-data): %!w(*fmt.wrapError=&{Failed to unmarshal user data as
YAML: yaml: control characters are not allowed 0xc000115e70})
2024-01-24 16:54:51 Console: User data format: unrecognized
2024-01-24 16:54:51 Console: EC2LaunchTelemetry: IsTelemetryEnabled=true
2024-01-24 16:54:51 Console: EC2LaunchTelemetry: AgentOsArch=windows_amd64
2024-01-24 16:54:51 Console: EC2LaunchTelemetry: IsAgentScheduledPerBoot=true
2024-01-24 16:54:51 Console: EC2LaunchTelemetry: AgentCommandErrorCode=0
2024-01-24 16:54:51 Console: Message: Windows is Ready to use
2024-01-24 16:54:51 Info: Initialize user-data state.
2024-01-24 16:54:51 Info: User-data state initialized successfully.
```

When using 5.6.2:
```
2024-01-24 16:35:35 Info: Getting user data
2024-01-24 16:35:35 Info: Try parsing user data in YAML format.
2024-01-24 16:35:35 Info: Parsing failed, fall back to XML format.
2024-01-24 16:35:35 Info: Converting user data to YAML format.
2024-01-24 16:35:35 Info: Frequency is set to: once
2024-01-24 16:35:35 Info: Run as user is set to: 'admin'.
[...]

PR #3759, opened 2024-02-12
fix: windows userdata does not support gzip

npalm:
Windows user data is not support gzip. This PR avoids window user_data scripts are gzipped.

close: #3736


Jegan-Kunniya (approved):
Approving the change, as it looks good to me.

GuptaNavdeep1983 (approved):

PR #3768, opened 2024-02-13
feat: Have sudo run the user's shell when calling config.sh from start-runner.sh

bshelton229:
We ran into a really interesting issue where we want the `$PATH` from the `$run_as` user's shell. Mostly things that are configured in the bash rc files, like `~/.local/bin`, which aren't present in the base system path. I couldn't find a better way to do this. It appears the `.path` file is written by the `config.sh` script, and is then locked into place. I think that happens here - https://github.com/actions/runner/blob/72559572f64f40554d43cfa04e4128725dc2274b/src/Misc/layoutroot/env.sh#L37. We're currently running from AMIs built with this patch.

The `-i` flag, according to sudo, will `Run the shell specified by the target user's password database entry as a login shell. ...`. Because it enters the shell, it resets PWD back to the home folder of the user, hence setting the full path of the `config.sh` script when calling it.

I can't think of any negative consequences to this, but it's definitely likely to cause runners to have a different path than they do now.


npalm:
I also don't see any negative side effects. However I want to run some tests. Can you share some details about your setup?

- Are you using the the standard Amazon AMI?
- Are you setting `run_as`?

npalm:
@bshelton229 please can you check my comment? Also not sure if there is really no impact.

npalm:
@bshelton229 would you have to check my comments? And share some details how to test?

bshelton229:
I don't know how, but I completely lost track of the notifications for this PR. We found another way around this. I'm hoping to open another PR shortly. Thanks!

PR #3769, opened 2024-02-14
chore(ci): drop lerna for nx

npalm:
Replacing Lerna by NX
- allow to cache parts of the build (local runs)


nadavsinai-philips:
@npalm the NCC package you're using for bundling/compiling is not a happy camper with NX, see discussion in https://github.com/nrwl/nx/issues/5005. I would recommend to use SWC for which an executor is included in @nx/js

npalm:
@nadavsinai-philips thx, your changes are now in this branch.

ScottGuymer (dismissed):
:shipit:

koendelaat (approved):

PR #3790, opened 2024-03-03
fix: make variables more consistent

### LINKED ISSUE #3740: Multi-Runner Input Names are Different than Normal Runner Inputs
When I migrated a setup using the normal runner module to the multi-runner module, I had some trouble getting it working because some of the input names differ. It would be helpful is we rename the multi-runner inputs so they are consistent. Here are some examples:

- `enable_fifo_build_queue` to `fifo`
- `enable_runner_on_demand_failover_for_errors` to `enable_on_demand_failover_for_errors`

There may be more, but I didn't check yet. Happy to open a PR to rename these, but want to make sure there isn't a reason why they're supposed to be different.

rvodden:
I've written this as a non-breaking change. The new variable names are preferred over the old if both are set. Intended as a fix for #3740


npalm:
CI is still not happy

rvodden:
Yeah I'll have a look this weekend!

PR #3798, opened 2024-03-05
feat: allow caller to provide custom userdata

ebirger:
Using a prebacked AMI we wish to provide a userdata script which requires information from terraform resources - e.g. endpoints for that environment.

The existing `userdata_template` is not sufficient for this purpose as it requires a static file.

It would be useful to be able to render the userdata script before providing it to the runner configuration.


npalm:
Can you provide a usage example?

ebirger:
> Can you provide a usage example?

Sure - but where should I provide it? I'd rather not change an existing working example for this, and a whole new example seems too much...

npalm:
The idea is to replace the full template by a static filled content for the user_data? Indeed I think it is better to keep the examples in tact. 

Maybe adding a small section to the configuration with some basic direction how to use this. We have embedded the start script in the AMI. WHich is antother approach. It should be clear to the user that the userdata_content is responsible for registring the runner based on the parameters created by the lambda.

npalm:
> > Maybe adding a small section to the configuration with some basic direction how to use this. We have embedded the start script in the AMI. WHich is antother approach. It should be clear to the user that the userdata_content is responsible for registring the runner based on the parameters created by the lambda.
> 
> The proposed wording includes:
> 
> `By providing your own user_data you have to take care of installing all required software, including the action runner`
> 
> taken from the original "template" version.
> 
> Is there any more information needed?

Could you make one small improvement to the current text. 

By providing your own user_data you have to take care of installing all required software, including the action runner and registering the runner. Be-aware configuration paramaters in SSM as well tags are treated as internals. Changes will not trigger a breaking release.

ebirger:
Thank you!
Updated per requested, and rebased on top of current main.

PR #3830, opened 2024-03-28
fix(lambda): repair repository allow list

y3ti:
I noticed that the environment variable was previously renamed from `REPOSITORY_WHITE_LIST` to `REPOSITORY_ALLOW_LIST`. However, this change was only applied in the `webhook` lambda, while Terraform continues to use the old name.

This pull request corrects the repository allow list functionality without introducing any breaking changes. It's unclear how deprecation and breaking changes are typically handled, so this approach avoids any potential disruptions.


y3ti:
Any chance to review my pull request?
cc: @npalm

y3ti:
@npalm I don't have any real usecase. Just tested this feature and found the bug :) In my case I set the webhook only for a single github repo.

Issue #3737, opened 2024-01-24
Windows: Invalid configuration provided for url


guicaulada:
On version 5.6.2, when using Windows runners with the default user data and start runner script the runner is not running any jobs, when investigating I noticed the `UserData.log` shows:
```
Configure GH Runner as user ec2-user
.\config.cmd : Invalid configuration provided for url. Terminating unattended configuration.
At line:1 char:1
+ .\config.cmd --unattended --name windows-x64_i-0c96a1237c96979a5 --wo ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Invalid configu... configuration.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

The runner shows up on GitHub, but it says it's offline.

guicaulada:
Apparently Windows doesn't support JIT config?

Setting `enable_jit_config: false` solved the issue for me.

npalm:
Windows should support JIT as well. But I am not sure if the open source start script support this. @GuptaNavdeep1983 Can you make a comment?

guicaulada:
I believer either the start script is not sending the correct arguments to `config.cmd` or the issue is on `config.cmd` itself and this could be moved to https://github.com/actions/runner

GuptaNavdeep1983:
@guicaulada, JIT is supported on Windows as well. Can you please create a PR to update the default start script to support this?

guicaulada:
@GuptaNavdeep1983 sorry, but I'm not sure what needs to be changed on the default start script for JIT to work correctly, it seems the arguments are passed correctly the same way it would on Linux but results in: `.\config.cmd : Invalid configuration provided for url. Terminating unattended configuration.`

shubhamsinha-sf:
facing the same issue here. The url seems to be empty in our case:

```

[2024-03-21 15:06:59Z INFO CommandSettings] Arg 'url': ''

```

rfsmart-bcleary:
Here's the help I found on the config.cmd:
```shell
.\config.cmd --help

Commands:
 .\config.cmd         Configures the runner
 .\config.cmd remove  Unconfigures the runner
 .\run.cmd            Runs the runner interactively. Does not require any options.

Options:
 --help     Prints the help for each command
 --version  Prints the runner version
[...]

PR #3842, opened 2024-04-09
fix: Adding github URL to config.cmd

rfsmart-bcleary:
#3737 This adds the variable gh_url, creates a SSM parameter containing the value from gh_url, and passes in the value of gh_url to the config.cmd call.


dhlaurap:
Hello, I am running into this issue, when could this be available in a release?

dhlaurap:
I was just looking at this more closely and since multiple repos can use the runner would it be better if the webhook lambda stored the repo URL (from the webhook payload) into SSM for the runner to pick-up when running the user_data script?

jonanderson10:
+1 I am seeing this issue as well

dhlaurap:
I actually ended up fixing this by using a custom image and modifying the start-runner script by removing the `.\config.cmd` call and adding the `--jitconfig` argument to the run.cmd action like this:

`$action = New-ScheduledTaskAction -WorkingDirectory "$pwd" -Execute "run.cmd" -Argument "--jitconfig $config"`

From what I could gather, a similar fix went into the linux start-runner script but the Windows one was never fixed.

ardole:
@dhlaurap can you confirm which version of the github runner are you using please ?

- i face the same issue with `--token`, to me its https://github.com/actions/runner/issues/1882 

- your solution with `jitconfig` seems to not work with [v2.317.0](https://github.com/actions/runner/releases/tag/v2.317.0) argument is not recognized and nothing better with argument `ephemeral`. i'm wondering why the `start-runner.ps1` and `start-runner.sh` are different because for windows platforms `enable_ephemeral_runners` and `enable_jit_config` are simply ignored, runner is always starting as a service.

this project is great but seems to suffer of a lack of testability (or i haven't found it), can somebody give tricks on how to quickly test the AMI without need to deploy the whole lambdas ? for instance `start-runner.ps1` depends on a **ssm-parameters** and registration done in lambdas.

ardole:
You are right, something was wrong with the `$config` that i generated manually to debug.
After new try i was able to make it working with the argument you suggested.

I opened a new issue https://github.com/philips-labs/terraform-aws-github-runner/issues/3950 and suggest a simple fix https://github.com/philips-labs/terraform-aws-github-runner/pull/3951

PR #3855, opened 2024-04-20
feat: adjust pool dynamically based on demand

galargh:
#### Description

This PR extends the `pool` lambda with:
- ability to adjust the pool based on the number of queued jobs
- support for Repo level runners
- support for multiple runner owners

###### Dynamic Pool

Sometimes, a runner startup fails, for example, due to a connectivity issue with GitHub servers. When using the self-hosted runners in an ephemeral mode, such failures can lead to starvation. By adding support for dynamic runner scaling based on the number of currently queued jobs to the pool adjust lambda, it becomes possible to counteract this issue.

###### Repo Level Runners

Up until now, the pool adjustment supported only Org-level runners. Adding support for Repo-level runners means that anyone using self-hosted runners in this manner will now be able to use the pool lambda as intended.

###### Multiple Runner Owners

By accepting multiple runners as input to the pool lambda, the overall number of lambdas created for the setup can be reduced. This might be desirable when someone is running Repo-level runners for many repositories.

###### Dynamic Pool

The use of this feature can be controlled via the `pool_size` input of the pool lambda. Setting it to `-1` will enable the dynamic pool. When the dynamic pool is enabled, the function determines the desired pool size by counting the number of currently queued jobs in the context of the runner owner.

###### Repo Level Runners

From now on, the pool lambda will accept `runner_owner` input in the form of `OWNER/REPO`. When the runner owner is a repository, the function will create repository-level runners instead of org-level ones. It will also check for idle runners in the repository context. It will not check for idle runners in the context of the organization.

###### Multiple Runner Owners

The pool lambda will expect the `runner_owner` input to be a comma-separated list of owners and process pool adjustment for each.

#### Testing

I have deployed the `runners.zip` and configured the dynamic pool in the setup I use for https://github.com/libp2p and https://github.com/ipfs.

I added a set of unit tests for the newly added functionality.


lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -18,6 +22,13 @@ interface RunnerStatus {
   status: string;
 }
 
+function canRunJob(workflowJobLabels: string[], runnerLabels: string[]): boolean {
```
npalm, line 25: Assume this code, is copied from webhook? Correct? In that case we should move this to a common module. Can be done later in a new PR. In that case can you create a issue, and link it here in the code?

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -36,7 +47,7 @@ export async function adjust(event: PoolEvent): Promise<void> {
   const launchTemplateName = process.env.LAUNCH_TEMPLATE_NAME;
   const instanceMaxSpotPrice = process.env.INSTANCE_MAX_SPOT_PRICE;
   const instanceAllocationStrategy = process.env.INSTANCE_ALLOCATION_STRATEGY || 'lowest-price'; // same as AWS default
-  const runnerOwner = process.env.RUNNER_OWNER;
+  const runnerOwners = process.env.RUNNER_OWNER.split(',');
```
npalm, line 50: Assume you propose to use a list split with comma for muliple owners, which can be a repo as well.

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -48,63 +59,123 @@ export async function adjust(event: PoolEvent): Promise<void> {
...
+
+    const numberOfRunnersInPool = calculatePooSize(ec2runners, runnerStatusses);
+    let topUp = 0;
+    if (event.poolSize >= 0) {
+      topUp = event.poolSize - numberOfRunnersInPool;
+    } else if (event.poolSize === -1) {
+      logger.info('Checking for queued jobs to determine pool size');
+      let repos;
+      if (runnerType === 'Repo') {
+        repos = [repo];
```
npalm, line 96: Would it not be easier to maintain to fetch th lis tof repo's from the app. Like you do for the org. In that case this if is not needed. And the pool can adjust regardless knowing if it is installed in an org or repo.

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -48,63 +59,123 @@ export async function adjust(event: PoolEvent): Promise<void> {
...
+      } else {
+        // @ts-expect-error The types normalized by paginate are not correct,
```
npalm, line 98: This way to dynamically scaling could work for smaller deployments. Since it will check all repo's and for each repo the potentially queued jobs. In case you have for example 1000 repo's. It means 1000 calls to check for jobs. Which mean the app will rate limit quickly.

GitHub does not provide a way to request queued jobs, so via the API there is no good alternative. But cause a rate limit will cause the whole control plane will not scale anymore till the rate limit is reset.

The only alternative I could think of is hooing on the events and use the events to  press less hard on the API. 

I am fine with this approach. But there should be some very clear warning in the documenntation. I also would opt we add a flag to enable dynamic scaling and mark it clearly experimental.

npalm (changes requested):

lambdas/functions/control-plane/src/pool/pool.ts
galargh: Yes, that's correct! Unfortunately, I couldn't find a more efficient way to implement this using the API alone. As you said, it could be possible to use webhook events to try to limit the search space if, instead of checking all the repositories, we checked only those that had some jobs scheduled "recently" that weren't scheduled yet. 

One way to implement it could be with an extra SQS queue where we would store workflow job queued events. The queue could have a delay configured. There could also be a lambda that consumes events from that queue. It would check whether a job the event is for has been scheduled already or not. If it has, that's it. If it hasn't, it would put the event back on that queue and on the queue that the scale up lambda consumes. Would something like that make sense to you?

> I am fine with this approach. But there should be some very clear warning in the documenntation. I also would opt we add a flag to enable dynamic scaling and mark it clearly experimental.

Sure, that makes sense! I'll add a new config option for enabling the "dynamic" mode explicitly instead of controlling it via setting `poolSize` to `-1`. I'll make sure to add a warning about this in the documentation too.

galargh (commented):
Thank you for the review, really appreciated 🙇 I'm aiming to apply the requested changes next week.

lambdas/functions/control-plane/src/pool/pool.ts
npalm: Both fine, in case you keep it. Please create an issue and refer in a comment ot the issue.

npalm:
@galargh Still a nice PR, but some time ago I have implemented a job retry mechanism, which used another queue to rety jobs. Is that also solvering your issue here? We use the job retry on our smaller fleets where we have not or very small pools. I leave teh PR open, sicne I see some improvements on the unit test that I would like bring back.

lambdas/functions/control-plane/src/pool/pool.ts
galargh: I added a comment about the space for deduplication for now. If we end up going through with this PR, I'll create an issue and add a link to it to the comment as well.

https://github.com/github-aws-runners/terraform-aws-github-runner/pull/3855/files#diff-57ecac144d87f7e8a6b12ecd7148c2dad084177cf9b28b665cb7e65e49acc13cR25-R26

lambdas/functions/control-plane/src/pool/pool.ts
galargh: I renamed the env var to `RUNNER_OWNERS` to better match the expected value here.

https://github.com/github-aws-runners/terraform-aws-github-runner/pull/3855/files#diff-57ecac144d87f7e8a6b12ecd7148c2dad084177cf9b28b665cb7e65e49acc13cR52

lambdas/functions/control-plane/src/pool/pool.ts
galargh: I added an explicit `dynamicPoolScalingEnabled` option - https://github.com/github-aws-runners/terraform-aws-github-runner/pull/3855/files#diff-57ecac144d87f7e8a6b12ecd7148c2dad084177cf9b28b665cb7e65e49acc13cR17

I also added explicit `EXPERIMENTAL` labels and potential rate limit issue warnings to the docs where applicable - e.g. https://github.com/github-aws-runners/terraform-aws-github-runner/pull/3855/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180R339

galargh:
> @galargh Still a nice PR, but some time ago I have implemented a job retry mechanism, which used another queue to rety jobs. Is that also solvering your issue here? We use the job retry on our smaller fleets where we have not or very small pools. I leave teh PR open, sicne I see some improvements on the unit test that I would like bring back.

Hey! Sorry, I didn't get back to you for so long. It's been quite an eventful end-of-year for me. The job retry mechanism is great! I'm testing it out right now in a couple of orgs where I run self-hosted runners fleets, and I haven't noticed any issues yet. So it definitely looks promising!

One scenario in which the dynamic pool scaling could still come in handy is if GitHub was experiencing issues and some job scheduling events weren't delivered correctly. In such a case, a pool scaled dynamically would kick in and bring up however many machines are needed (up to the set cap) to clear the "zombie" jobs.

Thank you again for the great feedback and developing this amazing project 🙇

PR #3866, opened 2024-04-25
fix: change log level for termination watcher if no match

npalm:
Changed log level to debug in case no match. Deploying the lambda multiple times in a single aws account will create unrelated warning logs otherwise.


Brend-Smits (approved):
LGTM!

PR #3873, opened 2024-04-29
fix: adding missing permissions to boundaries

gnawhleinad:
`setup-iam-permissions`: add missing `iam:TagInstanceProfile` required for `aws_iam_instance_profile "runner"`

The `aws_iam_instance_profile` resource in https://github.com/philips-labs/terraform-aws-github-runner/blob/74471de59ac97748581612efa62c3950344916f4/modules/runners/policies-runner.tf#L11-L16 requires the `iam:TagInstanceProfile` permission.


npalm (approved):
Thx!

PR #3909, opened 2024-05-13
fix(lambda): Prevent scale-up lambda from starting runner for user repo if org level runners is enabled

PerGon:
We run a setup where we only have "Org Level Runners" (`enable_organization_runners: true`).

When creating an action job in a repository that does not belong to an Organization (User level repository), that ScaleUp lambda will happily create a new VM for it.
Well, this new VM in turn, will never be able to register itself in GitHub because it cannot register itself as a "Org Level Runner" as such concept does not exist at the user level.
This VM will then become a "zombie" VM and will also cause issues with the ScaleDown lambda (described in `ToBeCreatedIssue`).

In this PR, we will pass the "Repo Owner Type" all the way from the Webhook lambda to the ScaleUp lambda (by adding a new field in the SQS payload). Then, the ScaleUp lambda can decide to drop the request if the "Repo Owner Type" is not of the type `Organization` and `enable_organization_runners` is set to `true`.


lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
```diff
@@ -305,6 +316,11 @@ describe('scaleUp with GHES', () => {
       expect(mockOctokit.paginate).toHaveBeenCalledTimes(1);
     });
 
+    it('Throws error if it is a User repo and org level runners is enabled', () => {
+      process.env.ENABLE_ORGANIZATION_RUNNERS = 'true';
+      expect(() => scaleUpModule.scaleUp('aws:sqs', TEST_DATA_USER_REPO)).rejects.toBeInstanceOf(Error);
```
npalm, line 321: Would prefer not to add a global event (module scope) for only a single test. Please can you update the event in the test to set the repoUserType

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -341,6 +345,17 @@ async function createStartRunnerConfig(
   }
 }
 
+function validateRepoOwnerTypeIfOrgLevelEnabled(payload: ActionRequestMessage, enableOrgLevel: boolean) {
+  if (enableOrgLevel && payload.repoOwnerType !== 'Organization') {
+    logger.warn(`Repository ${payload.repositoryOwner}/${payload.repositoryName} does not belong to a GitHub` +
+      `organization and organization runners are enabled. This is not supported. Not scaling up for this event.`);
+    throw Error(
```
npalm, line 352: Please do not throw an error, this will send the message back to SQS. And trigger a retry.

npalm (changes requested):
Thanks for taking the time crating the PR. What I don't understand why the user event is sent to the webook. The common way of using the module is installing repositories to an app. 

For me it is not 100% clear what you mean with user space repo's. Are that forks of org repo's?

PerGon:
Howdy! Thanks for having a look.

Sorry I was not very clear 😅 I'll try to elaborate!

Let's consider the following scenario with 2 different repos:

1. `github.my_enterprise.com/some_team/app1` - In this case we have a repo called `app1` in one GitHub Org called `some_team`
2. `github.my_enterprise.com/john_doe/some_poc` - In this case we have a repo called `some_poc` in a GitHub User called `john_doe`

In our setup we use a GitHub App. We also have plenty of GitHub Orgs in our enterprise - we tell the users who want to use our centrally maintained GHA Runners to install the specific GitHub App in their GitHub Orgs so they can use GHA (we have other CIs in our company 🙄 ) - in our internal documentation we clearly state "Do not install the GitHub App in your GitHub User account - ONLY install in GitHub Organizations".
But, even tho this is clearly stated in our internal docs, some engineers will eventually install the GitHub App for GHA runners in their GitHub User account - and once they try to create Action in their "personal" repos, it will cause the erratic behavior I described in OP.

As far as we can see, there's no way to prevent a GitHub App from being installed in GitHub User accounts 😢 

The TL:DR; is that we have a self-service model where engineers can onboard their GitHub Orgs to use GHA Runners, but if they install the GitHub App in their own account (instead of their org) to try our GHA, it will cause issues.

Here's an example of two GitHub Apps installations, one that is OK and one that is not OK.

### This is an OK use case

image


### This is an NOT an OK use case

image


Sorry for the wall of text 😅 hopefully I could exemplify the case that causes issues. Let me know if it made things a bit more clear 😸 and also thanks for the comments in the code - will address them as soon as I have a few minutes!

npalm:
I was out of the office for some time, back. Will check the PR next week. Sorry for the delay.

npalm:
Sorry the PR complete was out of my sight. Will catch up again. Sorry once again

npalm:
Please can you run in the meantime the command `yarn run format` from the `lambdas` dir to address the CI issue about formatting.

npalm (approved):
Thanks for the great contribution. All looks good!

PR #3931, opened 2024-06-10
fix(lambda): reuse http clients

samchungy:
Creating Http Clients per request is slow, as the client always has to re-establish a new connection with AWS. Not to mention, they're actually cause for concern when it comes to memory leaks, but fortunately with Lambdas they aren't long lived enough.

This PR changes the code to create single clients so we should see noticeable improvements with successive invocations of the Lambdas.

It's [best practice](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html#function-code) to initialise these outside of the handler.


lambdas/libs/aws-ssm-util/src/index.ts
```diff
@@ -1,8 +1,10 @@
 import { GetParameterCommand, PutParameterCommand, SSMClient, Tag } from '@aws-sdk/client-ssm';
 import { getTracedAWSV3Client } from '@terraform-aws-github-runner/aws-powertools-util';
 
+let ssmClient: SSMClient;
```
samchungy, line 4: As this is a lib, I don't want to create a new client on import, so we're caching a client only when the below functions are called;

samchungy: Alternatively, the caller should pass in a client

npalm:
Thx for creating the PR. I thought that caching was not needed anymore with v3. I will check your PR next week.

samchungy:
> Have you seen that the canges made are breaking the tests?

Interesting, I'll take a peek when I get some time 👀

PR #3934, opened 2024-06-12
feat(lambda): add option to define explicit lambda tags

wadherv:
Introduced a new variable, lambda_tags, to both the main and submodule. This variable will include any additional lambda function-specific tags, enhancing the governance part of resources.


npalm (approved):
Sorry for the late response, due to leave all PR's are delayed. The PR looks good.

PR #3940, opened 2024-06-18
chore: Execute var.custom_shell_commands in a separate shell provisioner in Packer

### LINKED ISSUE #3939: ec2-user is not seen as a member of the docker group when var.custom_shell_commands is executed
We log into Dockerhub and ECR in `var.custom_shell_commands` during AMI build to pull commonly used base images and pre-warm the Docker cache.

In `var.custom_shell_commands`, the ec2-user does not yet have access to Docker because it was added to the docker group in the same shell. This results in permission errors, such as:
```sh
permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock:
```

This forces us to use sudo when logging in and pulling images and we end up doing things like:

```sh
echo "Login to Dockerhub"
sudo aws ssm get-parameter --name "/github-action-runners/ops/dockerhub_token" --with-decryption --query "Parameter.Value" --output text | sudo docker login --username "$(sudo aws ssm get-parameter --name "/github-action-runners/ops/dockerhub_username" --with-decryption --query "Parameter.Value" --output text)" --password-stdin

echo "Login to ECR"
sudo aws ecr get-login-password | sudo docker login --username AWS --password-stdin ************.dkr.ecr.eu-west-1.amazonaws.com

echo "Move docker config.json to /home/ec2-user/.docker"
sudo mkdir -p /home/ec2-user/.docker
sudo mv /root/.docker/config.json /home/ec2-user/.docker/config.json
sudo chown ec2-user:ec2-user /home/ec2-user/.docker -R
sudo chmod g+rwx "/home/ec2-user/.docker" -R
```

There are workarounds with subshells and other methods, but they quickly become messy.

**Proposed Solution**

If `var.custom_shell_commands` is executed in a separate shell provisioner instead in the Packer templates, logging in and pulling can be done as ec2-user, and there is no need to move .docker/config.json around.

espizo:
This commit fixes #3939


npalm (changes requested):
Please can you check the CI job. The validation is failing with the error no script provided. It seems a default empy array is now allowed. Maybe adding a default with `echo no custom script set` could do the trick.

stuartp44:
@espizo nudge for you.

PR #3942, opened 2024-06-20
fix(webhook): Don't log warning when secondary job queue is empty

iainlane:
Right now the Terraform module is causing `${SQS_WORKFLOW_JOB_QUEUE}` to be set to an empty string. Since we pass this through currently, and explicitly check for `!== undefined` - not any falsy value - we end up trying to send to an empty queue and logging a warning.

Doesn't break anything, but it's noisy in the logs.

Fix this by checking for any falsy value instead, and also using `||` instead of `??` when setting the variable in the first place, so an empty string ends up as `undefined`. Also, modify the testsuite to check for the `SQS` being created at all, since that happens earlier on and reproduces the failure.

This is a companion to #3943. This one stops the warning, and that one fixes the original cause by not setting the env var in the first place. Both could be merged, ideally.


npalm (approved):
Thanks looks all good

PR #3943, opened 2024-06-20
fix(webhook): Don't set `${SQS_WORKFLOW_JOB_QUEUE}` to empty string

iainlane:
We currently indicate that this feature is disabled by setting the env var to an empty string. But instead we could omit it, to avoid polluting the environment, and causing a warning to be logged.

This is a companion to #3942. That one stops the warning mentioned, this one fixes the original cause by not setting the env var in the first place. Both could be merged, ideally.

### DIFF
--- a/modules/webhook/webhook.tf
+++ b/modules/webhook/webhook.tf
@@ -23,15 +23,17 @@ resource "aws_lambda_function" "webhook" {
 
   environment {
     variables = {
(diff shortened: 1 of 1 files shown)


npalm (approved):
Looks good to me, we should really improve the naming. This SWS_WORKFLOW_.... name is confusing.

Issue #3950, opened 2024-06-26
Ephemeral runners based on windows-core-2022 start but never connect to github due to missing argument on run.cmd


ardole:
Base image example windows core 2022 always install the runner as a service.

```
$action = New-ScheduledTaskAction -WorkingDirectory "$pwd" -Execute "run.cmd"
```

This fail when `enable_jit_config` or `enable_ephemeral_runners` is true, because of missing `-Argument "--jitconfig $config"`

Without this, runner will never be connected to github.

Context:
- Github cloud
- Organization runner
- Github runner v2.317.0

PR #3951, opened 2024-06-26
fix: adding missing argument --jitconfig for ephemeral windows runner (#3950)

ardole:
Update of `start-runner.ps1` to make it able to connect windows ephemeral runners


npalm (changes requested):
The change looks good. But when testing the instance is not getting ready. I have added `enable_ephemeral_runners: true` to the configuration in the examples for multi-runners (`examples/multi-runners/templates/`. 

Without ephemeral instances (no jit) instances are getting started and registered. But once setting ephemeral, which by default using JIT the instance is not getting ready.

Can you share you test details, or adjust the multi-runner exampl to test?

ardole:
> The change looks good. But when testing the instance is not getting ready. I have added `enable_ephemeral_runners: true` to the configuration in the examples for multi-runners (`examples/multi-runners/templates/`.
> 
> Without ephemeral instances (no jit) instances are getting started and registered. But once setting ephemeral, which by default using JIT the instance is not getting ready.
> 
> Can you share you test details, or adjust the multi-runner exampl to test?

npalm:
@ardole would you have time to run a quick cross check on PR #4164

PR #3953, opened 2024-06-28
feat: add support for matcher config tiering options

Brend-Smits:
This feature is useful as a workaround when you have a larger number of runner configurations and run into problems where the parameter store limit is being reached (4KB). The limit of Standard is 4KB and Advanced is 8KB.

⚠️ **It should be noted that if you changed the parameter store tier for the matcher config manually, this change might override your changes!** 

Signed-off-by: Brend Smits <brend.smits@philips.com>


npalm (approved):
Thanks for providing the fix. This will fix indeed for the problem for some time. We have to start thinking where to store this configuration on the long run.

PR #3962, opened 2024-07-02
fix: improve logging when there is an error retrieving ssm parameters

Brend-Smits:
In case the AMI ID Parameter was set to something that did not exist, it would produce an obscure error: "message": "Ignoring error: UnknownError".
Now it produces better error messages:
"message": "Ignoring error: Error: Failed to lookup runner AMI ID from SSM parameter: /github-action-runners/staging/ubuntu-x64/runners/config/ami-id,\n       ParameterNotFound: UnknownError",

Signed-off-by: Brend Smits <brend.smits@philips.com>


npalm (approved):
Thx, looks good

PR #3997, opened 2024-07-17
refactor: scale down tests

npalm:
The test structure and magic for scaling down makes it hard to make any change. This PR refactor the tests cases in suche way

- only global default mocking
- per test explicit mocking
- per test setup test data, avoiding global data mesh

Test coverage is 100% for scale-down.

Note: Codescene warnings are not new, but a bit the fact of how jest works


stuartp44 (approved):
LGTM

stuartp44 (approved):

ScottGuymer (approved):

PR #4000, opened 2024-07-17
feat(webhook): Enable authorizer assignment to webhook

SSW-SCIENTIFIC:
We want to add a lambda authenticator to webhook API endpoint to restrict requests by source IP address. To achieve this, the following two options are possible:

1. add attributes to pass authorizer resource information to this module,
2. or simply ignores changes related to authorizer of `aws_apigatewayv2_route` resource,

However, 1. requires all informations to create `aws_apigatewayv2_authorizer` and `aws_lambda_permission`, this is little bit need consideration. Option 2. is, of cource, needs resouce update by hand, however, the very simple to achieve by ignoring attribute changes.

This PR aims to implement option 2.

### DIFF
--- a/modules/webhook/main.tf
+++ b/modules/webhook/main.tf
@@ -17,7 +17,8 @@ resource "aws_apigatewayv2_route" "webhook" {
 
   lifecycle {
     ignore_changes = [
-      # ignore authorization related attributes to enable authenticator assignment to API route
+      # Ignore authorization related attributes to enable authenticator assignment to API route.
+      # NOTE: We consider the ignores as a system intenral. Future changes will not trigger a breakig change.
       authorizer_id,
       authorization_type,
       authorization_scopes,
--- a/modules/webhook/main.tf
+++ b/modules/webhook/main.tf
@@ -14,6 +14,15 @@ resource "aws_apigatewayv2_route" "webhook" {
   api_id    = aws_apigatewayv2_api.webhook.id
   route_key = "POST /${local.webhook_endpoint}"
   target    = "integrations/${aws_apigatewayv2_integration.webhook.id}"
+
+  lifecycle {
+    ignore_changes = [
+      # ignore authorization related attributes to enable authenticator assignment to API route
+      authorizer_id,
+      authorization_type,
+      authorization_scopes,
+    ]
+  }
 }
 
 resource "aws_apigatewayv2_stage" "webhook" {


npalm:
Maybe great if you even can share an example how to attach the authroizer. I digged in some time ago but at that time it seems not a real option with the version of the gateway we are using. The only option I saw was an authorizer lambda to run the checks.

SSW-SCIENTIFIC:
@npalm 
Yes, you are right, only we can add lambda authorizer for API Gateway v2.

I show the example below:

1. create lambda for authorizer as follows:

```typescript
import * as process from "node:process";

import { APIGatewayRequestSimpleAuthorizerHandlerV2 } from "aws-lambda";
import { SSMClient, GetParameterCommand } from "@aws-sdk/client-ssm";
import { containsCidr } from "cidr-tools";

console.log("Loading function");

export const handler: APIGatewayRequestSimpleAuthorizerHandlerV2 = async (event) => {
  const allowlist = await fetchIPAllowList();

  if (containsCidr(allowlist, event.requestContext.http.sourceIp)) {
    return { isAuthorized: true };
  }

  return { isAuthorized: false };
};


const client = new SSMClient();

const fetchIPAllowList = async () => {
  const result = await client.send(
    new GetParameterCommand({
      Name: process.env.IP_ALLOWLIST_PARAMETER_NAME,
    }),
  );
  return JSON.parse(result.Parameter?.Value || "[]") as unknown as string[];
};
```

and create lambda resouce:

```
module "authorizer" {
  source = "terraform-aws-modules/lambda/aws"

  publish = true

  function_name = "source-ip-address-authorizer"
  description   = "Lambda Authorizer to protect API Gateway."
  handler       = "index.handler"
  runtime       = "nodejs20.x"

  environment_variables = {
    "IP_ALLOWLIST_PARAMETER_NAME" = module.ip_allowlist.ssm_parameter_name
  }

  allowed_triggers = {
    api-gateway = {
      service = "apigateway"
    }
  }

  attach_policy_json = true
  policy_json        = data.aws_iam_policy_document.authorizer.json

  source_path = "./authorizer/dist/index.mjs"
}

module "ip_allowlist" {
  source = "terraform-aws-modules/ssm-parameter/aws"

  name   = "/services/public-api/ip-allowlist"
  values = var.ip_allowlist
}

data "aws_iam_policy_document" "authorizer" {
  statement {
    effect    = "Allow"
    actions   = ["ssm:GetParameter"]
    resources = [module.ip_allowlist.ssm_parameter_arn]
  }
}
```

2. create lambda authorizer resouce for Terraform AWS GitHub Runner:

```terraform
resource "aws_apigatewayv2_authorizer" "authorizer" {
  authorizer_type                   = "REQUEST"
  api_id                            = module.github_runners.webhook.gateway.id
  authorizer_uri                    = module.authorizer.lambda_function_invoke_arn
  name                              = "source-ip-address-authorizer"
  authorizer_payload_format_version = "2.0"
  enable_simple_responses           = true
}
```

Here `module.github_runners` is (in my case) `multi-runner` module of `terraform-aws-github-runner`.

3. modify `aws_apigatewayv2_route.webhook` resouce manually, for example, by Web UI to attach lambda authorizer:

```diff
  resource "aws_apigatewayv2_route" "webhook" {
    api_id    = aws_apigatewayv2_api.webhook.id
    route_key = "POST /${local.webhook_endpoint}"
    target    = "integrations/${aws_apigatewayv2_integration.webhook.id}"

+   authorizer_type = "CUSTOM"
+   authorizer_id   = aws_apigatewayv2_authorizer.authorizer.id
  }
```

Thank you.

npalm:
I will catch-up with the PR this week was on leave last 1.5 week.

npalm:
As long we have no authorizer part of the module this change is fine. I hope to refactor the module soon to split the webhook in two parts. One responsible for accepting the event from GitHub including checks of the signature and possible IP allow lists. Once accpeted put the event on the eventbridge (AWS) and next let the a dispatcher put in on a queue for a scale-up functions. 

This change woudl allow anyone to use easier own streams of the job events, and even more events for any other logic. Including metrics.

modules/webhook/main.tf
```diff
@@ -14,6 +14,15 @@ resource "aws_apigatewayv2_route" "webhook" {
   api_id    = aws_apigatewayv2_api.webhook.id
   route_key = "POST /${local.webhook_endpoint}"
   target    = "integrations/${aws_apigatewayv2_integration.webhook.id}"
+
+  lifecycle {
+    ignore_changes = [
+      # ignore authorization related attributes to enable authenticator assignment to API route
```
npalm, line 20:
```suggestion
      # Ignore authorization related attributes to enable authenticator assignment to API route.
      # NOTE: We consider the ignores as a system intenral. Future changes will not trigger a breakig change.
```

npalm (changes requested):
Tiny suggestion.

npalm (approved):

PR #4001, opened 2024-07-17
feat: mark orphan runners before removing them

npalm:
## Problem
Orphan runners are deleted right after detection. This can be clash with self termination (ephemeral) runners. Typically the runner is waiting a few sseconds before exectuing a self termination. 

## Solution
In this solution we first mark a runner orphan, but not delete the runner. In a next cycle of the scale down function. First all orphan runners are terminated.

## Improvements

- Improved logging, only logging the main flow once at info. All other logs moved to debug
- Scale-down write permissions limitted to the envirnoment

## Todo

- [x] Update docs
- [x] Test default runner deployment
- [x] Test mult runner deployment

## Example of log

- Two instances
- One made orphan by removing the runner from GitHub
- In the log
  - Idle runner got removed
  - Orphan get marked as orphan
  - Next cycle orphan terminated.  
image


lambdas/functions/control-plane/src/aws/runners.ts
```diff
@@ -46,6 +48,9 @@ function constructFilters(filters?: Runners.ListRunnerFilters): Ec2Filter[][] {
       ec2FiltersBase.push({ Name: `tag:ghr:Type`, Values: [filters.runnerType] });
       ec2FiltersBase.push({ Name: `tag:ghr:Owner`, Values: [filters.runnerOwner] });
     }
+    if (filters.orphan) {
+      ec2FiltersBase.push({ Name: 'tag:ghr:orphan', Values: ['true'] });
```
stuartp44, line 52:
```suggestion
      ec2FiltersBase.push({ Name: 'tag:ghr:Orphan', Values: ['true'] });
```
Everything else uses caps?

lambdas/functions/control-plane/src/aws/runners.ts
```diff
@@ -85,6 +90,7 @@ function getRunnerInfo(runningInstances: DescribeInstancesResult) {
             type: i.Tags?.find((e) => e.Key === 'ghr:Type')?.Value as string,
             repo: i.Tags?.find((e) => e.Key === 'ghr:Repo')?.Value as string,
             org: i.Tags?.find((e) => e.Key === 'ghr:Org')?.Value as string,
+            orphan: i.Tags?.find((e) => e.Key === 'ghr:orphan')?.Value === 'true',
```
stuartp44, line 93:
```suggestion
            orphan: i.Tags?.find((e) => e.Key === 'ghr:Orphan')?.Value === 'true',
```
As Above

lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
```diff
@@ -293,10 +295,31 @@ describe('Scale down runners', () => {
...
+        // act
+        await scaleDown();
+
+        // assert
+        checkTerminated(runners);
+        checkNonTerminated(runners);
+
+        expect(mockTagRunners).toHaveBeenCalledWith(orphanRunner.instanceId, [
+          {
+            Key: 'orphan',
```
stuartp44, line 314:
```suggestion
            Key: 'Orphan',
```
As above?

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -194,11 +193,26 @@ async function evaluateAndRemoveRunners(
   }
 }
 
-async function terminateOrphan(instanceId: string): Promise<void> {
+async function markOrphan(instanceId: string): Promise<void> {
   try {
-    await terminateRunner(instanceId);
+    await tag(instanceId, [{ Key: 'orphan', Value: 'true' }]);
```
stuartp44, line 198:
```suggestion
    await tag(instanceId, [{ Key: 'Orphan', Value: 'true' }]);
```
as above

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -194,11 +193,26 @@ async function evaluateAndRemoveRunners(
...
+  }
+}
+
+async function terminateOrphan(environment: string): Promise<void> {
+  try {
+    const orphanRunners = await listEC2Runners({ environment, orphan: true });
+
+    for (const runner of orphanRunners) {
+      await terminateRunner(runner.instanceId).catch((e) => {
+        logger.error(`Failed to terminate orphan runner '${runner.instanceId}'`, { error: e });
```
stuartp44, line 211: Is there any need to do a catch twice?

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -221,14 +235,18 @@ async function listRunners(environment: string) {
...
 function filterRunners(ec2runners: RunnerList[]): RunnerInfo[] {
-  return ec2runners.filter((ec2Runner) => ec2Runner.type) as RunnerInfo[];
+  return ec2runners.filter((ec2Runner) => ec2Runner.type && !ec2Runner.orphan) as RunnerInfo[];
 }
 
 export async function scaleDown(): Promise<void> {
   githubCache.reset();
-  const scaleDownConfigs = JSON.parse(process.env.SCALE_DOWN_CONFIG) as [ScalingDownConfig];
   const environment = process.env.ENVIRONMENT;
+  const scaleDownConfigs = JSON.parse(process.env.SCALE_DOWN_CONFIG) as [ScalingDownConfig];
```
stuartp44, line 244: I have noted here that we parse without sanity checking it first?

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
stuartp44, line 211: Is there any point doing a catch twice? maybe worth doing it once in the catch block

stuartp44 (changes requested):

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
npalm: The config is typed in Terraform, but it is a loose coupling.

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
npalm: Did quite some testing with it, but it needed due to all the async executing.

lambdas/functions/control-plane/src/aws/runners.ts
npalm: Most of the tags we use are in lower case like `ghr:environment`, `ghr:ssm_config_path` would suggest we stick to lower case an update owner and type to lower as well in a sperate PR.

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
npalm: in fact this should be `ghr:orphan`

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
npalm: yes the inner try/catch is catchting error during termination. The outer any other inlcuding the listEC2Runners

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
npalm: created #4026 for formatting of the keuys.

Brend-Smits (approved):
LGTM!

Issue #4012, opened 2024-07-26
Webhook secret is printed out in the terraform logs


jizi:
## Expected Behavior
Plaintext webhook secret is not present in the terraform logs. Example
```
  # null_resource.update_app will be created
  + resource "null_resource" "update_app" {
      + id       = (known after apply)
      + triggers = {
          + "webhook_endpoint" = "https://xxxxxxxxx.execute-api.eu-central-1.amazonaws.com/webhook"
          + "webhook_secret"   = (sensitive value)
        }
    }
```

## Current Behavior
Plaintext webhook secret is printed out in the terraform logs. Example
```
  # null_resource.update_app will be created
  + resource "null_resource" "update_app" {
      + id       = (known after apply)
      + triggers = {
          + "webhook_endpoint" = "https://xxxxxxxxx.execute-api.eu-central-1.amazonaws.com/webhook"
          + "webhook_secret"   = "not_so_secret"
        }
    }
```
Module version: `5.12.2`

## Possible Solution
It should be enough to mark the `github_app` variable as sensitive in [modules/webhook-github-app/variables.tf](https://github.com/philips-labs/terraform-aws-github-runner/blob/main/modules/webhook-github-app/variables.tf)

PR #4013, opened 2024-07-26
fix: mark github_app variable as sensitive

jizi:
Resolves https://github.com/philips-labs/terraform-aws-github-runner/issues/4012


npalm (approved):
Thanks!

Issue #3966, opened 2024-07-04
Upgrade ockokit/rest  and octokit/auth-app breaks scale-up during authorization


npalm:
## Upgrading `@octokit/rest  ^20.0.1  20.0.2  ❯  20.1.1` is breaking the authentication of the app.

Running the scale-up lambda results in an error while getting an access-token for the app. 

```
{"level":"ERROR","message":"response.json(...).catch is not a function","service":"runners-scale-up","timestamp":"2024-07-04T14:01:17.879Z","region":"eu-central-1","environment":"npalm-multi-linux-x64","name":"HttpError","status":500,"request":{"method":"POST","url":"https://api.github.com/app/installations/123/access_tokens","headers":{"accept":"application/vnd.github.v3+json","user-agent":"octokit-request.js/8.4.0 Node.js/21","authorization":"bearer [REDACTED]"},"request":{}}}
```

## Upgrade octokit/auth-app 6.0.3 to 6.1.1 breaks build

Update auth-app introduce transive breaking dependencies requiring ESM.

wolfy1339:
Can you explain which transitive dependency you think is causing an issue?

The only dependency that has a `"type": "module"` in it`s `package.json` is `lru-cache` which ships CJS + ESM builds in the same package

npalm:
I tried to reproduce the problem today again. Seems both auth-app and rest updates are causing the issue. When update auth from 6.1.0 to 6.1.1 the error stats to occure. 

See here the change set. https://github.com/philips-labs/terraform-aws-github-runner/pull/4042/files

PR #4020, opened 2024-07-31
fix: bump node dependencies and cleanup

npalm:
- migrate to yarn4
- fix duplicate dependencies in lock file
- upgrade all possible upgrade
- octokit/rest and octokit/auth-app both break updates see #3966

Tested with deployment of multi runner example


Brend-Smits (approved):
LGTM!

PR #4041, opened 2024-08-06
fix: update octokit auth

npalm:
Tested with default example.


npalm:
Update to 6.1.0 works, 6.1.1 is breaking. See https://github.com/octokit/auth-app.js/issues/634

PR #4042, opened 2024-08-06
fix: update octokit auth from 6.1.0 to 6.1.1

npalm:
Updating octkit auth app to 6.1.1. fails runtime. Not in our tests due to most of the SDK layer is mocked.

Only information we catch is a generic Error.
```
 HttpError: response.json(...).catch is not a function
```


lambdas/functions/control-plane/src/gh-auth/gh-auth.ts
```diff
@@ -32,12 +30,12 @@ export async function createOctoClient(token: string, ghesApiUrl = ''): Promise<
   return new CustomOctokit({
     ...ocktokitOptions,
     throttle: {
-      onRateLimit: (options: { method: string; url: string }) => {
+      onRateLimit: (retryAfter: number, options: Required<EndpointDefaults>) => {
```
npalm, line 33: change forced by update octokit/rest

.github/dependabot.yml
```diff
@@ -36,8 +36,6 @@ updates:
         update-types: ["version-update:semver-major"]
       - dependency-name: "@octokit/*"
         update-types: ["version-update:semver-major"]
-      - dependency-name: "@octokit/rest"
```
npalm, line 39: remove dependabot ignores for octokit rest and auth app

npalm:
Tested with the default example:

Logs scale-up
image

Tracing example. Removing the axios fetch does NOT break the tracing of the github calls.
image

stuartp44 (approved):
LGTM

PR #4046, opened 2024-08-07
fix: align inlince policies

### LINKED ISSUE #4045: Missing name for inlince xray policy
Inline policy for xray scale-up lambda. And most likely all other lambda's is missing a name. See https://github.com/philips-labs/terraform-aws-github-runner/blob/c15c99db84132b11f886f6f8e5291af5e357b41e/modules/runners/scale-up.tf#L140

npalm:
## Description
Most of the lambdas are defining inline policies in the dedicated role for the lambda. The naming of the inline policies is a bit chaotic. This PR aligns the naming removes redundant parts of the naming like the the lamba name or prefix. This is allready part of the role

fix #4045


stuartp44 (approved):
LGTM

PR #4063, opened 2024-08-13
feat: add time zone support for pool schedules

### LINKED ISSUE #4056: Support Time Zone for Runner Pool
Hey, thanks for the work you've done on this project (and for open sourcing it)!

Would you consider adding support for providing the Time Zone as part of the `pool_config` variable?

For context, we want to keep a pool of runners ready during office hours, but we'd currently need to convert the `pool_config[*].schedule_expression` to UTC and update that whenever the clocks change (e.g., I'm in the UK, where the start/end of "British Summer Time" changes every year).

## Workaround

Unfortunately, AWS EventBridge Rules (currently used to trigger the pool function) don't support time zones, but [AWS EventBridge Scheduler does](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/scheduler_schedule). This feature is [a drop-in replacement](https://aws.amazon.com/blogs/compute/introducing-amazon-eventbridge-scheduler/) for EventBridge Rules (including the cron expression) and [the same price](https://aws.amazon.com/eventbridge/pricing/).

My current workaround is to define the `aws_scheduler_schedule` resources outside the GitHub runners module and trigger the pool Lambda, but that has a couple of issues:
* It relies on implementation details of your module, so may break in the future (specifically, the payload to pass to the pool function)
* At least one `pool_config` expression is required, otherwise the pool resources aren't created

## Proposal

1. Add an optional `time_zone` property to the `pool_config` variable
2. Replace the use of `aws_cloudwatch_event_rule` in the `pool` module with `aws_scheduler_schedule` (and a single `aws_scheduler_schedule_group` and `aws_iam_role`), setting `schedule_expression_timezone` if necessary.

Alternatively, dynamically switch between `aws_cloudwatch_event_rule` and `aws_scheduler_schedule`, depending on whether the `time_zone` is set. This would increase the complexity of the module, but would avoid making any changes to existing users.

If you'd consider accepting one of these as a PR, I can look at implementing one of them.

janslow:
Fixes #4056

The timezone can be set via the `schedule_expression_timezone` property via the main module or `multi-runners` module.

Replaces the `aws_cloudwatch_event_rule` in the `pool` module (which only support UTC) with `aws_scheduler_schedule`, which supports arbitrary time zones via the `schedule_expression_timezone` attribute.

In this situation AWS EventBridge Scheduler is a drop in replacement for schedule-based AWS EventBridge Rules ([see AWS blog post](https://aws.amazon.com/blogs/compute/introducing-amazon-eventbridge-scheduler/)), including the expression syntax and pricing. The main (internal) difference is that AWS Scheduler requires the use of an IAM Role, instead of Lambda permissions.


janslow:
Note, I've tried to run the GitHub Action to update the auto-generated Terraform Docs ([as per the contribution guide](https://github.com/philips-labs/terraform-aws-github-runner/blob/main/CONTRIBUTING.md#submitting-a-pull-request)), but it's [failing](https://github.com/janslow/terraform-aws-github-runner/actions/runs/10369127870/job/28704207502) because [it requires credentials](https://github.com/janslow/terraform-aws-github-runner/blob/main/.github/workflows/update-docs.yml#L22-L23) that I don't have access to.

In terms of testing, I couldn't find any existing automated tests for the pool, although I have tested it on my existing install (both with and without the new property)

npalm:
> Note, I've tried to run the GitHub Action to update the auto-generated Terraform Docs ([as per the contribution guide](https://github.com/philips-labs/terraform-aws-github-runner/blob/main/CONTRIBUTING.md#submitting-a-pull-request)), but it's [failing](https://github.com/janslow/terraform-aws-github-runner/actions/runs/10369127870/job/28704207502) because [it requires credentials](https://github.com/janslow/terraform-aws-github-runner/blob/main/.github/workflows/update-docs.yml#L22-L23) that I don't have access to.

Terraform docs will updated on either release or main branch. Feel free to create an issue to address the shortcoming in the release notes
> 
> In terms of testing, I couldn't find any existing automated tests for the pool, although I have tested it on my existing install (both with and without the new property)

We only have automated tests for the lambda's, not for the full module. We doe typically a manual test dependending on the change with one of the examples. Would be great if we have an full auto test test. Any idea would be welcome how to run auto tests for IaC on an OS repo in a secure way.

janslow:
>  Feel free to create an issue to address the shortcoming in the release notes

👍 [I've created a PR to remove the reference from the contribution guide](https://github.com/philips-labs/terraform-aws-github-runner/pull/4065), as it's no longer relevant when creating a PR.

janslow:
> Would be great if we have an full auto test test. Any idea would be welcome how to run auto tests for IaC on an OS repo in a secure way.

Unfortunately, I'm not aware of any magic solution.

For internal apps, I've used [Terratest](https://terratest.gruntwork.io/) to spin up ephemeral environments that are realistic, then run integration tests against them. I think that'd be tricky for this project though, especially given the fact it's a public project.

One possibility could be to use [Terraform's native test features](https://developer.hashicorp.com/terraform/language/tests#syntax) to plan the example modules (using [mock providers](https://developer.hashicorp.com/terraform/language/tests/mocking), so no AWS access is required), then running assertions against it. For example, asserting that all of the `aws_iam_role` resources have the correct path/permissions boundary (which I originally forgot in my PR).

npalm:
> > Would be great if we have an full auto test test. Any idea would be welcome how to run auto tests for IaC on an OS repo in a secure way.
> 
> Unfortunately, I'm not aware of any magic solution.
> 
> For internal apps, I've used [Terratest](https://terratest.gruntwork.io/) to spin up ephemeral environments that are realistic, then run integration tests against them. I think that'd be tricky for this project though, especially given the fact it's a public project.
> 
> One possibility could be to use [Terraform's native test features](https://developer.hashicorp.com/terraform/language/tests#syntax) to plan the example modules (using [mock providers](https://developer.hashicorp.com/terraform/language/tests/mocking), so no AWS access is required), then running assertions against it. For example, asserting that all of the `aws_iam_role` resources have the correct path/permissions boundary (which I originally forgot in my PR).

Thanks for the feedback! That was also my take and the reason why there no intergration tests (yet). Internally for the runners we deploy to a staging and ran validation workflows as verification. All automated.

examples/ephemeral/main.tf
```diff
@@ -62,8 +62,9 @@ module "runners" {
   # # Example of simple pool usages
   # pool_runner_owner = "philips-test-runners"
   # pool_config = [{
-  #   size                = 3
-  #   schedule_expression = "cron(* * * * ? *)"
+  #   size                         = 3
+  #   schedule_expression          = "cron(* * * * ? *)"
+  #   schedule_expression_timezone = "Australia/Sydney"
```
npalm, line 67: Maybe also update the exmple cron expression that make sense for a certain time zone.
```suggestion
  #   size                         = 3
  #   schedule_expression = "cron(0/3 14 * * ? *)" # every 3 minutes between 14:00 and 15:00
  #   schedule_expression_timezone = "Europe/Amsterdam"

```

npalm (changes requested):
Thanks, nice contribution. Tiny suggestion.

npalm:
Tested with ephemeral example, works as expected.

janslow:
Thanks for the review, I've accepted your suggestion

npalm (approved):

npalm:
Thanks for adding the feature! Feel free to update the other time triggers as well!

npalm (approved):

PR #4064, opened 2024-08-13
feat: scale up for long waiting jobs (job retry)

npalm:
## Description

This feature add the capability to retry scaling a runner when a job is still queued after a defined delay. This feature is added to avoid pool for ephemeral runners.

## Implementation

The module is extended with configuration top optional enable one or more retries. Once enabled the scale-up lambda will publish the same message as it recieves extend with a counter on a retry-job-queueu with a delay. A new lambda will pick the message from this queue and checks if the job is still queued (via GitHub API). In case it is still queued it is published again on je the job queue, incoming queue of the scale-up lambda

## Consequences

- This feature is meant for small fleets with ephemeral runners. Each retry check is casuing a GitHub API which can trigger a rate limit for the app. 
- This feature should make ephemerla runners more resposnive without having a pool to pick up missed jobs.
- The module allows you to force a job check before scaling, this check should be disabled.
- The delay should be set to a time that is higher than the normal boottime of a runner.

## Testing

Testing can be done as follow
- Trigger a workflow
- Terminate the created instance before the job starts
- Wait, after the delay the retry job should publish the message again which triggers a new instance creation.

- [x] Multi runners.
- [x] Default runners, not enabled requires configuraton update

## Tasks

- [x] Update docs
- [x] Update multi-runner
- [x] Check CMK keys for SQS
- [x] Limit delay to max delay of a queue.
- [x] Add optional metric for retry
- [x] Update issue with more details


examples/default/main.tf
```diff
@@ -116,6 +116,18 @@ module "runners" {
     }
   }
 
+  # enable job_retry feature. Be careful with this feature, it can lead to API rate limits.
```
npalm, line 119: By default disabled in the example, code is for inspiration

lambdas/functions/control-plane/jest.config.ts
```diff
@@ -6,10 +6,10 @@ const config: Config = {
   ...defaultConfig,
   coverageThreshold: {
     global: {
-      statements: 97.99,
-      branches: 97.26,
-      functions: 95.45,
-      lines: 97.92,
+      statements: 97.78,
```
npalm, line 9: Tiny change due to refactoring in gh-auth

lambdas/functions/control-plane/src/gh-auth/gh-octokit.ts
npalm, line 1: The current way how we setup octokit is hard to mock. Started moving function around in this new module to make mocking of creating the client easier.

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
```diff
@@ -0,0 +1,90 @@
...
+  queueUrl: string;
+}
+
+export async function publishRetryMessage(payload: ActionRequestMessage): Promise<void> {
+  if (process.env.JOB_RETRY_CONFIG === undefined) {
+    logger.debug('Job retry config not found, skipping retry');
+    return;
+  }
+
+  const jobRetryConfig = JSON.parse(process.env.JOB_RETRY_CONFIG) as JobRetryConfig;
```
Brend-Smits, line 26: What would happen here if the JOB_RETRY_CONFIG is invalid? So if parsing fails.. I think an error should be logged and process should return.

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
```diff
@@ -0,0 +1,90 @@
...
+  payload.retryCounter = payload.retryCounter != undefined ? payload.retryCounter + 1 : 0;
+
+  if (jobRetryConfig.enable && payload.retryCounter < jobRetryConfig.maxAttempts) {
```
Brend-Smits, line 29: Could move `jobRetryConfig.enable` into an if statement before. This will make things easier to read and make logging easier. Depending on the state, the log could simply state "Job retry is disabled, skipping retry". Would leave the user not guessing in the future.

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
```diff
@@ -0,0 +1,90 @@
...
+    logger.debug(`Job retry is enabled and retry counter is below max attempts, publishing message for retry`, {
+      message: payload,
+      config: jobRetryConfig,
+    });
+    let delay = jobRetryConfig.delayInSeconds * Math.pow(jobRetryConfig.delayBackoff, payload.retryCounter);
+    delay = Math.min(delay, 900); // max delay of 15 minutes
```
Brend-Smits, line 35: I would move this into it's own function and test it.

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
```diff
@@ -0,0 +1,90 @@
...
+    await publishMessage(JSON.stringify(payload), jobRetryConfig.queueUrl, delay);
+    logger.info(`Messages published for retry check with a delay of: '${delay}' seconds`);
+  } else {
+    logger.debug(`Job retry is disabled or max attempts reached, skipping retry`, { payload });
```
Brend-Smits, line 39: Once you add an additional if statement to check if the job retry is even enabled. This log message can also be clarified to only include the message about max-attempts being reached.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -223,7 +231,6 @@ export async function scaleUp(eventSource: string, payload: ActionRequestMessage
   const runnerLabels = process.env.RUNNER_LABELS || '';
   const runnerGroup = process.env.RUNNER_GROUP_NAME || 'Default';
   const environment = process.env.ENVIRONMENT;
-  const ghesBaseUrl = process.env.GHES_URL;
```
Brend-Smits, line 226: Why did you remove this?

modules/runners/job-retry.tf
```diff
@@ -0,0 +1,34 @@
...
+    role_permissions_boundary   = var.role_permissions_boundary
+    s3_bucket                   = var.lambda_s3_bucket
+    tracing_config              = var.tracing_config
+    github_app_parameters       = var.github_app_parameters
+    enable_organization_runners = var.enable_organization_runners
+    sqs_build_queue             = var.sqs_build_queue
+  }
+}
+
+module "job_retry_check" {
```
Brend-Smits, line 29:
```suggestion
module "job_retry" {
```
to align naming with the rest..

modules/runners/job-retry/README.md
```diff
@@ -0,0 +1,101 @@
+# Module - Termination Watcher
```
Brend-Smits, line 1: This whole documentation needs updating. This is not the README for the job-retry, but a copied version of the termination watcher?

modules/runners/job-retry/main.tf
```diff
@@ -0,0 +1,95 @@
+locals {
+  lambda_zip = var.config.zip == null ? "${path.module}/../../../lambdas/functions/control-plane/runners.zip" : var.config.zip
+  name       = "job-retry-check"
```
Brend-Smits, line 3: So.... job-retry or job-retry-check?
If it's job-retry-check, we should align naming of files and directories. IF not, name should be changed.

Brend-Smits: The same goes for every _check below.

modules/runners/scale-up.tf
```diff
@@ -1,3 +1,13 @@
+locals {
+  job_retry_config = local.job_retry_enabled ? {
+    enable         = var.job_retry.enable
+    maxAttempts    = var.job_retry.max_attempts
+    delayInSeconds = var.job_retry.delay_in_seconds
+    delayBackoff   = var.job_retry.delay_backoff
+    queueUrl       = module.job_retry_check[0].job_retry_check_queue.url
```
Brend-Smits, line 7: So if this variable is not set, the queueUrl will be empty? What happens if it's empty? Will something down the line crash?

Brend-Smits (changes requested):
Nice! I have yet to test these changes, will do it after pressing this request changes button.

docs/index.md
```diff
@@ -66,10 +66,26 @@ The AMI cleaner is a lambda that will clean up AMIs that are older than a config
 
 ### Instance Termination Watcher
 
-> This feature is Beta, changes will not trigger a major release as long in beta.
+!!! Warning
```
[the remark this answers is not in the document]
npalm: No this is intended, is is mkdocs markdown

docs/index.md
```diff
@@ -66,10 +66,26 @@ The AMI cleaner is a lambda that will clean up AMIs that are older than a config
...
+    This feature is Beta, changes will not trigger a major release as long in beta.
 
 The Instance Termination Watcher is creating log and optional metrics for termination of instances. Currently only spot termination warnings are watched. See [configuration](configuration/) for more details.
 
+
+### Job Retry
+
+!!! Warning
+
+    This feature is Beta, changes will not trigger a major release as long in beta.
```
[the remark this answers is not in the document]
npalm: mkdocs markdown, this is ok

docs/index.md
```diff
@@ -66,10 +66,26 @@ The AMI cleaner is a lambda that will clean up AMIs that are older than a config
...
+
+The Job Retry will allow you to retry scaling when a job is not started. When enabled the scale up lambda will send a retry message to the a SQS queue. The job retry lambda will check after a delay if the job is still queued. And if so it will send a retry command de the scale up lambda via SQS. The feature is designed to be used with ephemeral runners. The feature is opt in, it will not be created by default.
```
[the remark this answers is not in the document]
npalm: Title is here, see line 76

modules/runners/job-retry/README.md
npalm: LOL the better copy paste :)

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
npalm: logic is moved to a function. The GHES_URL is part of the unit test so certainly checked.

docs/index.md
```diff
@@ -66,10 +66,26 @@ The AMI cleaner is a lambda that will clean up AMIs that are older than a config
...
+!!! Warning
+
+    This feature is Beta, changes will not trigger a major release as long in beta.
 
 The Instance Termination Watcher is creating log and optional metrics for termination of instances. Currently only spot termination warnings are watched. See [configuration](configuration/) for more details.
 
+
+### Job Retry
+
+!!! Warning
```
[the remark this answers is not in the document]
npalm: mkdocs markdown slang

modules/runners/job-retry/README.md
npalm: good catch

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
npalm: for just 2 lines, lets keep it here

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
npalm: Parse error. Not handled at all by the module. But the terraform input here is part of the system (interal) and typed. In case you would like to address this futerh. i suggest we make an issue to fix this in general.

lambdas/functions/control-plane/src/scale-runners/job-retry.ts
npalm: done

Brend-Smits (approved):
Looks good. I tested it and works :) 
Good job 👍🏼

mariusfilipowski:
Thanks for this interesting option. 
1) What do you mean with small fleets? 
2) I assume that these rate limits won't apply to GHES, will they?
3) Can this also be used with idle/pooled runners or do they have to be turned off? We have certain problem in the night, where we set the pool to 0, where we would expected that this approach could help us to get more stability.

PR #4095, opened 2024-08-27
chore: lock and update dependencies, clean-up

npalm:
- update and lock workflows
- remove dependabot auto approve, due to bad practice


rjaegers (approved):

rjaegers (changes requested):
There are some instances left that are (or were previously) managed by ratchet. Search for 'ratchet:', I don't know if you want to replace those as well, pin them and put them under Dependabot control.

rjaegers (approved):

PR #4160, opened 2024-10-02
refactor: split webhook accepting and queue dispatching logic

npalm:
## Description

This PR refacting the webhook lambda. It splits the webhook logic in a part responsible to accept the event and a part to dispatch to the runner queues. Secondly it refactor and removes duplicate logic in tests.

This change should allow to split the lambda in two lambda and and use the AWS EventBridge to accept events, and use the dispacher only for delegating an event to a runner queueu.


lambdas/functions/webhook/src/webhook/index.ts
npalm, line 1: the webhook module will be the entry point for the messages. When adding the eventbridge the webhook will deliver to the bus instead forward to the dispatcher.

lambdas/functions/webhook/src/lambda.test.ts
```diff
@@ -85,28 +85,28 @@ describe('Test scale up lambda wrapper.', () => {
...
   });
 
   it('Errors are not thrown.', async () => {
     const mock = mocked(handle);
     const logSpy = jest.spyOn(logger, 'error');
     mock.mockRejectedValue(new Error('some error'));
     const result = await githubWebhook(event, context);
-    expect(result).toMatchObject({ statusCode: 500 });
-    expect(logSpy).toBeCalledTimes(1);
+    expect(result).toMatchObject({ body: 'Check the Lambda logs for the error details.', statusCode: 500 });
```
stuartp44, line 109: Can we not be consistent with responses, if we respond with JSON normally, can we not respond with JSON if there is a error?

lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -0,0 +1,87 @@
...
+          eventType: githubEvent,
+          installationId: body.installation?.id ?? 0,
+          queueId: queue.id,
+          queueFifo: queue.fifo,
+          repoOwnerType: body.repository.owner.type,
+        });
+        logger.info(`Successfully dispatched job for ${body.repository.full_name} to the queue ${queue.id}`);
+        return {
+          statusCode: 201,
+          body: `Successfully queued job for ${body.repository.full_name} to the queue ${queue.id}`,
```
stuartp44, line 52: Can we not respond with JSON rather than plain text in the body?

lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+async function handleWorkflowJob(
+  body: WorkflowJobEvent,
+  githubEvent: string,
+  matcherConfig: Array<RunnerMatcherConfig>,
+): Promise<Response> {
+  if (body.action === 'queued') {
+    // sort the queuesConfig by order of matcher config exact match, with all true matches lined up ahead.
```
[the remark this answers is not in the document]
npalm: the comment was already here, it is pointing to the sort below.

lambdas/functions/webhook/src/runners/dispatch.ts
npalm: This should be a json, this is the standard for javascript function. I only refactored the internals. And planning in the next cycle to refactor the exception to the same result objects. Results objects are very common in languages like javascript.

lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+    logger.warn(`Received event contains runner labels '${body.workflow_job.labels}' that are not accepted.`);
+    return {
+      statusCode: 202,
+      body: `Received event contains runner labels '${body.workflow_job.labels}' that are not accepted.`,
+    };
+  }
+  return {
+    statusCode: 201,
+    body: `Received not queued and will not be ignored.`,
```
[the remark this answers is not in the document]
npalm: this is a json object. Also not changing the internal. The body goes in the HTTP respons. And is shown in the GitHub app. Feel free to see how we can improve. But for this refactoring I do not want to change the outside behaviour

lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -0,0 +1,87 @@
...
+  runnerLabelsMatchers = runnerLabelsMatchers.map((runnerLabel) => {
+    return runnerLabel.map((label) => label.toLowerCase());
+  });
+  const matchLabels = workflowLabelCheckAll
+    ? runnerLabelsMatchers.some((rl) => workflowJobLabels.every((wl) => rl.includes(wl.toLowerCase())))
+    : runnerLabelsMatchers.some((rl) => workflowJobLabels.some((wl) => rl.includes(wl.toLowerCase())));
+  const match = workflowJobLabels.length === 0 ? !matchLabels : matchLabels;
+
+  logger.debug(
+    `Received workflow job event with labels: '${JSON.stringify(workflowJobLabels)}'. The event does ${
```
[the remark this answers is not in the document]
npalm: I think a single line for a log statement is better, also this is not changed.

stuartp44 (approved):

PR #4161, opened 2024-10-02
fix: reduce preferred version to match the rest of the modules

rishta:


npalm (changes requested):
Thx, can you adjust also adjust `modules/download-lambda/versions.tf` in the same PR?

stuartp44:
I have moved this to issue #4204

PR #4174, opened 2024-10-08
docs: Change root to Top Level Directory

stuartp44:
docs: After seeing a few issues being opened regarding failed usage of modules, I think the language surrounding the menus is wrong or misleads a little. I have reworded this to hopefully help.

### DIFF
--- a/docs/modules/runners.md
+++ b/docs/modules/runners.md
@@ -1,4 +1,4 @@
-# Runner module (root)
+# Runner module (Top Level Directory)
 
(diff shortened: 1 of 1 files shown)


docs/modules/runners.md
```diff
@@ -1,4 +1,4 @@
-# Runner module (root)
+# Runner module (Top Level Directory)
```
npalm, line 1: In terraform term we speaking about modules, the ones below the first level (parten / top-level / root) are called submodule. When would like to make it explict here, would prefer to keep it short (so it renders nice). Either parent or top-level.
```suggestion
# Runner module (Top Level)
```

npalm (changes requested):

docs/modules/runners.md
npalm: Please see PR for a suggestion

Issue #1257, opened 2021-10-11
Securing the API Gateway / Lambda Invocation?


toast-gear:
What is the current way of locking down the solution to only being callable by GitHub's webhook IP addresses? https://api.github.com/meta

Would it be worth raising a PR to have a solution for this baked in? I would think anyone deploying this would want to ensure only GitHub and perhaps the ability to include some custom IPs can call the webhook Lambda?

EDIT Perhaps it's worth adding an explicit role to the API Gateway Lambda integration to introduce a condition assessing the  `aws:SourceIp` key for GitHub's `hook` address ranges? https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_condition-keys.html

<img width="903" alt="image" src="https://user-images.githubusercontent.com/15716903/136852933-89daad72-85e2-4426-a846-66db8d76879f.png">

The language isn't great but I think this is the attribute we would need to set
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/apigatewayv2_integration#credentials_arn

I'm basing this off the fact that in v1 https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_integration#credentials credentials definitely the correct attribute

npalm:
Thanks for the feedback, current lock down is limited. Only the signature is verified. For well signed event the lambda only scales if the repo is existing, part of the org and have jobs queued. 

We certainly like to have a better protection. GitHub published IP addresses from which traffic is sent, so limitting on the IP addresses is a posibility. But we have also to keep in mind that for GHES the situation is different. So that requires that we have build a flexible and configurable solution. AWS also has this blog post that explains how to limit traffic to the API Gateway: https://aws.amazon.com/premiumsupport/knowledge-center/api-gateway-resource-policy-access/

Limitting by role and policy that only the API gateway can call the webhook lambda sound as a good plan to me.

Feel free to start a PR (while working on the PR feel free to create a draft PR). Maybe an ideaa to split the PR in two PR's

PR #4205, opened 2024-10-24
Initial implementation of IPv4 allow list in the webhook endpoint

PerGon:
(OK... this is embarrassing... as I'm about to open the PR and write the description - most code already written - [I found this one here](https://github.com/philips-labs/terraform-aws-github-runner/pull/4000) 🤦  damn! I still think there is some value in the PR tho)

There [have been requests before](https://github.com/philips-labs/terraform-aws-github-runner/issues/1257) of having some sort of IP allowlist.

This PR implements something like that. It creates a lambda that will be used as an `api gateway authorizer` to validate if the client ID is part the configured allowed CIDR ranges. (it is, pretty much exactly what was suggested [here](https://github.com/philips-labs/terraform-aws-github-runner/pull/4000))
Initially this will mostly be valid for self-hosted Github, but it can eventually be extended to load the IPs for Github cloud from wherever Github publishes them.

I'm creating as a draft PR as there's a couple of things missing (var descriptions), but I was keen on early feedback before taking the time to do those things 😅 

So, for the maintainers: should I spend more time in making this PR "mergeable" so it can be available for more people? No hard feelings if not 😅 I can add the relevant capability just in our case like explained in the other PR.


npalm:
Thx for taking the time to start the work on a allow list for incoming traffic. Indeed we had a previous PR, but we also have not always thte time to check PR. I know it is a pitty but a fact at the moment.

We also have refactored thw webhook module, which creates breaking changes in this PR. However we are open to support an IP allow list. I think we could consider several options.

1. Refactor the webhook further and allow to bring your own API GA, with the API GW 1 IP filterlign can be implemented, it also allows to put the WAF in front of it.
2. Add support for the API GW 1
3. Use indeed the API autherozier, but this will be antother lambda to run. 
4. Implment logic in the lambda.

## Lambda option

Adding support to the lambda is most likely the simplest approach. Adding the support straitgh away is OK to me. Would suggest to mark it eperimental, which allows us to replace it with one of the other options over time.

Some implementation directions / thoughts.
- Add a method to check here, just before the signature check: https://github.com/philips-labs/terraform-aws-github-runner/blob/0d87aec9b9cc487a221e7eabc075e6be77252021/lambdas/functions/webhook/src/webhook/index.ts#L20
- Add configuration item for IP_ALLOW_LIST to https://github.com/philips-labs/terraform-aws-github-runner/blob/main/lambdas/functions/webhook/src/ConfigLoader.ts to both `ConfigWebhook` and `ConfigWebhookEventBridge`
- Add top level configuration to runner (root module) as well multi-runner. Inject to the webhook mddule. And via the configuration object is the module sent downstream to both the eventbridge and direct module to configure the lambda.

PR #4226, opened 2024-10-31
fix: remove unused example spot_instance_termination_watcher

piscue:
I don't see this being used, and the only example available, it lead to confusion

### DIFF
--- a/examples/multi-runner/main.tf
+++ b/examples/multi-runner/main.tf
@@ -103,11 +103,6 @@ module "runners" {
   # Enable debug logging for the lambda functions
   # log_level = "debug"
 
-  # Enable spot termination watcher
-  # spot_instance_termination_watcher = {
-  #   enable = true
-  # }
-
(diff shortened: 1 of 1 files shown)


npalm (commented):
Thx for taking the time to make a contribution. However this example is there for a reason. People can quickly see what is possible and enable it. 

But I think we should find a better way how we do this for examples, because drawback is we can not verify the code in CI.

piscue:
I did double check the adding it to our multi runner setup and the plan show us the error

```hcl
An argument named "spot_instance_termination_watcher" is not expected here.
```

it could be that was renamed to the example after? we use `instance_termination_watcher` without problems

```hcl
  # Enable spot termination watcher
  # spot_instance_termination_watcher = {
  #   enable = true
  # }

  # Enable to track the spot instance termination warning
  # instance_termination_watcher = {
  #   enable         = true
  # }
```

npalm (approved):

PR #4243, opened 2024-11-07
fix: trigger cold start dispatcher after update SSM parameters

npalm:
## Description

Currently we use the lifecyle trigger in Terraform to enforce lambda reload. Unwanted side-effect is resources getting destroyed. Even leading to the lambda trigger for the dispatch functiong not getting re-created.

Instead of adding the lifecyle event to multiple resources, the version of the SSM paramater is added to environment variables of the Lambda to enforce [cold start](https://docs.aws.amazon.com/lambda/latest/operatorguide/execution-environments.html) after updates. 

## Tested
- default example
- multi runner, changing labels


Brend-Smits (approved):
LGTM!

Issue #4248, opened 2024-11-08
Add in packer image example improvements


iataylor:
Overview : After implementing and experimenting with the runners for my whole company, I've happened upon quite a few improvements to the default images. I'd like to propose adding them to the project, but I'll need permission to open a branch first.

Items : 
- Winrm provisioner is slow, so I set up an ssh provisioner by default instead, which almost halves our build time once we have multiple provisioner scripts.
- For any practical use of the Windows runners, the default storage space (I think it's 30 gigs) is way too small. Updating the default images' volume size to 100 gigs is a much more reasonable size.
- Improvements in the chocolatey install that takes less time, and ensures the path is set at startup.

npalm:
@iataylor improved examples would be very welcome. But no need to have permissions on this repository. You simply fork this repository. Make the changes on the fork (or an branch in the fork). Next you open a PR back to this repository to contibute back. Looking forward to your PR (pull request).

iataylor:
https://github.com/philips-labs/terraform-aws-github-runner/pull/4258

PR #4258, opened 2024-11-13
perf: Improve Windows packer examples

iataylor:
Changes made per the issue here https://github.com/philips-labs/terraform-aws-github-runner/issues/4248.

- The SSH packer provisioner is much more reliable, and much quicker, than the WinRM provisioner originally used. 
- The disk size should be increased from 30 -> 100 gigs by default for Windows instances, as we fill most of that space just by creating the image with basic dependencies.
- Adding AWS account variable for the ami_users parameter for amazon-ebs module.
- Changing the way we install chocolatey and the cloudwatch-agent from Invoke-WebRequest to System.Net.WebClient due to slowness issues during builds with the former.
- Removing InitializeInstance in Windows Server 2022. Not appropriate for Server 2022 (only 2016 and 2019), so I pulled it out. Not necessary either, as my runners run totally fine without it.

That should be most of it.


images/windows-core-2019/github_agent.windows.pkr.hcl
```diff
@@ -77,19 +92,16 @@ source "amazon-ebs" "githubrunner" {
...
   }
-  user_data_file = "./bootstrap_win.ps1"
-  winrm_insecure = true
-  winrm_port     = 5986
-  winrm_use_ssl  = true
-  winrm_username = "Administrator"
 
   launch_block_device_mappings {
     device_name           = "/dev/sda1"
+    volume_size           = 100
```
iataylor, line 104: We could leave this parameterized. I've just set this to 100 gigs hardcoded.

iataylor:
All tests passed locally.

npalm:
The CI checks for packer are failing. I will mark the PR as draft. Please feel free to mark it ready for review when you are.

stuartp44:
The CI checks are failing on formatting, just do a packer fmt and it should then actually test the changes @iataylor.

Issue #4260, opened 2024-11-15
[Feature Request] Configuration option for runner job hooks


winwinashwin:
Action runners allow configuring pre/post job hooks. This is officially intended for administrators to configure custom scripts to be invoked at the start/end of every job.

Refer official ADR: [Notification Hooks for Runners](https://github.com/actions/runner/blob/main/docs/adrs/1751-runner-job-hooks.md)

Prior to release `5.12.2`, this configuration via terraform was not possible for reasons outlined [here](https://github.com/philips-labs/terraform-aws-github-runner/issues/3854#issuecomment-2162696151)

The internal hook was removed in https://github.com/philips-labs/terraform-aws-github-runner/commit/bbcb4707f698b29c470c0aadf81c9ac2c37781f8, as a consequence, it is possible to configure the hook with the `userdata_post_install` option (but in a rather obscure way) as explained [here](https://github.com/philips-labs/terraform-aws-github-runner/issues/3854#issuecomment-2478625092).

This is still error prone as the user might not be aware "if" an internal hook is set in `start-runner.sh` similar to versions prior `5.12.2`.

I personally prefer the transparency in config options that is there today like [`userdata_template`](https://philips-labs.github.io/terraform-aws-github-runner/modules/runners/#input_userdata_template) as the user is clearly made aware that they are "replacing" the default template.

Requesting to add out of the box configuration options to set `ACTIONS_RUNNER_HOOK_JOB_STARTED` and `ACTIONS_RUNNER_HOOK_JOB_COMPLETED` hooks.

Related to: https://github.com/philips-labs/terraform-aws-github-runner/issues/3854

I'm happy to contribute if there is a decision to move ahead with the feature request.

Issue #3854, opened 2024-04-19
Clear working directory between runs?


TaxBusby:
Hi there, I'm using the non-ephemeral runners, which is fine with me for performance/isolation tradeoff.

However, it looks like the working directory given to workflows does not get cleared between jobs. This can lead to surprise for developers who expect a clean working directory.

Is there an option available to have the runners clear the working directory between runs?

--- 

Here's a short workflow I'm using to test this case:

```yaml
name: Ephemerality Test
on:
  workflow_dispatch:

jobs:
  write:
    runs-on: [self-hosted]
    name: Write
    steps:
      - name: Write files
        run: |
          echo Current Directory: $(pwd)
          ls
          echo "Local File Exists" > hello.txt
          echo "Temp File Exists" > /tmp/hello.txt
          
  read:
    runs-on: [self-hosted]
    needs: [write]
    name: Read
    steps:
      - name: Test for persisted files
        run: |
          echo Current Directory: $(pwd)
          ls
          cat hello.txt || echo "Local dir cleared"
          cat /tmp/hello.txt || echo "Tmp dir cleared"
          
          if [ -f hello.txt ]; then
              echo "Local file was not cleared!"
              exit 1
          fi
          if [ -f /tmp/hello.txt ]; then
              echo "Temp file was not cleared!"
              exit 1
          fi
```

Outputs:

```
// Write
Current Directory: /opt/actions-runner/_work/gh-self-hosted-runners/gh-self-hosted-runners
hello.txt

// Read
Current Directory: /opt/actions-runner/_work/gh-self-hosted-runners/gh-self-hosted-runners
hello.txt
Local File Exists
Temp File Exists
Local file was not cleared!
```

winwinashwin:
We were facing the same issue at our organization. We used [pre/post job hooks](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job) to get around this.

You can set the environment variables in `/etc/environment` in your AMI for runners.

sykhro:
This is the standard behaviour when using self-hosted runners with GitHub Actions. You are responsible for cleanup in your own workflows

winwinashwin:
[...]

PR #4263, opened 2024-11-16
feat: Natively support runner pre/post job hooks

winwinashwin:
Adds two optional string arguments to the module
- `runner_hook_job_started`
- `runner_hook_job_completed`

If not empty, registers the hook in `user-data.sh`

Ref: https://github.com/actions/runner/blob/main/docs/adrs/1751-runner-job-hooks.md

Closes: https://github.com/philips-labs/terraform-aws-github-runner/issues/4260
Discussion: https://github.com/philips-labs/terraform-aws-github-runner/issues/3854


winwinashwin:
@npalm Could you please review?

npalm (commented):
Nice contribution! I will check the PR in detail later. But could you check if you can enrich the multi-runner example with the pre / post hook. I would suggest to use one of the amazon linux examples.

winwinashwin:
> Enrich the multi-runner example with the pre / post hook.

Thanks. Added examples in the PR.

npalm:
> @npalm Anything I can help with to get this merged? Thanks for your time.

Not much, clear my calendar. I started last week from the bottom of the PR list. Do my best to check the PR this week.

modules/multi-runner/variables.tf
```diff
@@ -78,6 +78,8 @@ variable "multi_runner_config" {
       cloudwatch_config                       = optional(string, null)
       userdata_pre_install                    = optional(string, "")
       userdata_post_install                   = optional(string, "")
+      runner_hook_job_started                 = optional(string, "")
```
npalm, line 81: can you update the documentation section for this variable please?

examples/multi-runner/templates/runner-configs/linux-x64-ubuntu.yaml
```diff
@@ -52,3 +52,12 @@ runner_config:
       prefix_log_group: true
       file_path: /opt/actions-runner/_diag/Runner_**.log
       log_stream_name: "{instance_id}/runner"
+  runner_hook_job_started: |
```
npalm, line 55: can you remove the pre post from this example. This to have one example (linux) based without the pre post. With the two other we cover the standard (amazonn linux() as well ubuntu via custom user_data.

modules/multi-runner/variables.tf
winwinashwin: Interesting. I was unaware that this (documentation) needs to be explicitly handled as for the other submodules docs are autogenerated.

Just a thought - If the plan is to move to multi-runner by default, then maybe we should think about making this more clear. Either we should think through autogenerating docs or make it clear to contributors that documentation needs to be handled manually.

winwinashwin:
@npalm Thank you for your review. Updated changes as suggested.

winwinashwin:
@npalm Anything pending here?

npalm:
pushed a change to simply the test script in examples, now it works

image

modules/multi-runner/variables.tf
npalm: You right, most of the docs is generated. Let improve the docs in a seperate PR. We moving the repo first before moving modules around.

examples/multi-runner/templates/runner-configs/linux-x64.yaml
```diff
@@ -29,3 +29,12 @@ runner_config:
...
     delay_in_seconds: 180
+  runner_hook_job_started: |
+    echo "Running pre job hook as \$(whoami)"
+
+    # Clean github workspace from previous runs
+    if [[ -n "\$GITHUB_WORKSPACE" ]]; do
+      rm -rf "\$GITHUB_WORKSPACE"
+    done
+  runner_hook_job_completed: |
+    echo "Running post job hook as \$(whoami)"
```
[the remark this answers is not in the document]
npalm: I have simplified the script, by the way if and do are a mismatch :)

npalm (approved):

examples/multi-runner/templates/runner-configs/linux-x64.yaml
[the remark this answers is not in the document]
winwinashwin: Oops, lost my presence of mind a little over there. Thanks for the changes!

Issue #4137, opened 2024-09-17
Output for syncer binary location is incorrect


driskell:
https://github.com/philips-labs/terraform-aws-github-runner/blob/v5.10.4/outputs.tf#L28

It appears to have a variable that doesn't exist but it is also a literal not a variable. I think it is meaning:

```
location         = "s3://${module.runner_binaries[0].bucket.id}/${module.runner_binaries[0].runner_distribution_object_key}"
```

npalm:
Would you have time to make a fix?

PR #4274, opened 2024-11-20
fix: Incorrect syncer binary location in tf outputs

winwinashwin:
The output location for syncer binary was not declared correctly in tf outputs.

Fixes: #4137


npalm (approved):
@winwinashwin thx!

PR #4276, opened 2024-11-20
fix!: remove deprecated runners_scale_up_Lambda_memory_size as it breaks CDKTF

stefanrusu-loctax:
Hi. Thanks for this module.

I come from a CDKTF shop. CDKTF basically has a code generator based on the HCL code which generates the appropriate objects and types.

TypeScript uses camel case as default convention for symbols so that code generator produces invalid code as both the current `runners_scale_up_lambda_memory_size` and the legacy `runners_scale_up_Lambda_memory_size` are converted to the same camel cased string and the TypeScript compiler rejects that code when imported as it contains duplicate symbols.

Removing this deprecated variable enables the CDKTF users who run into this problem to successfully use this module.

This is a breaking change but I don't know for how long you do keep deprecated vars.

### DIFF
--- a/variables.deprecated.tf
+++ b/variables.deprecated.tf
@@ -21,14 +21,6 @@ variable "enable_event_rule_binaries_syncer" {
   }
 }
 
-
-# tflint-ignore: terraform_naming_convention
-variable "runners_scale_up_Lambda_memory_size" {
-  description = "Memory size limit in MB for scale-up lambda."
-  type        = number
-  default     = null
-}
-
 # tflint-ignore: terraform_unused_declarations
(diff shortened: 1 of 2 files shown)


variables.deprecated.tf
```diff
@@ -21,14 +21,6 @@ variable "enable_event_rule_binaries_syncer" {
   }
 }
 
-
-# tflint-ignore: terraform_naming_convention
```
npalm, line 25: This should remove as part of breaking release. I am planning one soon to remove all deprecated vars. And remove the fifo queues as well.

npalm (approved):

PR #4278, opened 2024-11-22
feat(packer): Allow passing in an instance_profile and custom shell scripts that run after the runner installation

bshelton229:
We ran into a couple obstacles when building our custom AMIs. One was needing to pass a custom instance profile to packer in order to get AWS IAM credentials for pre-pulling some ECR images. The other was wanting to run some custom build scripts after the agent had already been installed.

One of our primary reasons for the custom script after agent installation was to write the `ACTIONS_RUNNER_HOOK_JOB_STARTED=` environment variable to the `/opt/actions-runner/.env` file pointing to some cleanup scripts we lay down on disk during AMI building. This may be taken care of if https://github.com/philips-labs/terraform-aws-github-runner/pull/4263 ends up merging, but we have another couple smaller things we're doing to configure the agent installation directory.

I noticed what I think was a missing `instance_type` override variable in the windows-core-2022 file. I'm not sure if that was an oversight, or if it was omitted for a reason. It's been a long time since I've been proficient in Windows, so I wasn't able to test that part of this very well. Because the windows script seemed to already execute after agent installation, I thought it made sense to have the same input variables for all the profiles, and to just run this script directly after the other custom_shell_script that already exists. I thought it safer to make this additional variable instead of moving the custom_shell_script to after agent installation just in case that caused unforeseen issues for others. That being said, we're only using the custom script after agent installation, so moving the existing script to run after agent installation would also solve our problem.

Thanks so much for this project, this has been immensely beneficial for us.

### DIFF
--- a/images/linux-al2023/github_agent.linux.pkr.hcl
+++ b/images/linux-al2023/github_agent.linux.pkr.hcl
@@ -42,6 +42,12 @@ variable "instance_type" {
   default     = "m3.medium"
 }
 
+variable "iam_instance_profile" {
+  description = "IAM instance profile Packer will use for the builder. An empty string (default) means no profile will be assigned."
+  type        = string
+  default     = ""
+}
+
 variable "root_volume_size_gb" {
   type    = number
   default = 8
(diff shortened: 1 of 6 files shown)


npalm:
@bshelton229 can you check why the CI is failing please?

bshelton229:
@npalm thanks for taking a look. It looks like the `shell` provisioner can't take the empty `concat([])`. I was able to run the validator locally in codespaces. Unless you feel the `custom_shell_commands_post_runner_install` hook is valuable, I'm going to refactor this PR to only add the `iam_instance_profile` packer input. We're going to be able to use - https://github.com/philips-labs/terraform-aws-github-runner/pull/4263 to do what we were doing with the hook proposed here. We were using this to register our custom script in the `/opt/actions-runner/.env` file, and that's now directly possible.

npalm (approved):

PR #4333, opened 2025-01-02
fix: Disable interpolation of HEREDOC strings containing runner hook scripts

Peter-Sh:
Interpolation was applied within the user-data.sh context, causing unexpected empty values for variables and other unintended effects.

While this behavior might be useful in certain cases, it requires consistently escaping everything in hook scripts. This approach feels error-prone and unintuitive.

For example, the following configuration will echo an empty string and the date of string interpolation, but not the date of the job's start:
```
  runner_hook_job_started: |
    echo $GITHUB_WORKSPACE
    echo $(date)
```

This PR disables interpolation of the hook script contents when they are being written to the file.

Also the link in user-data.sh to GitHub start/completed docs has been updated.


npalm (approved):
thx, tested and all good

PR #4335, opened 2025-01-06
feat: opt-out ssm parameters for github app

AppliNH:
Hi! 👋

Throughout my recent experiences, we had strong requirements to **not** have secret values in the state. To handle that, I suggest to make optional the creation of SSM parameters as part of the regular flow so it doesn't leak secrets in Terraform state.

We could alternatively use the latest [`ephemeral`](https://developer.hashicorp.com/terraform/language/resources/ephemeral) feature, but I'm not sure everyone is using Terraform 1.10+ atm since it's quite recent. I also think relying on ephemerals should be part of a new breaking release as it won't be compatible with Terraform < 1.10.

Let me know what you think of this!


examples/manual-ssm-parameters/README.md
```diff
@@ -0,0 +1,77 @@
+# Amazon Linux X64 (default)
```
npalm, line 1: Do you intend to keep this example. Having the example is good. But in that case it needs a bit more cleaning.

- Readme needs to be adjusted, with short explanation
- Would call it external managed SSM secrets
- To keep the example convenient, you can create outside the module with terraform the ssm parameters. So the example is simple to use.

npalm (changes requested):
@AppliNH thx for the contribution.

Supporting external maanaged ssm parameters is fine. A view things to consider are

- Update docs see docs dir to explain the usages
- Explain or express via an example how to ajust the policies so the lambda can access the secrets. Lambda roles are exported. 
- I have added a comment related to the variables you introduces. I think it can done in a single app variable. Which avoid the introdduction of new variables. It also makes it easier to validat the variable.

examples/manual-ssm-parameters/README.md
AppliNH: Hi! I've renamed the example and improved the doc for it, lmk what you think

variables.tf
```diff
@@ -34,10 +34,30 @@ variable "enable_organization_runners" {
 variable "github_app" {
   description = "GitHub app parameters, see your github app. Ensure the key is the base64-encoded `.pem` file (the output of `base64 app.private-key.pem`, not the content of `private-key.pem`)."
   type = object({
-    key_base64     = string
-    id             = string
-    webhook_secret = string
+    key_base64     = optional(string)
```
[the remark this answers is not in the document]
AppliNH: Good point, I've refactored in that sense 🙏

AppliNH:
Hi 👋 Really sorry for the delay, I've been terribly busy. I've updated everything, let me know what you think of it!

modules/ssm/variables.tf
```diff
@@ -1,9 +1,21 @@
...
+    id = optional(string)
+    id_ssm = optional(object({
+      arn  = string
+      name = string
+    }))
+    webhook_secret = optional(string)
+    webhook_secret_ssm = optional(object({
+      arn  = string
+      name = string
+    }))
```
npalm, line 18: I think a condition block to verify the varialbe could help. Either the valeu or the ssm locaiton should be provided.

npalm: can you update the description of the variable as well.

npalm (commented):
I am busy with the review. I am fine with the approach. Please can you update the docs as well, see this page: https://github-aws-runners.github.io/terraform-aws-github-runner/configuration/ I think we should add this as major configuration choice. and / or make a remark in the ssm paramater section. 

Please can you also make a note that the provided value has preference ofer the ssm paramater, at least this is the current order in reading the values via the terraform function coalense

I have tested th default example (with pool). Need to test the added example and the multi runners as well.

modules/ssm/variables.tf
AppliNH: Oh yes, I forgot that indeed. I've added them in the last commit 👍

variables.tf
```diff
@@ -32,12 +32,30 @@ variable "enable_organization_runners" {
...
+      arn  = string
+      name = string
+    }))
+    webhook_secret = optional(string)
+    webhook_secret_ssm = optional(object({
+      arn  = string
+      name = string
+    }))
   })
+  default = {}
```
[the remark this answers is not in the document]
AppliNH: done!

AppliNH:
> Please can you also make a note that the provided value has preference ofer the ssm paramater, at least this is the current order in reading the values via the terraform function coalense
@npalm 
Reading the [`coalesce` documentation](https://developer.hashicorp.com/terraform/language/functions/coalesce), the SSM parameter has preference over the the value though.
Do you suggest it should be done otherwise?


Right now it's
```
github_app_parameters = {
    id             = coalesce(var.github_app.id_ssm, module.ssm.parameters.github_app_id)
    key_base64     = coalesce(var.github_app.key_base64_ssm, module.ssm.parameters.github_app_key_base64)
    webhook_secret = coalesce(var.github_app.webhook_secret_ssm, module.ssm.parameters.github_app_webhook_secret)
  }
```

npalm:
Would keep the order in this way

docs/configuration.md
```diff
@@ -24,17 +24,44 @@ The module uses the AWS System Manager Parameter Store to store configuration fo
...
-| `token_path`                        | The path where tokens are stored.                                                                 |
+| Parameter name           | Description                                                                                       |
+| ------------------------ | ------------------------------------------------------------------------------------------------- |
+| `agent_mode`             | Indicates if the agent is running in ephemeral mode or not.                                       |
+| `disable_default_labels` | Indicates if the default labels for the runners (os, architecture and `self-hosted`) are disabled |
+| `enable_cloudwatch`      | Configuration for the cloudwatch agent to stream logging.                                         |
+| `run_as`                 | The user used for running the GitHub action runner agent.                                         |
+| `token_path`             | The path where tokens are stored.                                                                 |
+
+### Note regarding GitHub App secrets provisioning in SSM
```
AppliNH, line 39: Added a note here @npalm wdyt?

docs/configuration.md
```diff
@@ -24,17 +24,44 @@ The module uses the AWS System Manager Parameter Store to store configuration fo
...
+    }
+    id_ssm             = {
+      name = "/your/path/to/ssm/parameter/id"
+      arn = "arn:aws:ssm:eu-west-3:402738628039:parameter/your/path/to/ssm/parameter/id"
+    }
+    webhook_secret_ssm = {
+      name = "/your/path/to/ssm/parameter/webhook-secret"
+      arn = "arn:aws:ssm:eu-west-3:402738628039:parameter/your/path/to/ssm/parameter/webhook-secret"
+    }
+  }
```
npalm, line 59:
```suggestion

```
github_app = {
    key_base64_ssm     = {
      name = "/your/path/to/ssm/parameter/key-base-64"
      arn = "arn:aws:<your-region>:<account-id>:parameter/your/path/to/ssm/parameter/key-base-64"
    }
    id_ssm             = {
      name = "/your/path/to/ssm/parameter/id"
      arn = "arn:aws:ssm:<your-region>:<account-id>:parameter/your/path/to/ssm/parameter/id"
    }
    webhook_secret_ssm = {
      name = "/your/path/to/ssm/parameter/webhook-secret"
      arn = "arn:aws:ssm:<your-region>:<account-id>:parameter/your/path/to/ssm/parameter/webhook-secret"
    }
  }
```

File: examples/external-managed-ssm-secrets/README.md
Comment (npalm) [inline examples/external-managed-ssm-secrets/README.md:1]: please only ref to the example file like other examples does. Only add the follwoing line:

--8<-- "examples/external-managed-ssm-secrets/README.md"

File: examples/external-managed-ssm-secrets/ssm.sh [hunk 7]
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
npalm, line 1: @AppliNH I have created a simple script to test the example a bit easier. Please have a look.

examples/external-managed-ssm-secrets/main.tf
```diff
@@ -0,0 +1,48 @@
+locals {
```
npalm, line 1: @AppliNH I have simplified the example and removed some outputs. The one you provided was not working. Please can run a check?

examples/external-managed-ssm-secrets/ssm.sh
AppliNH: Yes good idea lgtm!

examples/external-managed-ssm-secrets/README.md
AppliNH: You mean like I did [here](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4335/files#diff-c2fbf7ed77f92bf94bcc6c4f479949d8fc54fb89755cc4c528485ed306bd90a5) ?

docs/configuration.md
AppliNH: Fixed!

examples/external-managed-ssm-secrets/README.md
AppliNH: Ohh ok I got confused cause I [did do that here](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4335/files#diff-c2fbf7ed77f92bf94bcc6c4f479949d8fc54fb89755cc4c528485ed306bd90a5)

`docs/examples/external-managed-ssm-secrets.md` has

```
--8<-- "examples/external-managed-ssm-secrets/README.md"
```

But I found out I forgot to add it to the index ([just did now](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4335/files#diff-5e297a681a7f0e124ef51134b519ec471a165f08499fbcb724dfa0e58edd08c1R12)), is it ok now?

examples/external-managed-ssm-secrets/main.tf
AppliNH: Checked!

examples/external-managed-ssm-secrets/README.md
```diff
@@ -0,0 +1,117 @@
+# Externally managed SSM secrets
+
+This example is based on the [default setup](../default/README.md), but shows how to use configure runners with already existing SSM parameters that you'd have created manually.
+
+Manually creating the SSM parameters that hold the configuration of your GitHub App avoids leaking critical plain text values in your terraform state and version control system. This is a recommended security practice for handling sensitive credentials.
+
+## Prerequisites
```
AppliNH, line 7: @npalm just refactored the doc a bit here to make it a bit less confusing regarding the various options we have to provision the ssm parameters for the example.

If it looks good to you, feel free to resolve this and I guess we can merge?

examples/external-managed-ssm-secrets/README.md
npalm: Yupz, some proper CI for docs as well some docs is missing. I have also added an entry to the top-level mkdocs.yml

AppliNH: Alright, I had missed that one, thank you for checking!

AppliNH (commented):
I've reviewed and tested it all again, LGTM!
It's provisioned in production for us 👌

npalm (approved):

PR #4341, opened 2025-01-07
fix: Add prefix to the policies which breaks multi runner configuration

prashanthbgoud:


prashanthbgoud:
@npalm could you review?

npalm (commented):
Can you please explain why you adding the prefix, the `aws_iam_role_policy` is adding inline policies to a role. Which means no need to prefix per environment. As long the role is unqique there will be no clash.

prashanthbgoud:
we use https://github.com/philips-labs/terraform-aws-github-runner/tree/main/modules/multi-runner, I see there are some policies for scaleup and scaledown which are created per runner config. if we have a multi-runner config, it creates policies with the same name multiple times.

npalm:
> we use https://github.com/philips-labs/terraform-aws-github-runner/tree/main/modules/multi-runner, I see there are some policies for scaleup and scaledown which are created per runner config. if we have a multi-runner config, it creates policies with the same name multiple times.

The module (multi-runner) creates roles and inside the roles customer managed inline policies. Those policies only exists inside a role. The role already have the prefix.

See here an example of a lambda role (scale-up)
image

PR #4390, opened 2025-01-28
feat: Support GitHub Enterprise Cloud with Data Residency

neethu-p:
Opening up the PR again to address the CI issue - https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4389


neethu-p:
@npalm Looks like the lambda workflows are not getting triggered. Can you check

npalm:
> @npalm The Build seems to be working now! Have modified the test class to handle the new feature

thx, will check asap.

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
```diff
@@ -675,6 +675,310 @@ describe('scaleUp with public GH', () => {
   });
 });
 
+describe('scaleUp with Github Data Residency', () => {
```
npalm, line 678: I should made this remakr before. And happy to merg the PR. But we should find way no not duplicate all the tests. Wondering did you tried to avoid duplicaiton of the tests? 

I think the best way forward is to create a new issue. To simply the test so we can better maintain the module. What is your opnion?

npalm (commented):
Looks good, CI looks fine. I have made one remark. Looking for your opnion.

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
neethu-p: Yes, make seems. I can try to merge the GH cloud tests into one, if that make it simpler. But I would recommend we do it another PR may be if that is okay 😊

npalm (approved):

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
npalm: issue created: #4393

PR #4392, opened 2025-01-28
feat: Add option to let the module manage the webhook secret

npalm:
## Description

This PR is adding the option to let the module manage the webhook secret if not provided. In case the secret is provided nothing is changing. When NOT provided the module will

- Generate a secret
- Manage the GitHub app secret and webhook url
- Trigger a rotation when terraform is called after x days (default 30).


docs/configuration.md
```diff
@@ -14,6 +14,7 @@ To be able to support a number of use-cases, the module has quite a lot of confi
...
 - Disable default labels for the runners (os, architecture and `self-hosted`) can achieve by setting `runner_disable_default_labels` = true. If enabled, the runner will only have the extra labels provided in `runner_extra_labels`. In case you on own start script is used, this configuration parameter needs to be parsed via SSM.
+- Managed vs self-managed webhook secret. The module can manage the webhook secret for you. In that case simply do not provide a value for `github_app.webhook_secret`. If you want to manage the secret yourself, provide a value for `github_app.webhook_secret`. The secret will be managed and a rotation is triggered once running terraform apply again after `github_app.webhook_secret_rotation_days` days. **Important note**: THe managed webhook secret depends on a local-exec (bash) to update the secret in GitNub. It will also update the webhook url.
```
koendelaat, line 17:
```suggestion
- Managed vs self-managed webhook secret. The module can manage the webhook secret for you. In that case simply do not provide a value for `github_app.webhook_secret`. If you want to manage the secret yourself, provide a value for `github_app.webhook_secret`. The secret will be managed and a rotation is triggered once running terraform apply again after `github_app.webhook_secret_rotation_days` days. **Important note**: THe managed webhook secret depends on a local-exec (bash) to update the secret in GitNub. It will also update the webhook url. Note: for automatic rotation of the secret a `terraform apply` is required every `github_app.webhook_secret_rotation_days` days
```

koendelaat (approved):

npalm:
Moving this PR back to draft, I would like to see if there is a better way

PR #4394, opened 2025-02-01
fix: define user agent

npalm:
## Desription

Setting the user-agent to call API call. By default the agent is set to `aws-github-runners` the value can be overriden.

## Notes

cc: @SvanBoxel


npalm:
@rjaegers @koendelaat can one of you quickly theck the PR?

rjaegers (approved):

PR #4407, opened 2025-02-05
chore(ci): update .github/workflows/update-docs.yml to use latest actions/cache

ricky-undeadcoders:
Hi all!

GitHub will be deprecating older versions of `actions/cache` on March 1st, including version `4.0.2` which is used in `.github/workflows/update-docs.yml`. 

This is a quick PR to get that workflow updated to the newest version.

See [GitHub's notice](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down) and the [full list of deprecated versions](https://github.com/actions/cache/discussions/1510) for more info.

### DIFF
--- a/.github/workflows/update-docs.yml
+++ b/.github/workflows/update-docs.yml
@@ -67,7 +67,7 @@ jobs:
         with:
           python-version: 3.x
       - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


npalm (approved):
Thx, LGTM. But weird it was not updated via dependabot.

ricky-undeadcoders:
> Thx, LGTM. But weird it was not updated via dependabot.

Agreed, we had a similar miss at Grafana. If I find anything I'll revisit with related PRs.

PR #4475, opened 2025-03-11
fix(syncer): add default filter with empty prefix due to changes in AWS Provi…

bdandoy:
…der 5.90.0

In 5.90.0 of the AWS provider they are now providing the warning below. This PR keeps the same functionality while getting rid of the warning.

```
No attribute specified when one (and only one) of [rule[0].prefix.<.filter] is required

This will be an error in a future version of the provider
```


npalm:
Thx for the PR. Please can you update the description with what you trying to fix please?

bdandoy:
Description has been updated

npalm (approved):
@bdandoy thx!

PR #4488, opened 2025-03-20
feat: add ami-updater lambda function with tests and config

dgokcin:
## Summary

This PR introduces a new AMI Updater Lambda function with comprehensive configuration, testing, and infrastructure setup.

## Changes

- Added AMIManager class with methods to handle AMI updates
- Added unit tests for AMIManager methods and Lambda handler
- Added configuration management for environment variables
- Added ESLint and TypeScript configuration
- Added Terraform configuration for Lambda infrastructure
- Added IAM policies and CloudWatch permissions
- Added README documentation for the AMI Updater module
- Added package.json with build and test scripts
- Added local run function for testing
- Added Vitest configuration for testing

## Additional Notes

- This is the first time I dive this deep into this codebase  so I probably messed up, thats why the PR is still a draft.

Issue #4460, opened 2025-03-07
Question: How to avoid drift in AMI IDs


dgokcin:
Hello and thank you for this great project. I have the following multi-runner configuration where I use an AMI by [runs-on](https://runs-on.com/) since they are really close to the GitHub Actions runners and are updated regularly. One thing I have noticed is once the terraform module is applied, if the ami-id is deleted by the ami provider, the builds get stuck. And I need to re-apply the module to update the id in the launch template.

Is there a way to make this more dynamic or easier to maintain?

Thanks!

```yaml
        #... multi_runner_config  
        enable_runner_binaries_syncer = true
        ami_owners                    = ["135269210855"] # runs-on ami account id

        userdata_template             = "./templates/user-data-runs-on.sh" #userdata for runs-on
        ami_filter = {
          name  = ["runs-on-v2.2-ubuntu24-full-x64-*"],
          state = ["available"]
        }

```

npalm:
The only option the module currently supportss is to refer the AMI ref via an SSM paramater. In that case the one in the launch template is ingored. Terraform apply will still drift, but the AMI can be managed without runnig terraform.

dgokcin:
I have written a lambda function that triggers daily and updates the ami ref in the ssm parameter. As you said, there is still drift but at least I don't have to remember updating AMI IDs every now and then.

@npalm do you think that this is a good approach and could be added to the base module? I can work on a PR if you think this is a good addition to the module capabilities.

npalm:
I think we should make, also have a look at https://github.com/github-aws-runners/terraform-aws-github-runner/discussions/4464

- May we can move the AMI id completly to SSM.
- Otherwise would be great if we find a way to avoid the drift if SSM is used in the Lambda.

Can you eloborate a bit more what the Lambda function does. Adding it to the base module could be fine. But let's also see if we can work on simply it all.

@claytonolley any thoughts?

claytonolley:
Some background on my use case:

- I build custom AMI's weekly using packer and they automatically update some SSM parameters marking them as "dev" runners
- I have GitHub actions running periodic tests on these "dev" runners as well as some dev only workflows in other repos
- If at the end of the week (prior to the next scheduled AMI build), all these tests are successful, then I have another action that will promote the "dev" SSM parameter value to the "prd" parameter which is what the bulk of my repos will use for their runners.
[...]

PR #4517, opened 2025-03-31
feat: migrate launch template to use SSM for AMI lookup

npalm:
# Description

This PR migrates the launch template to use SSM to foor looking up AMI ID. This PR is inteded to migrate to use SSM for AMI lookups. In later PR we will remove logic from scale-up and remove old parateters.

By default, the module still uses the same AMI filters, but difference is the resulting AMI ID is stored in SSM instead of the launch template. As alternative the module user can provide `ami_id_ssm_parameter_arn` pointing to an SSM parameter containing the AMI ID to be used (data type aws:ec2:image). 

When using a custom managed AMI this will avoid drift in terraform when `ami.id_ssm_parameter_arn` is used instead of the deprecated `ami_id_ssm_parameter_name`. Another way to avoid the drift is to ensure the filter is poiting to the same image. See also #4460

In PR #4488 and updater via Lambda is introduced. I think we should firt think how we can better support the AMI id in the core. What this PR is supposed todo. NExt the update can be an nice way to auto update AMI's without running terraform.

The PR also refactor the variables for `ami_*` and move the to a dedicated object `ami`. The old ones remains till the next major release. Deprecattion warnings are added to the output.

## Notes

- Adjuest multi runner examples
- Test with multi runners
- Test with default runners + pool

## Fix

- Deprecation warning inline_poilicy for pool


modules/runners/main.tf
```diff
@@ -140,7 +162,7 @@ resource "aws_launch_template" "runner" {
   }
 
   instance_initiated_shutdown_behavior = "terminate"
-  image_id                             = data.aws_ami.runner.id
+  image_id                             = "resolve:ssm:${local.ami_id_ssm_module_managed ? aws_ssm_parameter.runner_ami_id[0].arn : var.ami_id_ssm_parameter_arn}"
```
dgokcin, line 165: is there a reason behind using an arn instead of the parameter name 🤔

npalm: The only thought it would be more specific. I just pushed an example that works with a public parameter as well.

npalm:
Updated:

- Add to multi runner example, the usages of a public aws managed paramater.
- Added deprecation output

Some more todo work on my list
- Crack my had on introducing an object for AMI related settings.
- Update docs

@dgokcin thanks for already having a look. Would you have time to review the PR once I done with the draft. Early feedback is always welcome.

modules/runners/main.tf
npalm: Do you think a name will works as well. I thought using the arn makes it also simple to setup IAM policies.

npalm: I did inentially intoduce a new parameter, since the SSM parameter requires the correct data type. The old parameter is handled by the lambda. The new one fully by the launch template. I want to get rid of the old one in a breaking change release.

npalm:
@dgokcin would you have time to test hte PR?

dgokcin:
> @dgokcin would you have time to test hte PR?

can try it on monday :)

koendelaat (approved):

npalm:
> > @dgokcin would you have time to test hte PR?
> 
> can try it on monday :)

did you had time to run a quikc check?

dgokcin (approved):
it looks good to me! I think this is a good step towards my initial problem with drift in ami names!

PR #4531, opened 2025-04-11
feat: added missing launch template tag_specifications

konkerama:
Hello,

Currently the launch template specification adds tags only on resource types of `instance` and `volume` this leaves the other 2 supported resource types (`spot-instances-request` and `network-interface`) untagged which by extensions does not produce accurate data on Cost Explorer.

This pr fixes this by adding the missing `tag_specifications` configurations.


npalm (changes requested):
For some reason the main.tf in pool is not formatted propertly. Please can you run a terraform fmt -recursive on the root.

Change looks good to me

npalm:
sorry messed up, creating a fresh PR

PR #4544, opened 2025-04-16
feat: added missing launch template tag_specifications

npalm:
copy from #4531

Currently the launch template specification adds tags only on resource types of `instance` and `volume` this leaves the other 2 supported resource types (`spot-instances-request` and `network-interface`) untagged which by extensions does not produce accurate data on Cost Explorer.

This pr fixes this by adding the missing `tag_specifications` configurations.


konkerama:
@npalm thank you very much :)

PR #4562, opened 2025-04-22
fix: not try to tag spot requests when on-demand is set

### LINKED ISSUE #4561: spot-instances-request tags breaks on-demand instances
Hi.
 Today we upgraded the module from version 6.4.3 to 6.5.1. Our groups based only on on-demand instances got stuck right after the upgrade. The scale-up lambda silently fails with this error:
"You cannot specify tags for Spot instances requests if there are no Spot instances requests being created by the request."

Below you can see the whole lambda logs:

[logs.txt](https://github.com/user-attachments/files/19853170/logs.txt)

How can we avoid putting those tags in the launch template when running only on-demand instances?

Thanks!

npalm:
fix #4561

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -206,18 +206,21 @@ resource "aws_launch_template" "runner" {
     )
   }
 
-  tag_specifications {
-    resource_type = "spot-instances-request"
-    tags = merge(
-      local.tags,
-      {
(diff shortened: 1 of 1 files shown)


npalm (commented):
Tested with both spot and not spot

koendelaat (approved):

PR #4564, opened 2025-04-24
test: find vitest setup file by absolute path

iainlane:
I've been struggling with running the tests in Neovim. I was getting

```console
FAIL  functions/control-plane/src/github/client.test.ts [ functions/control-plane/src/github/client.test.ts ]
Error: Failed to load url <root>/../aws-vitest-setup.ts (resolved id: <root>/.../aws-vitest-setup.ts). Does the file exist?
```

(but with absolute paths)

Turns out this was because the tests couldn't be run using `yarn vitest --config-file=lambdas/... path/to/test.test.ts`.

When we resolve the paths using `node:path.resolve()` then this works.

### DIFF
--- a/lambdas/functions/ami-housekeeper/vitest.config.ts
+++ b/lambdas/functions/ami-housekeeper/vitest.config.ts
@@ -1,9 +1,11 @@
+import { resolve } from 'path';
+
 import { mergeConfig } from 'vitest/config';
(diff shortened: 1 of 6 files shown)


npalm (approved):

npalm:
thx

PR #4585, opened 2025-05-14
fix: only tag spot requests if no on-demand fallback

pwo3:
Hi,

This PR prevents tagging spot instance requests when an on-demand fallback is configured.

It addresses [my comment on this issue](https://github.com/github-aws-runners/terraform-aws-github-runner/issues/4561#issuecomment-2875651681).

The previous fix was working only if the `instance_target_capacity_type` is set to `on-demand` but not in case of on-demand fallback from a spot request.

This approach isn’t ideal but I didn’t find a cleaner way to cancel the tagging directly in `lambdas/functions/control-plane/src/aws/runners.ts` when the on-demand fallback is triggered.

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -206,6 +206,10 @@ resource "aws_launch_template" "runner" {
     )
   }
 
+  # We avoid including the "spot-instances-request" tag_specifications block when on_demand_failover_for_errors is defined,
+  # because when using on-demand fallback, the spot instance request resource is not created and thus the tags would not apply.
+  # Additionally, tagging spot requests via the CreateFleetCommand in the Lambda function does not work as expected,
+  # so we rely on Terraform to manage these tags only when spot is exclusively used without on-demand failover.
   dynamic "tag_specifications" {
     for_each = var.instance_target_capacity_type == "spot" && var.enable_on_demand_failover_for_errors == null ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
     content {
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -207,7 +207,7 @@ resource "aws_launch_template" "runner" {
   }
 
   dynamic "tag_specifications" {
-    for_each = var.instance_target_capacity_type == "spot" ? [1] : [] # Include the block only if the value is "spot"
+    for_each = var.instance_target_capacity_type == "spot" && var.enable_on_demand_failover_for_errors == null ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
     content {
       resource_type = "spot-instances-request"
       tags = merge(
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -211,7 +211,7 @@ resource "aws_launch_template" "runner" {
   # Additionally, tagging spot requests via the CreateFleetCommand in the Lambda function does not work as expected,
   # so we rely on Terraform to manage these tags only when spot is exclusively used without on-demand failover.
   dynamic "tag_specifications" {
-    for_each = var.instance_target_capacity_type == "spot" && var.enable_on_demand_failover_for_errors == null ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
+    for_each = var.instance_target_capacity_type == "spot" && length(var.enable_on_demand_failover_for_errors) == 0 ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
     content {
       resource_type = "spot-instances-request"
       tags = merge(


modules/runners/main.tf
```diff
@@ -207,7 +207,7 @@ resource "aws_launch_template" "runner" {
   }
 
   dynamic "tag_specifications" {
-    for_each = var.instance_target_capacity_type == "spot" ? [1] : [] # Include the block only if the value is "spot"
+    for_each = var.instance_target_capacity_type == "spot" && var.enable_on_demand_failover_for_errors == null ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
```
npalm, line 210: This will solve the problem, but will avoid to tag the spot request if ondeamdn failover is active. A better place would be the lambda in my point of view.

[const instancesOnDemand = await createRunner({](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/57fce77e7e633abc5ca452c2f18b82372cc98e4c/lambdas/functions/control-plane/src/aws/runners.ts#L190)

Would you have time to provide a fix in the lambda?

pwo3: I don’t think we can apply a simple fix on the Lambda

Since the tags are defined in the launch template, it’s not possible to override them using the CreateFleetCommand

We could remove the spot-instances-request tags from the launch template and set them only in the Lambda’s CreateFleetCommand, but in that case, we lose access to the tags defined in the Terraform configuration

I don’t see a simple solution for now, do you have any thoughts on this?

Thanks

r-bk: I arrived to the same conclusion when I investigated a workaround.

r-bk:
@pwo3 @npalm 
Any updates on this PR?
Are there any outstanding blockers to merge it?

npalm:
I will have a look at the PR asap

modules/runners/main.tf
npalm: I think we should consider to move the tagging spot request the runner.ts instead of setting it in the template. In that case we can do it properly based if a spot instance is requested or not. Already some tags are set here. Would you like to give it a shot?

[TagSpecifications: [](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/57fce77e7e633abc5ca452c2f18b82372cc98e4c/lambdas/functions/control-plane/src/aws/runners.ts#L272-L283)

pwo3: @npalm I tried this approach, but we would lose all the tags defined in the Terraform module, only those tags would be applied
```
  const tags = [
    { Key: 'ghr:created_by', Value: runnerParameters.numberOfRunners === 1 ? 'scale-up-lambda' : 'pool-lambda' },
    { Key: 'ghr:Type', Value: runnerParameters.runnerType },
    { Key: 'ghr:Owner', Value: runnerParameters.runnerOwner },
  ];
```

However, I can try fetching the tags from the launch template first and then applying them to the spot-instances-request via the TagSpecifications

pwo3: I pushed the changes, but I’m not really sure how to test it in real conditions

npalm: I will run some tests to check spot request are correctly tagged. The case spot is not available is not testable as far I know.

npalm (changes requested):
I have tested the PR, the code was not working do to the falt map was creating duplicated tags and missing describe permission for the lanunch template.

After making some changes the lambda was working. However no tags on the spot request. Changed several things. But did not got it working at all. Really strange, debug showed clearly the correct elements in the TagSpecification. 

After updating to main, I got my tag on the spot request back. Maybe we should revert back to the previous approach. And make a not in the terraform code tht this should not be the place but tagging via the spotfleetrequest was not working at all.

pwo3:
Hi @npalm, can we merge this PR if everything is OK with you? Thanks

modules/runners/main.tf
```diff
@@ -206,8 +206,12 @@ resource "aws_launch_template" "runner" {
     )
   }
 
+  # We avoid including the "spot-instances-request" tag_specifications block when on_demand_failover_for_errors is defined,
+  # because when using on-demand fallback, the spot instance request resource is not created and thus the tags would not apply.
+  # Additionally, tagging spot requests via the CreateFleetCommand in the Lambda function does not work as expected,
+  # so we rely on Terraform to manage these tags only when spot is exclusively used without on-demand failover.
   dynamic "tag_specifications" {
-    for_each = var.instance_target_capacity_type == "spot" ? [1] : [] # Include the block only if the value is "spot"
+    for_each = var.instance_target_capacity_type == "spot" && var.enable_on_demand_failover_for_errors == null ? [1] : [] # Include the block only if the value is "spot" and on_demand_failover_for_errors is not enabled
```
npalm: The default for `enable_on_demand_failover_for_errors` is an empty array. Means no failover to demand. Either we check here for an empty array or ensure the lambda can handle null and set null as default.

npalm (changes requested):
Apply the change to the default example is trigger a change. This bcase the default for enable_on_demand_failover_for_errors is `[]`. Can you include an empty array check in the if.

Sorro for the long response time on the PR

modules/runners/main.tf
pwo3: Oh yes sorry, didn't see the default value for enable_on_demand_failover_for_errors

I have updated the condition

npalm (approved):

npalm (approved):

claytonolley:
This change just bit us in the rear, now finops is mad our spot requests aren't being tagged. :( 

I guess our only recourse is to turn off on-demand fallback?

PR #4613, opened 2025-06-16
fix: Dynamic set SSM-paramater tier

ScottGuymer:
GitHub seems to have made a change to encoded_jit_config that has pushed it over the 4kb limit for standard tier of SSM.
 
We now must always use advanced tier of SSM config for encoded_jit_config.


lambdas/libs/aws-ssm-util/src/index.ts
```diff
@@ -30,6 +31,7 @@ export async function putParameter(
       Value: parameter_value,
       Type: secure ? 'SecureString' : 'String',
       Tags: options.tags,
+      Tier: tier,
```
npalm: Would suggest we keep the responsiblity to decide whic paramter tear in the putParamter function. 

Pseudo code:
```
// maybe limit to 4000 instead of 4096 to stay on the safe side
const tier = Buffer.byteLength(paramater_value 'utf8') > 4000 : ParameterTier.ADVANCED : ParameterTier.STANDARD,
logger.debug(....)
await client.send(...))
```

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -450,12 +450,19 @@ async function createJitConfig(githubRunnerConfig: CreateGitHubRunnerConfig, ins
...
 
     // store jit config in ssm parameter store
+    // Must used advanced tier for SSM as encoded_jit_config can be > 4kb
     logger.debug('Runner JIT config for ephemeral runner generated.', {
       instance: instance,
     });
-    await putParameter(`${githubRunnerConfig.ssmTokenPath}/${instance}`, runnerConfig.data.encoded_jit_config, true, {
-      tags: [{ Key: 'InstanceId', Value: instance }],
-    });
+    await putParameter(
```
npalm: The module creates parameter on 2 or 3 palces, would be better to move this logic to put paramter and avoid larger are casuing errors.

npalm (changes requested):
@ScottGuymer thax for creating. I just checked the multi-runner example in which the x64 linux is using JIT configuration as well. And indeed we are just on the safe side with paramaters from about 4000 bytes.

I would prefer we solve the prbolem in the warpper function in a generic way there were the lambda is sotring pramaters. Initialy we stayed away from adavanced by to avoid costs. As long the paramters are correctly cleared costs should not be a concert. With about 50K of ephemeral runners this should stay under the $5 a month.

Can you also make a note to this (https://github-aws-runners.github.io/terraform-aws-github-runner/configuration/) section that dynamaic generated paramaters are using based on size are stored in the advanced tier. To avoid costs it is important the ssm housekeeper is running, which is enabled by default but can be disabled.

Tested a deployment (multi-runner), the x64 linux runner is using JIT. Thos paramaters got indeed now created as ADVANCED and runners got started.

npalm:
In case parameter size is exceeded the log wil show a log line like:

```
{"level":"WARN","message":"Ignoring error: ValidationException: Standard tier parameters support a maximum parameter value of 4096 characters. To create a larger parameter value, upgrade the parameter to use the advanced-parameter tier. For more information, see [https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html"](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html%22)"}
```

lambdas/libs/aws-ssm-util/src/index.ts
ScottGuymer: Good improvement, updated to this implementation.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
ScottGuymer: done

ScottGuymer:
Updates made, ready for re-review.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -450,6 +450,7 @@ async function createJitConfig(githubRunnerConfig: CreateGitHubRunnerConfig, ins
     metricGitHubAppRateLimit(runnerConfig.headers);
 
     // store jit config in ssm parameter store
+    // Must used advanced tier for SSM as encoded_jit_config can be > 4kb
```
npalm: Delete comment
```suggestion
```

npalm:
Quick test based on multi runner examples:
image

npalm (approved):

claytonolley:
I wish I would have seen this earlier, but I was only looking on the Issues page. 😭

Anyway, after a lot of fumbling around, I found that in AWS you can go into Parameter Store settings and turn on "Intelligent-Tiering" which will automatically make an Advanced Tier parameter if the size is over 4k.

ScottGuymer:
Yeah we also considered using intelligent tiering as an option when setting the value. 

But one of our team-mates said they had experienced issues with this in the past so we erred on the side of caution and just making it work OOTB.

andrewdibiasio6:
>Parameter Store settings and turn on "Intelligent-Tiering"

Worked for me to get things unblocked.

PR #4623, opened 2025-06-17
fix(ami-housekeeper): don't delete referenced AMIs in default config

### LINKED ISSUE #4571: AMI housekeeper deletes all AMIs
Since the migration of the launch template using SSM for AMI lookup, the AMI housekeeper function will delete all existing AMIs.

The filter [here](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/lambdas/functions/ami-housekeeper/src/ami.ts#L198) uses `ami-id` instead of `ami_id`, the latter of which is in the parameter name. As a result, the AMI in-use will be mistakenly considered not-in-use, thus deleted.

In addition, the `ssmParameterNames` in `options` parameter are not used. Perhaps the specified names should be used in the filter.

iainlane:
In 472cc5f6175d79e26260e0ed7a9cac7fc299411f the default config was migrated to use SSM for AMI lookup. A parameter is created which stores a reference to the AMI. By default, this parameter is called `${var.ssm_paths.root}/${var.ssm_paths.config}/ami_id`.

The housekeeper is a process that looks for AMIs which can be deleted because they're no longer used. It does this in a couple of ways:

1. Check the launch template for the AMI ID.
2. Check the SSM parameter.
3. Apply a threshold to not delete AMIs that are too new, according to the config.

The problem is that we were looking for SSM parameters like this:

```typescript
const ssmParams = await ssmClient.send(
  new DescribeParametersCommand({
    ParameterFilters: [
      {
        Key: "Name",
        Values: ["ami-id"],
        Option: "Contains",
      },
    ],
  }),
);
```

i.e. we were looking for parameters which contain the hardcoded string `ami-id`. This is different to the new default of `ami_id`. So we weren't considering the right AMIs to be in use.

What would be a better approach would be to reference the values dynamically. This means resolving from the template, and handling the passed-in options, if there are any. We're documenting that we support wildcards, so also support that here too.

The default value in the launch template became `resolve:ssm:<id or AMI>`, so we need to make sure to ask EC2 to resolve for us when looking up the template. In that way we get the actual AMI ID rather than the alias.

This can be a bit challenging to understand, so the comments are improved.

Comprehensive tests are added to try to ensure this all works as expected.

Closes: #4571

### DIFF
--- a/lambdas/functions/ami-housekeeper/src/ami.ts
+++ b/lambdas/functions/ami-housekeeper/src/ami.ts
@@ -217,11 +217,15 @@ async function getAmiInLatestTemplates(options: AmiCleanupOptions): Promise<(str
 
   // Discover launch templates, optionally filtered by specific names. If no
   // names provided, this will return all launch templates in the account
+  logger.debug('Describing launch templates', {
+    launchTemplateNames: options.launchTemplateNames,
+  });
   const launchTemplates = await ec2Client.send(
     new DescribeLaunchTemplatesCommand({
       LaunchTemplateNames: options.launchTemplateNames,
     }),
   );
+  logger.debug('Found launch templates', { launchTemplates });
 
   // For each template, fetch the default version and resolve any SSM aliases.
   const amiIdsNested = await Promise.all(
@@ -236,10 +240,12 @@ async function getAmiInLatestTemplates(options: AmiCleanupOptions): Promise<(str
         }),
       );
 
+      logger.debug('Found launch template versions', { versionsResp });
       return (versionsResp.LaunchTemplateVersions ?? []).map((v) => v.LaunchTemplateData?.ImageId);
     }),
   );
 
+  logger.debug('Found AMIs in launch templates', { amiIdsNested });
   return amiIdsNested.flat();
 }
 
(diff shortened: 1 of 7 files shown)


lambdas/functions/ami-housekeeper/src/ami.ts
```diff
@@ -158,61 +183,120 @@ async function deleteSnapshot(options: AmiCleanupOptions, amiDetails: Image, ec2
...
+    (launchTemplates.LaunchTemplates ?? []).map(async (template) => {
+      const versionsResp = await ec2Client.send(
         new DescribeLaunchTemplateVersionsCommand({
-          LaunchTemplateId: launchTemplate.LaunchTemplateId,
-          Versions: ['$Default'],
+          LaunchTemplateId: template.LaunchTemplateId,
+          Versions: ['$Default'], // Only check the default version
+          // This means that references like `resolve:ssm:<parameter arn>` are
+          // dereferenced.
+          ResolveAlias: true,
```
npalm, line 239: @iainlane this seems to require the GetParamaters permissions. I will push the change to your branch.

npalm (dismissed):

npalm:
@iainlane tested the PR. Looks all good. I have update the prebuilt example to for testing and pushed this change and the missing permission to your branch.

npalm (approved):

PR #4627, opened 2025-06-22
chore(ci): Add sigstore bundle to release

npalm:
Adding sigstore bundle to release, see https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,7 +84,7 @@ jobs:
         run: |
           # rename attest bundle to github-aws-runners-terraform-aws-github-runner-attestation-$attestation-id.sigstore
           # OpenSSF expects the attestation bundle to be named in this format (*.sigstore)
-          SIGSTORE_BUNDLE="github-aws-runners-terraform-aws-github-runner-attestation-${ATTESTATION_ID}.sigstore"
+          SIGSTORE_BUNDLE=$RUNNER_TEMP/github-aws-runners-terraform-aws-github-runner-attestation-${ATTESTATION_ID}.sigstore
           mv ${ATTESTATION_BUNDLE} $SIGSTORE_BUNDLE
           if [ -z "$SIGSTORE_BUNDLE" ]; then
             echo "No attestation bundle found, skipping attachment."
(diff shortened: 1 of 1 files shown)


rjaegers (approved):

.github/workflows/release.yml
```diff
@@ -74,3 +74,20 @@ jobs:
...
+        if: ${{ steps.release.outputs.releases_created == 'true' }}
+        env:
+          ATTESTATION_BUNDLE: ${{ steps.attest.outputs.bundle-path }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG_NAME: ${{ steps.release.outputs.tag_name }}
+          ATTESTATION_ID: ${{ steps.attest.outputs.attestation-id }}
+        run: |
+          # rename attest bundle to github-aws-runners-terraform-aws-github-runner-attestation-$attestation-id.sigstore
+          # OpenSSF expects the attestation bundle to be named in this format (*.sigstore)
+          SIGSTORE_BUNDLE=$RUNNER_TEMP/github-aws-runners-terraform-aws-github-runner-attestation-${ATTESTATION_ID}.sigstore
```
rjaegers, line 87: Only one minor comment.

If I interpret the [OpenSSF Scorecards documentation](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases) correctly only a SLSA provenance file (*.intoto.jsonl) will yield the maximum score of 10.

This data is available in the attached bundle, but part of a base64 encoded envelope.

In other words, while you have signed the release asset _and_ have the proper data available, you'll still not get the maximum score this way I suspect. That might be acceptable, so FYI.

npalm: Any clue how to obtain the intoto file?

rjaegers: Yes, you should be able to extract it from the data in the bundle-path.

Something like (untested):

`cat bundle-file.json | jq -c '.[] | .attestation.bundle.dsseEnvelope | select(.payloadType == "application/vnd.in-toto+json").payload' | base64 -d | jq .`

npalm: Will fix this in a new PR

PR #4654, opened 2025-07-02
fix: Safer chown call in runner scripts for custom AMIs

nmodi1:
Updated the install-runner and start-runner scripts to call `chown` on explicit path instead of just using the current directory. 

The current script will work for most users, however, this change will make it more safer and avoid risk of unintended changes. For example, if the `cd /opt/actions-runner` call in the template file (`images/start-runner.sh`) is missed, the script will change owner of the root directory (`/`) instead.

### DIFF
--- a/modules/runners/templates/install-runner.sh
+++ b/modules/runners/templates/install-runner.sh
@@ -70,5 +70,5 @@ if [[ "$os_id" =~ ^ubuntu.* ]]; then
 fi
 
 echo "Set file ownership of action runner"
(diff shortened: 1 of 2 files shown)


nmodi1:
There are a few other places that use the current directory (`./`) in the start-runner script. I can update them to use an explicit path (or an env variable) as well if needed?

npalm (approved):

npalm (approved):
@nmodi1 thx for your contribution

PR #4655, opened 2025-07-03
feat: Add release version to lambda conext for logging

npalm:
This pull request introduces changes to improve release management and enhance the lambda's loggin. 

Release updates lambda `package.json` file with the release version in the release PR. See example: https://github.com/cloud-scouts/terraform-aws-github-runner/pull/15
image

The version of the lambda is added to the log context.
(diffhunk://#diff-8e1eeb7f895c1c12c548abcc3b9bcf66106ea5366250af44a364db81b5384446R40)`)
image


.release-please-manifest.json
npalm, line 1: This file is mandatory when using a release manifest

lambdas/libs/aws-powertools-util/src/logger/index.ts
```diff
@@ -1,17 +1,34 @@
...
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+
 const defaultValues = {
   region: process.env.AWS_REGION,
   environment: process.env.ENVIRONMENT || 'N/A',
 };
 
+function getReleaseVersion(): string {
+  let version = 'unknown';
+  try {
```
npalm, line 18: Inention is here to avoid an excetion can crash the lambda.

lambdas/libs/aws-powertools-util/src/logger/logger.test.ts
```diff
@@ -29,6 +30,31 @@ const context: Context = {
   },
 };
 
+const versionFilePath = path.resolve(__dirname, 'version.json');
```
koendelaat: What is the purpose of this version file?
I only can find deletions of the file, not the creation.

lambdas/libs/aws-powertools-util/src/logger/index.ts
```diff
@@ -1,17 +1,34 @@
...
+    logger.debug(`Failed to read package.json for version: ${(error as Error)?.message ?? 'Unknown error'}`);
+  }
+  return version;
+}
+
 function setContext(context: Context, module?: string) {
   logger.addPersistentLogAttributes({
     'aws-request-id': context.awsRequestId,
     'function-name': context.functionName,
+    version: getReleaseVersion(),
```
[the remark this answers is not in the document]
npalm: Looks a good catch, but as far I know the context is only called once via pwoertools.

koendelaat (approved):

lambdas/libs/aws-powertools-util/src/logger/index.ts
```diff
@@ -20,6 +37,7 @@ function setContext(context: Context, module?: string) {
     childLogger.addPersistentLogAttributes({
```
iainlane, line 37: This isn't new in this PR so doesn't _matter_ for the review, but it caught my eye... Why do we have to store the child loggers? Should be possible to arrange that only the root logger is ever seen and so we don't need to repeat the attribute addition, right?

lambdas/libs/aws-powertools-util/src/logger/logger.test.ts
```diff
@@ -29,6 +30,31 @@ const context: Context = {
...
+    fs.unlinkSync(versionFilePath);
+  }
+});
+
+afterEach(() => {
+  // Clean up after each test
+  if (fs.existsSync(versionFilePath)) {
+    fs.unlinkSync(versionFilePath);
+  }
+});
```
iainlane, line 56: I'd be a bit concerned about messing with the real `package.json` file here,  what happens if tests get run in parallel, and the reinitialisaion which seems complex. I would consider doing this with dependency injection. Make `getReleaseVersion()` take a directory, like this:

```typescript
function getReleaseVersion(packageDir: string = __dirname): string {
  // ...
}
```

and then each test could make its own temp dir and set it up how it wants.

It would make sense to still have one integration test which uses the real file under `__dirname`, to ensure that there's always a version set in there.

lambdas/libs/aws-powertools-util/src/logger/index.ts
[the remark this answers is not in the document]
iainlane: Seems like a fair defensive thing to do,

```typescript
let releaseVersion: string | null = null;


function setContext(context: Context, module?: string) {
  if (releaseVersion === null) {
    releaseVersion = getReleaseVersion();
  }

  // ...
}
```

 but yeah probably minor, does look like it's only called once per function.

PR #4658, opened 2025-07-10
feat: Add workflow URL to EC2 instance tags

roberto-civitas:
Add the `workflow_url` to the instance tags so that it's easier to identify which build is being run by the instance.

The property is passed along from the webhook lambda, as part of the SQS event.

### DIFF
--- a/lambdas/functions/control-plane/src/aws/runners.d.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.d.ts
@@ -43,4 +43,5 @@ export interface RunnerInputParameters {
   amiIdSsmParameterName?: string;
   tracingEnabled?: boolean;
   onDemandFailoverOnError?: string[];
+  workflowUrl?: string;
 }
--- a/lambdas/functions/control-plane/src/aws/runners.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.ts
@@ -244,6 +244,10 @@ async function createInstances(
     tags.push({ Key: 'ghr:trace_id', Value: traceId! });
   }
 
+  if (runnerParameters.workflowUrl) {
+    tags.push({ Key: 'ghr:workflow_url', Value: runnerParameters.workflowUrl });
+  }
+
   let fleet: CreateFleetResult;
   try {
     // see for spec https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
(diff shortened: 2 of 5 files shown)


npalm:
@roberto-civitas thx for the PR. Good idea, but I am afraid it won't work.

Runners are indeed created based on the workflow_job event, however which job starts on the runner is orechestrated by GitHub. So the actual job running on a runner is not alwasy the event that leading to creation of the runner. The module supports XRay which allows you to see correlation between which events is causing runner creation.

Next to het workflow_job event created there is a event for in_progress. This event could be used to tag an instance with the job in progress. It should not hard to correlate based ont he the event since this event contians the runner name as well.

Wondering what your thoughts are.

roberto-civitas:
Thank you for having a look at the PR. I will modify it so that it uses the event you suggest.

npalm:
> Thank you for having a look at the PR. I will modify it so that it uses the event you suggest.

Maybe good to have a short discussion on how to implement this. This to avoid you wast time on a proposal that is not fitting. My first thought would be having an extra lambda as part of the control plane. But no time to dig in so far.

edersonbrilhante:
This can be done the same way I implemented in Forge using runner_template_hook_job_started.
Inside runner_template_hook_job_started, I added a command to update the EC2 tags, as you can see in the PR:
https://github.com/cisco-open/forge/pull/91

Here is where to pass the runner_template_hook_job_started to runners module:
https://github.com/cisco-open/forge/blob/df3e1a38cf522e03e49b290c3ff97c7e9351f485/modules/platform/ec2_deployment/main.tf#L154

edersonbrilhante:
@npalm  @roberto-civitas
Check out this idea. I am updating EC2 instances using the EventBridge event bus as a trigger. I created a new rule to subscribe to in_progress and completed events, which then triggers a Lambda to add tags with info about the job ID and job URL. This could be extended to include more properties.

It’s already working in Forge, so it could be added as part of terraform-aws-github-runner.

https://github.com/cisco-open/forge/blob/main/modules/platform/ec2_deployment/lambda/update_ec2_tags.py
https://github.com/cisco-open/forge/blob/main/modules/platform/ec2_deployment/update-ec2-tags.tf

npalm:
@edersonbrilhante Yup I think that approach is much beeter. Maybe good to put this by default in the scripts provided in this repo. However I would avoid any context info in this repo that is user controlled, e..g github_ref. Others should be fine.

PR #4667, opened 2025-07-20
fix: Tag non jit instances with runner id and add docs

npalm:
⚠️ This PR is on top of #4539

This pull request is an enhancement on top of #4539. The PR also introduces tagging via the start scripts for non-JIT configured instances. And it adds a state diagram to the ever growing complexity of the state diagram below terminating instances.

### Instance Tagging Enhancements:
* Add function to tag runners with runner id for non JIT instances. Functions are designed to ignore errors to avoid causing a crash of the runner registration process.
* side effect is that the instance is allowed to create the tag `ghr:github_runner_id` the instance is allowed to create the tag only on itself.

### Added docs
* State diagram for scale-down added

Untitled diagram _ Mermaid Chart-2025-07-20-140732

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -46,6 +46,11 @@ markdown_extensions:
   - admonition
   - pymdownx.details
   - pymdownx.superfences
+  - pymdownx.superfences:
+      custom_fences:
+        - name: mermaid
+          class: mermaid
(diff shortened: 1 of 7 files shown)


stuartp44 (approved):
LGTM

PR #4668, opened 2025-07-21
chore: Lock workflow dependencies (OSSF recommendations)

npalm:
This pull request introduces support for managing Python dependencies for the `mkdocs` documentation workflow using Dependabot and adds a `requirements.txt` file to specify these dependencies. This address one of the finings by OSSF. The changes aim to automate dependency updates and ensure consistent dependency management for the documentation build process. 

Also avoid pages got published via a branch.


.github/workflows/update-docs.yml
```diff
@@ -86,6 +86,10 @@ jobs:
           path: .cache
           restore-keys: |
             mkdocs-material-
-      - run: pip install mkdocs-material
-      - run: pip install mkdocs-material-extensions
-      - run: mkdocs gh-deploy --force -c -b gh-pages
+      - run: pip install -r .github/workflows/mkdocs/requirements.txt
```
rjaegers: If you choose to use hashes you need to modify this to

```suggestion
      - run: pip install --require-hashes -r .github/workflows/mkdocs/requirements.txt
```

.github/workflows/mkdocs/requirements.txt
rjaegers, line 1: If you really want to be paranoid about the versions, and the specific backing version that is eventually installed you can use a 'lockfile' here as well. That also helps in not needing to specify all transitive dependencies manually. Dependabot has native support for *.in files as well, it will use pip-compile to update both the *.in file and the *.txt file, including the hashes.

See: https://github.com/philips-software/amp-devcontainer/blob/main/.devcontainer/cpp/requirements.in
And: https://github.com/philips-software/amp-devcontainer/blob/main/.devcontainer/cpp/requirements.txt

You can generate this locally to bootstrap the process with: `pip-compile --generate-hashes requirements.in`

rjaegers (dismissed):

.github/workflows/mkdocs/requirements.txt
npalm: Can you add that to this pr?

.github/workflows/update-docs.yml
npalm: does this requires hashes in the txt?

rjaegers: Yup, it does. But the pip-compile command bootstraps that process for you. I'll add that to the PR.

rjaegers (approved):

Issue #4712, opened 2025-08-15
jitconfig USERNAME on Windows


sdarwin:
Hi,
Upgrading the terraform code, I switched from `enable_jit_config: false` to `enable_jit_config: true`

Ephemeral multi-runners on Windows.

The user before was `runner_run_as: Administrator`

Testing with JIT, the whoami command shows "SYSTEM". The environment variable USERNAME is "EC2AMAZ-759AC2P$".  Perhaps it's running as the user "SYSTEM". In any case, this is incorrect for our tests...

How can the runner user be set to "Administrator" with jit config, similar to `runner_run_as: Administrator`.

Returning back to  `enable_jit_config: false` it's still broken and uses "SYSTEM".  Why...  Notice in this file https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/runners/templates/start-runner.ps1

```
        Write-Host "Starting without jit config"
        Invoke-Expression ".\run.cmd"
```

No user is specified in that invocation.  

It runs as the Task Scheduler user which is "SYSTEM".  

How did it work before?  Answer: two years ago it would do this: `Register-ScheduledTask -TaskName "runnertask" -Action $action -Trigger $trigger -User $username -Password $password -RunLevel Highest -Force` registering a task, and then launching the task, with a particular user, so you could control the user with the variable `runner_run_as: Administrator`.

What is the solution?

Possibly to move all ".\run.cmd" invocations back inside ScheduledTasks.  

That used to work.  And it allows control of the Username.

For the moment, our tests are broken, because they can not run as the user "System",  and `runner_run_as: Administrator` is ignored.

Edit:

The whole thing could be reverted to the previous version of start-runner.ps1 from years ago.   The earlier strategy.  Almost.  At least, the same idea.

Except... you have added these new features:  

- Write-Host "Stopping CloudWatch Agent"

and 

- Write-Host "Terminating instance"

I don't know the details about those. Before, the instances would terminate eventually, and when they terminate the cloudwatch agent would stop.   Does that mean those commands are "optional" rather than "required"?     It would make it easier to revert the code back.  Let me know.  Perhaps I could submit a PR.

PR #4723, opened 2025-08-20
fix: start-runner.ps1 set username

sdarwin:
This is to facilitate discussion of issue #4712 

Usually the username on windows comes from `runner_run_as: Administrator`. Recently that has broken and it's always `System`.

In `start-runner.ps1` move everything back inside a "Register-ScheduledTask" so the username can be set.   This fixes the issue.

However, `start-runner.ps1` also included a relatively new section:

```
Write-Host "Terminating instance"
aws ec2 terminate-instances --instance-ids "$InstanceId" --region "$Region"
```

By removing it, instances fail to terminate. How did this work before?  ( in 2023)  Why is it necessary to terminate instance from `start-runner.ps1` instead of how it used to work before?    

By removing `terminate-instances`, the instances continue to run, which is obviously not the desired result.  

Where should `terminate-instances` be placed, or can the earlier solution handle that task.

### DIFF
--- a/images/windows-core-2019/windows-provisioner.ps1
+++ b/images/windows-core-2019/windows-provisioner.ps1
@@ -22,16 +22,19 @@ Set-Content -Path "$PsHome\Microsoft.PowerShell_profile.ps1" -Value $ChocoProfil
 
 refreshenv
 
-Write-Host "Installing cloudwatch agent..."
+Write-Host "Installing cloudwatch agent, part 1"
 Invoke-WebRequest -Uri https://s3.amazonaws.com/amazoncloudwatch-agent/windows/amd64/latest/amazon-cloudwatch-agent.msi -OutFile C:\amazon-cloudwatch-agent.msi
-$cloudwatchParams = '/i', 'C:\amazon-cloudwatch-agent.msi', '/qn', '/L*v', 'C:\CloudwatchInstall.log'
-Start-Process "msiexec.exe" $cloudwatchParams -Wait -NoNewWindow
-Remove-Item C:\amazon-cloudwatch-agent.msi
 
 # Install dependent tools
 Write-Host "Installing additional development tools"
 choco install git awscli -y
 refreshenv
+
+Write-Host "Installing cloudwatch agent, part 2"
+# Delayed part 2 to ensure the download from part 1 completed and was written to disk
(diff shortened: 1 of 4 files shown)


sdarwin:
@npalm this PR has been tested and is ready to review. 

Closes #4712

npalm:
Only in case of ephemeral runners, the instance itself should terminate after the job is completed. This is done by running th runner in the foreground (linux) once the process is finsihed it exits the runner process and terminates the instance.

sdarwin:
You write "Only in case of ephemeral runners, the instance itself should terminate after the job is completed."

There was a somewhat new section in start-runner.ps1, doing this:  

```
Write-Host "Terminating instance"
aws ec2 terminate-instances --instance-ids "$InstanceId" --region "$Region"
```

What the PR is doing now is moving that code into a `start-runner-service.ps1` file which is called with the correct username.  
- The `terminate-instances` command is present.   
- the script is called by a configurable windows username, jobs may run as 'Administrator' or other user.

npalm (changes requested):
@sdarwin Running a deployment test based on the multi-runner example in the examples directoruy. The change is not working. I had no time to dig deep. But effect is the job got not processed. Assuming something going wrong in the registration of or starting of the windows runner.

The PR als has a merege conflict, but strange enough github is not showing me the issue. 

Would be great if you have some time to check the PR also with the confifguration in the mutli-runner example for windows. Or a similar example.

sdarwin:
> The PR has a merge conflict

Another modification appeared recently, and caused a merge conflict.  I believe it's resolved now.  

> The change is not working. 

I've just tested this, and it worked for me.  Microsoft Windows seems to be slow, it took a long time to launch.  Set all timeouts such as runner_boot_time_in_minutes and minimum_running_time_in_minutes , etc , to 20 minutes. 

There are at least 3 types of runners:  
- ephemeral jit
- ephemeral non-jit
- non-ephemeral  

In your environment currently, are all of these working with Microsoft Windows 2019/2022/2025? Are you testing them?  That is a first step before this PR.  

If they are all fine, as-is, then next try this pull request.  Let me know specifically for each of these if you are seeing either success or failure, and what happens:  

- ephemeral jit
- ephemeral non-jit
- non-ephemeral

PR #4726, opened 2025-08-22
fix: add handling for 404 errors in scale-down tests and improve error logging

### LINKED ISSUE #4724: Stale orphan EC2 instances not terminated when GitHub runner API returns no state or fetch fails
### Issue
There are stale EC2 instances tagged as Orphan (`orphan=true`) but they are not being terminated as expected during the scale down process. This occurs when the scale-down lambda fails to fetch runner details from the GitHub API (returns no state or a failed fetch, such as a 404).

### Impact
- Orphaned EC2 instances remain active, causing resource leakage and increased costs.
- The termination logic does not handle cases where the GitHub runner state cannot be fetched, resulting in instances persisting indefinitely.

### Expected Behavior
- Orphaned EC2 instances should be terminated if the GitHub API fails to fetch runner details or returns no state (e.g., 404 Not Found) by re-verifying it.
- The scale-down process should treat missing runner details as sufficient to proceed with termination.

### Steps to Reproduce
1. Deregister a GitHub runner so it no longer exists in the GitHub API.
2. Ensure the corresponding EC2 instance is tagged with `orphan=true`.
3. Observe that the scale-down lambda fails to terminate the instance when the API fetch fails or returns 404.

stuartp44:
This PR helps with dealing with 404 issues when terminating runners and checking if the runner is actually there or not. At the moment, we don't recognize the fact 404 is actually a good response for runners that are no longer there. This means the rest of the logic is ignored to do last chance checking. This PR helps by returning null for a 404, hence able to act on this or the runner data if we get it. For all other logic, we still return the previous error.

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
@@ -403,6 +403,106 @@ describe('Scale down runners', () => {
         expect(mockTerminateRunners).toHaveBeenCalledWith(orphanRunner.instanceId);
       });
 
+      it('Should handle 404 error when checking orphaned runner (JIT) - treat as orphaned', async () => {
+        // arrange
+        const orphanRunner = createRunnerTestData(
+          'orphan-jit-404',
+          type,
+          MINIMUM_BOOT_TIME + 1,
+          false,
+          true,
+          true, // should be terminated when 404
+          undefined,
+          1234567890,
+        );
+        const runners = [orphanRunner];
+
+        mockGitHubRunners([]);
+        mockAwsRunners(runners);
+
+        // Mock 404 error response
+        const error404 = new Error('Runner not found');
+        (error404 as any).status = 404;
+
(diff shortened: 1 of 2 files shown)


stuartp44:
When looking at the API spec, it looks like 404 is not even a valid return code. [Get a self-hosted runner for an organization](https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#get-a-self-hosted-runner-for-an-organization). I have reached out to Github to ask if this a documentation problem or if its a API problem.

stuartp44:
@npalm any update on this?

stuartp44:
Github have come back and confirmed that this is correct behavior and is documented https://docs.github.com/en/enterprise-cloud@latest/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#404-not-found-for-an-existing-resource

npalm (approved):
@stuartp44 thx, tested and looks good!

PR #4776, opened 2025-09-18
feat: Allow custom tags to be created by the EC2 instances during runtime

stuartp44:
This pull request introduces a new feature that allows users to specify additional EC2 tags that can be created by the GitHub Actions runner instances. The main changes involve updating the IAM policy and module configuration to support a customizable list of allowed tag keys.

**EC2 Tag Creation Customization:**

* Added a new variable `ec2_custom_allowed_tags_creation` to the module, allowing users to specify a list of additional tag keys permitted to be created within the EC2 instance. (`modules/runners/variables.tf`, `modules/runners/README.md`) [[1]](diffhunk://#diff-23e8f44c0f21971190244acdb8a35eaa21af7578ed5f1b97bef83f1a566d979cR764-R769) [[2]](diffhunk://#diff-c93c35ed3db08660e1625071dd9ff26329b9a72ee976ca0775ce483330fda0e5R153)
* Updated the IAM policy template (`instance-create-tags-policy.json`) to include both the default tag (`ghr:github_runner_id`) and any user-specified tags in the allowed list for the `ec2:CreateTags` action.
* Modified the `aws_iam_role_policy` resource to pass the new variable into the policy template, ensuring the policy is generated with the correct allowed tags. (`modules/runners/policies-runner.tf`)


modules/runners/variables.tf
```diff
@@ -761,3 +761,9 @@ variable "user_agent" {
   type        = string
   default     = null
 }
+
+variable "ec2_custom_allowed_tags_creation" {
+  description = "Allows the EC2 instance to create tags listed here."
```
ScottGuymer, line 766:
```suggestion
  description = "List of tags that an EC2 instance can create on itself. Can be specific tags or a namespace such as `custom:*`"
```

ScottGuymer (commented):
Small suggestion on the description.

I would also suggest thinking about the name of the variable to make it more precise to what it does. 

e.g. `ec2_instance_allowed_tags`

modules/runners/variables.tf
npalm: Please can you add to the description the paramater should be set with care. allowing for example `custom:*` allows also users of the runners to crate any tag.

modules/runners/variables.tf
```diff
@@ -761,3 +761,9 @@ variable "user_agent" {
   type        = string
   default     = null
 }
+
+variable "ec2_custom_allowed_tags_creation" {
```
npalm, line 765: suggest to renamen the parameter
```suggestion
variable "runner_custom_allowed_tags" {
```

npalm (changes requested):
@stuartp44 thx for creating the PR. made some suggestion about naming. I had no time to check the PR. But doubt it is working since the variable is not added to the 2 top level modules (root, and multi-runner).

modules/runners/variables.tf
ScottGuymer: It still only allows it to create tags on itself though, nothing else.

npalm:
Just adding a note, the module already supports injecting policies to get attached to trunner instance role, search for runner_iam_role_managed_policy_arns in either root module or mulit-runner one.

PR #4789, opened 2025-09-25
feat: add cpu_options for EC2 launch template

kirschem-fernride:
This PR adds a feature which enables to use of `cpu_options` in the runner config.

Context: We have a workload where hyperthreading hurts performance. With this option, we can opt-out of hyperthreading for a certain runner configuration.

Tested by applying to our internal dev infra.


npalm (approved):
@kirschem-fernride thx looks good! Quick pro tip, when contributing to OS would be great if you can enable signed commits.

PR #4792, opened 2025-09-25
feat: support multiple SSM parameters for large runner matcher configs (#4790)

### LINKED ISSUE #4790: SSM Parameter Size Limit Hit for Runner Matcher Config
**Description:**

Currently, when updating the SSM Parameter `/github-action-runners/my-prefix/webhook/runner-matcher-config`, we occasionally hit the AWS SSM size limit for standard parameters:

```
Error: updating SSM Parameter (/github-action-runners/my-prefix/webhook/runner-matcher-config): 
operation error SSM: PutParameter, https response error StatusCode: 400, RequestID: ..., 
api error ValidationException: Standard tier parameters support a maximum parameter value of 4096 characters. 
To create a larger parameter value, upgrade the parameter to use the advanced-parameter tier.
```

This is caused by the number of labels per runner type combined with the number of runner types exceeding the 4096-character limit.

**Proposed Solutions / Improvements:**

1. **Support multiple SSM parameters via environment variable:**
   Update `PARAMETER_RUNNER_MATCHER_CONFIG_PATH` to accept multiple parameter paths, separated by a colon. Example:

   ```
   /github-action-runners/my-prefix/webhook/runner-matcher-config-1:/github-action-runners/my-prefix/webhook/runner-matcher-config-2
   ```

   This would allow the matcher config to be split across multiple SSM parameters.

2. **Use S3 for storing the matcher config:**
   Instead of storing the full JSON in a single SSM parameter, allow the environment variable to point to an S3 path containing the matcher config JSON.

**Impact:**

* Avoid hitting SSM size limits for large runner configurations.
* Improve scalability as the number of labels and runner types grows.

**References:**

* [[AWS SSM Parameter Store Advanced Parameters](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html)](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-advanced-parameters.html)

edersonbrilhante:
### **PR Description**

Implements support for splitting the runner matcher configuration across multiple SSM parameters.
`PARAMETER_RUNNER_MATCHER_CONFIG_PATH` can now accept multiple parameter paths separated by a colon (`:`).

This avoids SSM size limits for large configurations and improves scalability for environments with many runner types and labels.

Closes #4790


kpielack (approved):

lambdas/functions/webhook/src/ConfigLoader.test.ts
```diff
@@ -229,6 +277,32 @@ describe('ConfigLoader Tests', () => {
       expect(config.matcherConfig).toEqual(matcherConfig);
     });
 
+    it('should load config successfully from multiple paths', async () => {
```
npalm:
```suggestion
    it('should load config successfully from multiple paths with repo allow list', async () => {
```

npalm (commented):
@edersonbrilhante nice contribution. I just tried the parameter split with the multi-runner example in this repo. By adding a few config and adding more labes. Result is I got this error, related to the added `foreach`

```
│
│   on ../../modules/webhook/webhook.tf line 27, in resource "aws_ssm_parameter" "runner_matcher_config":
│   27:   for_each = { for idx, val in local.matcher_chunks : idx => val }
│     ├────────────────
│     │ local.matcher_chunks will be known only after apply
│
│ The "for_each" map includes keys derived from resource attributes that cannot be determined until apply, and so Terraform cannot determine the full set of keys that will
│ identify the instances of this resource.
│
│ When working with unknown values in for_each, it's better to define the map keys statically in your configuration and place apply-time results only in the map values.
│
│ Alternatively, you could use the -target planning option to first apply only the resources that the for_each value depends on, and then apply a second time to fully
│ converge.
```

Wondering you you tested the change? I was not able to get it working right now.

edersonbrilhante:
I did test it, but I forgot to include a large configuration and only tested the normal case. My bad. I found the same bug now.

edersonbrilhante:
@npalm I just pushed my fix. I tested a normal config and a tested a really big one with 7 ssm resources and it worked fine now.

edersonbrilhante:
@npalm Also, during this fix I found 1 bug created the last release.
Check this PR https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4806

modules/webhook/webhook.tf
```diff
@@ -24,11 +24,11 @@ locals {
 }
 
 resource "aws_ssm_parameter" "runner_matcher_config" {
-  for_each = { for idx, val in local.matcher_chunks : idx => val }
+  count = length(local.matcher_chunks)
```
npalm: This count still results in my case in the following error with a large configuration:

```
Plan: 118 to add, 7 to change, 0 to destroy.
╷
│ Error: Invalid count argument
│
│   on ../../modules/webhook/webhook.tf line 27, in resource "aws_ssm_parameter" "runner_matcher_config":
│   27:   count = length(local.matcher_chunks)
│
│ The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around
│ this, use the -target argument to first apply only the resources that the count depends on.

```

npalm: @edersonbrilhante just tested with the example multi-runner in the repo. With about 4 extra config files with quite some labels. But ending up with the error above.

edersonbrilhante: @npalm I see the issue now — I was testing in an environment where the SQS queues were already created, so that was a blind spot. Testing with your example helped me pinpoint the problem.

The root cause is that the config relies on the SQS ARN. Since Terraform only knows the ARN at apply time, trying to calculate the string ahead of time for data evaluation doesn’t work.

My workaround is to precompute the worst-case scenario (max AWS partition, region, and queue name lengths) to estimate the maximum string length. From that, I calculate the number of chunks needed, create the resources using that chunk count, and then populate each resource with the real config split across the chunks at apply time.

npalm:
Thx for the updates, will not be able to able to get the PR merged in the next weeks. I will catch up end of the month. Sorry for tthe delay.

edersonbrilhante:
No problem. :)

modules/webhook/eventbridge/dispatcher.tf
```diff
@@ -129,7 +129,12 @@ resource "aws_iam_role_policy" "dispatcher_ssm" {
   role = aws_iam_role.dispatcher_lambda.name
 
   policy = templatefile("${path.module}/../policies/lambda-ssm.json", {
-    resource_arns = jsonencode([var.config.ssm_parameter_runner_matcher_config.arn])
+    resource_arns = jsonencode(
+      concat(
+        [var.config.github_app_parameters.webhook_secret.arn],
```
npalm: This results in the dispatcher can access the github webhok secret. Which is not needed. Only the webhook lambda is required to access this secret. Adding the webhook secret paramater is correct for the direct case but not when the eventbride is used. So here `[var.config.github_app_parameters.webhook_secret.arn]` can be removed.

```
  ~ resource "aws_iam_role_policy" "dispatcher_ssm" {
        id          = "npalm-multi-dispatcher-lambda-72a4fc05:publish-ssm-policy"
        name        = "publish-ssm-policy"
      ~ policy      = jsonencode(
          ~ {
              ~ Statement = [
                  ~ {
                      ~ Resource = [
                          + "arn:aws:ssm:eu-central-1:123:parameter/github-action-runners/npalm-multi/app/github_app_webhook_secret",
                            "arn:aws:ssm:eu-central-1:123:parameter/github-action-runners/npalm-multi/webhook/runner-matcher-config",
                        ]
                        # (2 unchanged attributes hidden)
                    },
                ]
                # (1 unchanged attribute hidden)
            }
        )
        # (2 unchanged attributes hidden)
    }

```

modules/webhook/webhook.tf
```diff
@@ -4,12 +4,39 @@ locals {
 
   # sorted list
   runner_matcher_config_sorted = [for k in sort(keys(local.runner_matcher_config)) : local.runner_matcher_config[k]]
+
+  # Define worst-case dummy ARN/ID lengths
```
npalm, line 13: Would be helpfull to add a short description why thos word-case values are calculated. Now in context of the PR it is very clear. But once merged it woulde be helpfull to have a short text that this intermediate values are used to calculate the slices to split the matcher config cross multiple paramaters.

npalm (changes requested):
@edersonbrilhante took a while but tested and reviewed the PR.

- I left some small comments, one permission needs to be dropped
- can you have a look on the copilot suggestions as well
- I have tested incrasing the matching resulting in a split. and down size back to a single matcher variable. All works like a hcarm.

modules/webhook/webhook.tf
edersonbrilhante: Good point. I added the explanation in the code

lambdas/functions/webhook/src/ConfigLoader.ts
```diff
@@ -87,17 +87,44 @@ abstract class BaseConfig {
...
+    for (const path of paths) {
+      await this.loadParameter(path, 'matcherConfig');
+      combinedString += this.matcherConfig;
+    }
+
+    try {
+      this.matcherConfig = JSON.parse(combinedString);
+    } catch (error) {
+      this.configLoadingErrors.push(`Failed to parse combined matcher config: ${(error as Error).message}`);
+    }
```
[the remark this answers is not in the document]
edersonbrilhante: I try use this recommendation, it does not work with the test suite. I don't see value refactoring the whole testing to use this suggestion.

modules/webhook/eventbridge/dispatcher.tf
edersonbrilhante: Thanks. I fixed it :)

npalm:
Expect to have time later this week to re-review the PR. Thanks for the updates

README.md
```diff
@@ -189,7 +189,7 @@ Join our discord community via [this invite link](https://discord.gg/bxgXW8jJGh)
...
-| <a name="input_runner_cpu_options"></a> [runner\_cpu\_options](#input\_runner\_cpu\_options) | The CPU options for the instance. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#cpu-options for details. Note that not all instance types support CPU options, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-optimize-cpu.html#instance-cpu-options | <pre>object({<br/>    core_count       = number<br/>    threads_per_core = number<br/>  })</pre> | `null` | no |
+| <a name="input_runner_cpu_options"></a> [runner\_cpu\_options](#input\_runner\_cpu\_options) | TThe CPU options for the instance. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#cpu-options for details. Note that not all instance types support CPU options, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-optimize-cpu.html#instance-cpu-options | <pre>object({<br/>    core_count       = number<br/>    threads_per_core = number<br/>  })</pre> | `null` | no |
```
stuartp44:
```suggestion
| <a name="input_runner_cpu_options"></a> [runner\_cpu\_options](#input\_runner\_cpu\_options) | The CPU options for the instance. See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template#cpu-options for details. Note that not all instance types support CPU options, see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-optimize-cpu.html#instance-cpu-options | <pre>object({<br/>    core_count       = number<br/>    threads_per_core = number<br/>  })</pre> | `null` | no |
```

modules/webhook/webhook.tf
```diff
@@ -4,12 +4,45 @@ locals {
 
   # sorted list
   runner_matcher_config_sorted = [for k in sort(keys(local.runner_matcher_config)) : local.runner_matcher_config[k]]
+
```
stuartp44, line 7: General comment here, is all theese comments needed? it seems like some of it is repeating the same as the overview and in some cases you can clearly see what you are intending in the code.

stuartp44 (dismissed):
Small feedback from me

modules/webhook/webhook.tf
npalm: I think the comments are good to have, it would help tu understand why this difficult processing is in place here.

npalm (approved):

npalm:
@edersonbrilhante would you have time to have a look at this issue #4901. It seems this changes is breaking the module in some caes.

PR #4806, opened 2025-10-03
fix: prevent unnecessary updates when cpu_options is unset

edersonbrilhante:
This PR will replace:

```hcl
cpu_options {
  core_count       = var.cpu_options != null ? var.cpu_options.core_count : null
  threads_per_core = var.cpu_options != null ? var.cpu_options.threads_per_core : null
}
```

with:

```hcl
dynamic "cpu_options" {
  for_each = var.cpu_options != null ? [var.cpu_options] : []
  content {
    core_count       = try(cpu_options.value.core_count, null)
    threads_per_core = try(cpu_options.value.threads_per_core, null)
  }
}
```

This way, if `cpu_options` is not set, Terraform will **not force a resource update every time**.

This issue was introduced in PR #4789 

image

### DIFF
--- a/modules/runners/main.tf
+++ b/modules/runners/main.tf
@@ -160,9 +160,12 @@ resource "aws_launch_template" "runner" {
     }
   }
 
-  cpu_options {
-    core_count       = var.cpu_options != null ? var.cpu_options.core_count : null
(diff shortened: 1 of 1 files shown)


npalm (approved):
@edersonbrilhante thx

Issue #4833, opened 2025-10-19
Support for Custom Tags in AWS SSM Parameter Store


wadherv:
Currently, the Lambda function responsible for creating AWS SSM Parameter Store entries in the GitHub AWS Runners Terraform module only supports a hardcoded tag:

```
await putParameter(${githubRunnerConfig.ssmTokenPath}/${instance}, runnerConfig.data.encoded_jit_config, true, {  
   tags: [{ Key: 'InstanceId', Value: instance }],
})
```

Ref:

1. https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/lambdas/functions/control-plane/src/scale-runners/scale-up.ts#L182
2. https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/lambdas/functions/control-plane/src/scale-runners/scale-up.ts#L409
3. https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/lambdas/functions/control-plane/src/scale-runners/scale-up.ts#L467

This implementation does not allow for additional custom tags to be added via environment variables or configuration. Supporting custom tags would be highly beneficial for organizations that need to:
- Track cost allocation across different teams or projects.
- Manage ownership and accountability of resources.
- Integrate with internal tagging policies for governance and automation.

wadherv:
PR Raised: https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4834

PR #4834, opened 2025-10-19
feat(control-plane): tag control plane created SSM Parameters (#4833)

wadherv:
Introducing a new variable `parameter_store_tags` to add tags for all the SSM Parameter Store resources created via Scale-Up and Pool lambda function.


npalm:
@wadherv thx for your PR, I wont have no option to check open PRs till the end of the month, sorry for the delay

wadherv:
> @wadherv thx for your PR, I wont have no option to check open PRs till the end of the month, sorry for the delay

@npalm if you’ve got a moment, could you take a look at my PR? Would really appreciate it!

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -49,6 +49,7 @@ interface CreateGitHubRunnerConfig {
   disableAutoUpdate: boolean;
   ssmTokenPath: string;
   ssmConfigPath: string;
+  ssmParameterStoreTags: { Key: string; Value: string }[];
```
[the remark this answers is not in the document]
npalm: it is consistence in context of lambda

modules/runners/pool/variables.tf
```diff
@@ -16,6 +16,7 @@ variable "config" {
       timeout                        = number
       zip                            = string
       subnet_ids                     = list(string)
+      parameter_store_tags           = string
```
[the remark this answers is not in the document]
wadherv: Lambda function environment variable only accepts string, map value is converted into string at https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4834/files/7361092dac821cbf6679bc7ae8966f1288b26169#diff-5c6c370f29fac62114adf0ecc2806bc009702bcf32f96bad61a8d40c17448462

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -41,6 +41,9 @@ export async function adjust(event: PoolEvent): Promise<void> {
   const onDemandFailoverOnError = process.env.ENABLE_ON_DEMAND_FAILOVER_FOR_ERRORS
     ? (JSON.parse(process.env.ENABLE_ON_DEMAND_FAILOVER_FOR_ERRORS) as [string])
     : [];
+  const ssmParameterStoreTags = process.env.SSM_PARAMETER_STORE_TAGS
+    ? JSON.parse(process.env.SSM_PARAMETER_STORE_TAGS)
+    : {};
```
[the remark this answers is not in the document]
wadherv: fixed

npalm:
Working my way to my backlog, thx for the reminder

modules/runners/local.tf
```diff
@@ -0,0 +1,5 @@
+locals {
```
npalm, line 1: @wadherv sorry for the long wait. Just test the PR with a small adjustment. But would first discuss the intended working.

I see a fewo options, but looking for use cases.

1. Just apply the values in var.tags to the paramters. Most simple approach.
2. Allow users two set custom tags for the pramaters, which are applied instead of adding the defautl from var.tgas. But for this I looking for a use case.

I have adjust the this local.tf file for a quick check, optiotn 2.
```
locals {
  parameter_store_tags = "[${join(", ", [
    for key, value in length(var.parameter_store_tags) > 0 ? var.parameter_store_tags : var.tags : "{ key = \"${key}\", value = \"${value}\" }"
  ])}]"
}

```

In this case paramters got the valures from var.tags applices. Since I have not set parameter_store_tags. 

What do you think. Are you looking for way have custom tags, or just getting the standard tags applices. When only adding the standard tag set in the module, we could even drop the introduced paramter (reducing complexity) and simply inject var.tags into the lambda.

wadherv: @npalm For my requirement, I want to apply only the default set of tags to the SSM Parameter Store, while also providing users the option to add additional custom tags specific to SSM Parameter resources.
Now that you have highlighted this, may be a better way is to merge var.tags and var.parameter_store_tags so that it work as intended(similar to https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/runners/variables.tf#L716).
Let me know if this sounds reasonable, and i can make the required code changes.

wadherv: Something like below:
locals {
  parameter_store_tags = "[${join(", ", [
    for key, value in merge(var.tags, var.parameter_store_tags) : "{ key = \"${key}\", value = \"${value}\" }"
  ])}]"
}

wadherv: @npalm

npalm: Yes I think merging the custom (additonal tags would be the best choice).

wadherv: @npalm i have updated the code as suggested, please review

wadherv: @npalm can you please approve this PR if all looks good?

modules/runners/local.tf
```diff
@@ -0,0 +1,5 @@
...
+  parameter_store_tags = "[${join(", ", [
+    for key, value in merge(var.tags, var.parameter_store_tags) : "{ key = \"${key}\", value = \"${value}\" }"
+  ])}]"
+}
```
npalm, line 8: Format as valid JSON
```suggestion
locals {
  parameter_store_tags = jsonencode([
    for key, value in merge(var.tags, var.parameter_store_tags) : {
      Key   = key
      Value = value
    }
  ])
}
```

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -465,7 +474,7 @@ async function createJitConfig(githubRunnerConfig: CreateGitHubRunnerConfig, ins
       instance: instance,
     });
     await putParameter(`${githubRunnerConfig.ssmTokenPath}/${instance}`, runnerConfig.data.encoded_jit_config, true, {
-      tags: [{ Key: 'InstanceId', Value: instance }],
+      tags: [{ Key: 'InstanceId', Value: instance }, ...githubRunnerConfig.ssmParameterStoreTags],
```
npalm, line 614: The change was not working, made a suggestion to `locals.tf`. But effect was this line was casuign an exception since the tags were not proper formatted. The change in `locals.tf` is fixing this. Although if you see any option here to validate the tags (without API) call as safe gurd that woudl be great.

npalm (dismissed):
@wadherv tested the PR, and find a bug. The suggestion in `locals.tf` is address this. Bit wondering if we can check / validate the tags in anyway to avoid those errors.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
wadherv: added https://github.com/github-aws-runners/terraform-aws-github-runner/blob/d43ea2fbeedd1547280493029ac462d53c73cfe0/lambdas/functions/control-plane/src/scale-runners/scale-up.ts#L91 to validate the tags

wadherv: @npalm

npalm: Yes I know, but too busy. Sorry man, not having enough time for open source. We also getting a new major release out. This breaks your PR. Please can you see if you rebase to (again).

wadherv: @npalm I understand that you’re busy, and that’s completely fine. Just wanted to add some context from my side — governance from my organization has introduced new policies around mandatory tagging for all AWS resources. Because of this, these PR/changes are quite important for us.

I have rebased my branch, can you please have a look?

npalm: thx for rebasing

wadherv: Hi @npalm, Happy New Year! 🎉
Any chance you’ve had time to look at this PR? Would appreciate an ETA if possible.

Brend-Smits:
I've reviewed the PR comments. Here's the status:

**Comment 1** (modules/runners/pool/variables.tf:19 - type should be map(string)): 
The current type `string` is actually **correct**. This variable receives the JSON-encoded value from `local.parameter_store_tags` (defined in modules/runners/local.tf line 2), which is a string representation of the tags array. Lambda environment variables must be strings, so this type definition is appropriate.

**Comment 2** (lambdas/functions/control-plane/src/scale-runners/scale-up.ts:113 - validation rejects empty values):
This is a **valid concern**. AWS SSM Parameter Store does allow empty tag values, so the validation should be updated to permit them. I'll fix this now.

Edit: fixed it, now running tests.

modules/runners/pool/variables.tf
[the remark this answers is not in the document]
Brend-Smits: the current type string is correct since it receives a JSON-encoded string for the Lambda environment variable.

Brend-Smits (approved):
I've spend some time testing this and it works as advertised. Happy to merge this. Great work and appreciate the patience!

wadherv:
@npalm can you also please approve this? I cannot merge this without your approval.

wadherv:
@npalm

stuartp44:
@wadherv can you rebase as its failing due to the lack of verified signatures on commits.

wadherv:
Hey @stuartp44, my branch already has the latest changes(from 2 weeks ago).
Can you please point me to what exactly is missing?

stuartp44:
You can follow the guide at https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits for this!

npalm (approved):

PR #4843, opened 2025-10-22
feat: Add Enterprise Runners support

micr0zz:
This is to add support for Enterprise level runners. Although GitHub supports applications at enterprise level, this type of authentication cannot be used for most of the requests (register runner/get runner, etc...).

WIP


npalm:
@micr0zz thx, great you working on supporting enterprise level runners. I have not digged in detail in your PR. But quickly scanned the PR. I think we should introduce breaking changes to support enterprise. this to avoid configuration is getting super complex. What I mean here moving away from `enable_enterprise_runners` and `enable_org...`. And teplace this by an object  or single string verable that indicates the runner level. for example `runner_level` and use a condition to allow 3 values.

Also wonderig what the status is of the PR. Still in WIP? In that case ca you mark the PR as draft.

npalm:
@micr0zz I moved the PR to draft, it seems still in WIP. Are you still working on this?

PR #4856, opened 2025-10-30
fix(runners): support for AMIs with parentheses in the name

Sbozzolo:
Some AMIs, like the `AWS Deep Learning Base GPU AMI (Ubuntu 24.04)`, have parentheses in the name. Parentheses are not allowed in AWS tags, leading to an error. This commit strips the parentheses out of the name.


Sbozzolo:
I rebased to main and fixed the commit names.

modules/runners/main.tf
```diff
@@ -102,7 +102,8 @@ resource "aws_ssm_parameter" "runner_ami_id" {
   tags = merge(
     local.tags,
     {
-      "ghr:ami_name" = data.aws_ami.runner.name
+      # Remove parentheses from AMI name to comply with SSM tag constraints
```
npalm: Assuming you mean AWS tag contraints in general, right?
```suggestion
      # Remove parentheses from AMI name to comply with AWS tag constraints
```

npalm (changes requested):
@Sbozzolo thx, sorry for the long wait. Small suggestion.

modules/runners/main.tf
Sbozzolo: Yes, thank you! I rebased on main and changed this.

Sbozzolo:
@npalm, friendly ping, for when you have a minute :)

npalm (approved):

PR #4857, opened 2025-10-30
docs: fix typos

Sbozzolo:
As I went through the code, I found various typos, so I thought I would just fix them all in one go.

I used codespell: https://github.com/codespell-project/codespell


npalm (approved):
@Sbozzolo thanks. Would be even better if we can automate this. For your information all terraform docs in markdown is generated base on tf files.

PR #4875, opened 2025-11-06
feat(runner-role): Enable using separate IAM role for runners

maratinvitae:
### What
Allow customization of runner IAM role

### Description
This PR introduces the ability to explicitly specify an IAM role and instance profile for the runner instances. This is motivated by a need to accommodate legacy IAM roles that remain from previous infrastructure migrations. 
Proposed change is backward-compatible.

### DIFF
--- a/modules/multi-runner/variables.tf
+++ b/modules/multi-runner/variables.tf
@@ -186,6 +186,17 @@ variable "multi_runner_config" {
         lambda_timeout     = optional(number, 30)
         max_attempts       = optional(number, 1)
       }), {})
+      iam_overrides = optional(object({
+        override_instance_profile = optional(bool, null)
+        instance_profile_name     = optional(string, null)
+        override_runner_role      = optional(bool, null)
+        runner_role_arn           = optional(string, null)
+        }), {
+        override_instance_profile = false
+        instance_profile_name     = null
+        override_runner_role      = false
+        runner_role_arn           = null
+      })
     })
     matcherConfig = object({
       labelMatchers = list(list(string))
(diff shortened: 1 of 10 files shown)


variables.tf
```diff
@@ -108,6 +108,23 @@ variable "runner_group_name" {
   default     = "Default"
 }
 
+variable "iam_overrides" {
```
npalm, line 111: not a review yet, but just a quick note. Adding a variable to the root module, also rquires to update the multi-runner module. Since that module can be used as a root module as well.

maratinvitae: thanks - updated.

modules/multi-runner/variables.tf
```diff
@@ -718,3 +718,20 @@ variable "user_agent" {
   type        = string
   default     = "github-aws-runners"
 }
+
+variable "iam_overrides" {
```
npalm, line 777: Only defining the variable, but not passing it into the module won't work. 

I think the variable needs to be defined ad part of the runner_config. And next passed into the runner module.

npalm:
Sorry for keep you waining, PR is still on the radar.

npalm:
@maratinvitae sorry for the long wait, but we working hard to get through the backlog. Please can you rebase the PR?

npalm (commented):
PR looks fine, but want to run a quick test.

maratinvitae:
Hi @npalm, any updates on this PR? Just as a note, I have tested assuming external role successfully.

andrecastro:
One thing that could be really interesting here would be to move `policies-runner.tf` into a separate submodule under the runners module.

That would let us reuse the submodule to create the IAM role + instance profile independently from the rest of the runner infrastructure, while still leveraging all the default policies and existing configuration.

I’m really excited about this feature because in my case I need to control the role name ahead of time, and I also want to reuse the same role across multiple runner groups.

nomeelnoj:
> One thing that could be really interesting here would be to move `policies-runner.tf` into a separate submodule under the runners module.
> 
> That would let us reuse the submodule to create the IAM role + instance profile independently from the rest of the runner infrastructure, while still leveraging all the default policies and existing configuration.
> 
> I’m really excited about this feature because in my case I need to control the role name ahead of time, and I also want to reuse the same role across multiple runner groups.

I think that increases the scope of this change a bit beyond what it was intended to do.  I love the idea, but would like to see this landed and the submodule introduced in a different PR.

Brend-Smits (approved):
Hey!

Apologies for the long delay on this one.
We're trying to go through the massive backlog of PR's. But this one is now ready to get merged.
I will need to squash to get all commits signed, sorry about that.

Again, apologies for the delay but thanks a lot for the contribution. Future pull requests will be faster! 

You're awesome! :rocket:

stuartp44 (approved):

Issue #4901, opened 2025-11-20
Invalid count argument error with aws_ssm_parameter.runner_matcher_config in v6.9.0


m-artin-t:
## Describe the bug

When using module version 6.9.0, Terraform planning fails with an invalid count argument error in the webhook module.

## Error Message

```
Error: Invalid count argument
  on .terraform/modules/github_runners_v4.github_runner/modules/webhook/webhook.tf line 41, in resource "aws_ssm_parameter" "runner_matcher_config":
  41:   count = local.total_chunks

The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the count depends on.
```

## Configuration

Using the module with standard configuration:

```hcl
module "github_runner" {
  source  = "github-aws-runners/github-runner/aws"
  version = "6.9.0"

  # Standard configuration
  aws_region = "us-east-1"
  vpc_id     = "vpc-xxx"
  subnet_ids = ["subnet-xxx"]
  
  github_app = {
    key_base64    = "..."
    id            = "..."
    client_id     = "..."
    client_secret = "..."
  }
  
  enable_organization_runners = true
  runners_maximum_count       = 400
  
  # Using default matcher_config_parameter_store_tier = "Standard"
}
```

## Environment

- Terraform Cloud (remote execution)
- Module version: 6.9.0
- AWS Provider: 5.77.0

## Context

This appears related to PR #4792 which added support for multiple SSM parameters for large runner matcher configs. The PR description mentions fixing a `for_each` error, but the `count` issue in `aws_ssm_parameter.runner_matcher_config` still exists.

The `local.total_chunks` calculation depends on runtime values, making it impossible for Terraform to determine the count during plan phase.

## Workaround

The suggested `-target` workaround doesn't work with Terraform Cloud's remote execution model.

## Expected behavior

The module should be able to plan successfully without requiring staged applies. Either:
- Make `local.total_chunks` computable at plan time
- Provide a variable to set a fixed chunk count
- Use a different approach that doesn't require dynamic count

## Additional notes

This issue doesn't occur in v6.8.5 (before PR #4792). The multi-SSM parameter feature is useful for large configs, but breaks standard deployments.

npalm:
@m-artin-t thx for reporting the problem.

npalm:
Quickly validated the example with the default example. And having the same issue.

[...]

PR #4903, opened 2025-11-21
fix: use a static key in runner_matcher_config (#4901)

edersonbrilhante:
# Description

Fix the issue found in #4901.

It now uses a static key instead of waiting for the queue ID to be created before generating the object, allowing the size to be calculated immediately.


edersonbrilhante:
@npalm @m-artin-t Can you take a look?

koendelaat (approved):

npalm (approved):

npalm:
fix: #4901

npalm:
@edersonbrilhante thx, a bit better option then mine #4902

PR #4915, opened 2025-11-26
feat(image-examples): add windows 2025 and ubuntu 24.04 examples

jonanderson10:
- Added Windows Server 2025 example image
- Added Ubuntu 24.04 (Noble Numbat) example image
- Bumped all example images to build with newer instance types to save time


npalm:
No time yet to check in detail. Any reason to update linuxe instance types?

jonanderson10:
> No time yet to check in detail. Any reason to update linuxe instance types? 

Much much faster builds. I figure if someone is really cost conscious they can override it since it's a variable. Most builds are going to be under 1hr for Linux anyway. So what's a few extra cents for less waiting.

PR #4929, opened 2025-12-03
feat: add support to use placement group in launch template

edersonbrilhante:
## Description

This PR adds support for configuring EC2 placement groups for GitHub Actions runners in the multi-runner module. It plumbs a new placement option from the runner configuration through to the underlying EC2 runner module.

## Details

Updated modules/multi-runner/runners.tf to pass placement = each.value.runner_config.placement into the runners module.
This allows specifying AWS placement groups for EC2 runners, enabling tighter control over instance placement.
The change is backwards compatible: if placement is unset in runner_config, behavior remains unchanged.

## Motivation / Future work

Placement groups are a prerequisite for supporting macOS runners, which require a host_id.
A follow-up PR will add explicit macOS support leveraging this new placement wiring.


modules/multi-runner/variables.tf
```diff
@@ -130,6 +130,17 @@ variable "multi_runner_config" {
         core_count       = number
         threads_per_core = number
       }), null)
+      placement = optional(object({
+        affinity          = optional(string)
+        availability_zone = optional(string)
+        # group_id                = try(placement.value.group_id, null) # aws >= 6.12.0
```
npalm: Why is the comment. Requiring a newer version of the provider?

npalm (changes requested):
Looks good, small question

modules/multi-runner/variables.tf
edersonbrilhante: The provider aws also allow to use group ID in placement, but only not in all versions. I left the commented so it can be added in the future. I can remove if is misleading

npalm: I will release next week a new major version, including bump of aws provider, and update to node24. You could update the PR to merge in next instead of main.

edersonbrilhante: Sure thing.

npalm (approved):

npalm (changes requested):
Sorry tiny change,

Can you add the group_id in both variables file with optional(null). And enable the line in the main.tf. The next branch is on provider 6.21

edersonbrilhante:
@npalm Updated :)

npalm (approved):
@edersonbrilhante thx

Issue #2069, opened 2022-05-19
macOS Runners


ChrisEdson:
Hi there -

Is it possible to support macOS runners using this module?

Technically, GitHub [states that it is possible](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#macos) - and AWS also has mac instances that it can spin up.

Thanks!

npalm:
Yes this should certainly be possible. We have crafted the module for Amazon Linux back in the days. Later support was added for multiple architectures, linux distros, and windows. All with the great support of the community. 

Differences betwoon OS are in general
- The base AMI (AWS)
- The launch script to setup the instance and install the runner. Aka the user_data script

We certainly welcome PR's to support MacOS as well.

s0undt3ch:
FYI, besides the AWS Mac VMs costing a lot more, their minimum reservation is 24 hours.
So, ephemeral Mac runners is off the table.

4sudiptodas:
We are trying to re-use the module for macos runners, by changing the ami filter and user data. However there is a variable runner_os which can only have (linux/windows) as valid inputs. How do we workaround this? any suggestions @npalm

npalm:
Would be great when you are able to spport macos. I think it requires some refactoring the terraform module to allow multiple values. As well the model the scal-up is creating and scale-down is destroying is not fitting the model for OSX runners. How are you approaching this?

davidnewhall:
> However there is a variable `runner_os` which can only have (linux/windows) as valid inputs.

Set `runner_os` to `linux`, set `runner_disable_default_labels` to `true` and set these variables appropriately:
- `userdata_content` 
- `ami`
- `runner_extra_labels` (set all the labels you need including `self-hosted`)
- `instance_types` (`["mac1.metal"]`, etc)
- `minimum_running_time_in_minutes` (should be 1440)

The only other place `runner_os` seems to be used is in the cloudwatch logs configuration, so just use the same paths as linux for logs and it should be fine.

The hardest part about this is what to put into `userdata_content`. You can start with the linux template, but then you quickly realize is depends on things like `svc.sh` that you have no familiarity with. Then you say to yourself "sheesh, I have no idea how to actually test this script effectively." It gets daunting real quick, don't ask how I know. 

Once you figure out what works in userdata_content you should then back-port your work into the custom AMI so the runners spin up quickly.  And at the same time, open a PR against this repo to share your success with everyone else. <3

[...]

PR #4930, opened 2025-12-03
feat: add support for macos runners

edersonbrilhante:
## Summary
This PR adds support for provisioning GitHub Actions runners on macOS EC2 instances by wiring placement group configuration through the multi-runner module and introducing macOS-specific bootstrap scripts.

## Motivation

macOS EC2 instances require a dedicated host (host_id), which is managed through placement groups.
This work prepares the module to request mac-based EC2 runners by consuming placement configuration as defined in #4929

This PR closes #2070 #2069

Test Plan:
- deploy examples/dedicated-mac-hosts
- update file examples/multi-runner/templates/runner-configs/mac-arm64-26.yaml with host_resource_group_arn
and license_configuration_arn
- deploy examples/multi-runner


edersonbrilhante:
@npalm what do you think? 

I didn't test yet. I gonna test during this week.

npalm:
Sorry had no time yet to dig in.

edersonbrilhante:
@npalm I think the code is ready for review.

edersonbrilhante:
@koendelaat @guicaulada When the time permits can you also take a look?

Brend-Smits (changes requested):
Hey @edersonbrilhante 

This is a very exciting new feature, great work! Appreciate it a lot.

I've spend quite a bit of time testing this and there are a few things that I noticed:
- While there is a module now for setting up the dedicated hosts for macos, I wonder if this really makes sense and is the right approach. Not saying this is bad or anything, just wondering. It's quite a challenge now to set up and organize a macos based runner and requires a lot of manual effort still. Also requiring to input the license and host arn can make this less trivial to use than other parts of the framework.
- I could not get the spawning of instances to work, likely this was caused by the switch on the main branch to using CreateFleet instead of CreateInstance recently. Switching to 'RunInstancesCommand' with a temporary fix like:
```JS
async function createInstancesWithRunInstances(
  runnerParameters: Runners.RunnerInputParameters,
  amiIdOverride: string | undefined,
  ec2Client: EC2Client,
): Promise<string[]> {
  const tags = [
    { Key: 'ghr:created_by', Value: runnerParameters.numberOfRunners === 1 ? 'scale-up-lambda' : 'pool-lambda' },
    { Key: 'ghr:Type', Value: runnerParameters.runnerType },
    { Key: 'ghr:Owner', Value: runnerParameters.runnerOwner },
  ];

  if (runnerParameters.tracingEnabled) {
    const traceId = tracer.getRootXrayTraceId();
    tags.push({ Key: 'ghr:trace_id', Value: traceId! });
  }

  try {
    const instanceType = runnerParameters.ec2instanceCriteria.instanceTypes[0] as _InstanceType;
    const runInstancesCommand = new RunInstancesCommand({
      LaunchTemplate: {
        LaunchTemplateName: runnerParameters.launchTemplateName,
        Version: '$Default',
      },
      InstanceType: instanceType,
      MinCount: runnerParameters.numberOfRunners,
      MaxCount: runnerParameters.numberOfRunners,
      SubnetId: runnerParameters.subnets[0],
      ...(amiIdOverride ? { ImageId: amiIdOverride } : {}),
      TagSpecifications: [
        {
          ResourceType: 'instance',
          Tags: tags,
        },
        {
          ResourceType: 'volume',
          Tags: tags,
        },
      ],
    });

    const result = await ec2Client.send(runInstancesCommand);
    const instanceIds = result.Instances?.map((i) => i.InstanceId!).filter(Boolean) || [];

    if (instanceIds.length === 0) {
      throw new Error('RunInstances returned no instances for dedicated host.');
    }

    return instanceIds;
  } catch (e) {
    logger.warn('RunInstances request failed for dedicated host.', { error: e as Error });
    throw e;
  }
}
```
resolved the issues I had and resulted in machines getting spawned. So in this case you would have something like:
```JS
  // EC2 Fleet (CreateFleet) does not support launching instances onto dedicated hosts
  // for instance types like mac*.metal. Use RunInstances directly instead.
  if (runnerParameters.useDedicatedHost) {
    logger.info('Using RunInstances for dedicated host placement (CreateFleet does not support dedicated hosts).');
    const instances = await createInstancesWithRunInstances(runnerParameters, amiIdOverride, ec2Client);
    logger.info(`Created instance(s) via RunInstances: ${instances.join(',')}`);
    return instances;
  }
  ```
 - In addition to the above, I don't think these instance types support `spot` instance target capacity, meaning it should also not be an option. The example should reflect this as well by using the `instance_target_capacity_type: "on-demand"` runner option.

I'm sorry for the delay in testing/providing feedback on this PR. It's been a bit of a challenge getting this set up and with things going on outside of work, I've prioritized other things.

**Comment by edersonbrilhante** [comment]
@Brend-Smits 
> While there is a module now for setting up the dedicated hosts for macos, I wonder if this really makes sense and is the right approach. Not saying this is bad or anything, just wondering. It's quite a challenge now to set up and organize a macos based runner and requires a lot of manual effort still. Also requiring to input the license and host arn can make this less trivial to use than other parts of the framework.

That is restriction on AWS side, in a analogy I see mandatory requirement to use macos, like we have for VPC/subnet, but just for macos types.

> I could not get the spawning of instances to work, likely this was caused by the switch on the main branch to using CreateFleet instead of CreateInstance recently. Switching to 'RunInstancesCommand' with a temporary fix like:

I will rebase the code and test again.

**Comment by edersonbrilhante** [comment]
@Brend-Smits 
I added your code and it worked.

About your concerns:

Host groups require a self-managed license. When using a host group (instead of a specific host), AWS will select one host from the host group.

The “manual” step requires creating a host group, a host, and a license. Then you associate the license with the host group, and associate the host group with the host. This can be done via Terraform and only needs to be done once upfront.

After that, you just configure the module to use the appropriate placement and license_specifications.

This is a caveat when using Mac runners. I think a short doc explaining this should be enough.

**Comment by npalm** [comment]
@Brend-Smits Thanks for checking. I would suggest we clearly mark in docs as well in variables.

Suggestions:
- add clear not to variable `runner_os`
- update the docs add here a clear note: https://github-aws-runners.github.io/terraform-aws-github-runner/configuration/#experimental-features
- We not cover in docs what we in general not can test. I think we should find a good way to be very clear that mac we not test in general. Some hodls for all the example AMIs as well GHES.

File: examples/dedicated-mac-hosts/README.md
Comment (npalm) [inline examples/dedicated-mac-hosts/README.md:1]: - Can we add here a quick quide for the steps to setup the example.
- Please check the docs dir, new examples needs to be added here as well.

File: docs/mac-runners.md [hunk 1]
```diff
@@ -0,0 +1,188 @@
...
+
+- **Dedicated hosts required.** EC2 Mac instances must run on dedicated hosts. Each dedicated host can run **only one Mac VM at a time** (1:1 ratio). The module uses `RunInstances` directly instead of the `CreateFleet` API when `use_dedicated_host` is enabled.
+- **Longer boot times.** macOS instances can take 6–20 minutes to launch, significantly longer than Linux (~1 min) or Windows (~5 min). The default `minimum_running_time_in_minutes` for `osx` is set to 20 minutes to prevent premature scale-down.
+- **~50 minute host recycle time.** After an EC2 Mac instance is terminated, AWS performs install cleanup and software upgrades on the dedicated host before it becomes available again. This process takes approximately **50 minutes**, during which the host cannot launch a new instance.
+- **ARM64 (Apple Silicon) and x64 (Intel) support.** Both `mac1.metal` (Intel), `mac2.metal` (M1), and `mac2-m2.metal` (M2) instance types are supported. Set `runner_architecture` accordingly (`x64` or `arm64`).
+- **Only ephemeral mode is recommended.** Due to the long host allocation time and dedicated host cost model, we recommend using ephemeral runners.
```
npalm, line 16: Maybe I miss somthing. But to me epheral sounds a bad fit. Due to long cycle times. Reserved host allocaton. Or is it the host htat has the long cuycle time and the vm on the host not?

examples/multi-runner/templates/runner-configs/mac-arm64-26.yaml
npalm, line 1: Would dorp the mac setup from the multi-runner for now. Keep it to a stand-alnoe exmple. Unless it is convienet for you test. But it makes this test example much bigger, and in most cases we won't use it.

npalm (changes requested):
@edersonbrilhante great job, for me almost good to go.

- @Brend-Smits did you tsted the PR, I was not able to do this yet.
- Made a few tiny comments.

docs/mac-runners.md
edersonbrilhante: The good side, you don't pay if the host is not available.

edersonbrilhante: We can remove this recommendation then. I am using as ephemeral, so I can have fresh instances for security requirements.

examples/multi-runner/templates/runner-configs/mac-arm64-26.yaml
edersonbrilhante: You mean, just remove the example?

Brend-Smits:
> @edersonbrilhante great job, for me almost good to go.
> 
> * @Brend-Smits did you tsted the PR, I was not able to do this yet.
> * Made a few tiny comments.

Yes I did test this and it worked :)

edersonbrilhante:
Can we merge now?

Brend-Smits (dismissed):

Brend-Smits (dismissed):

edersonbrilhante:
@npalm I can resolve the conversations?

modules/runners/scale-down-state-diagram.md
```diff
@@ -145,6 +145,6 @@ stateDiagram-v2
 ## Configuration Parameters
 
 - **Cron Schedule**: `cron(*/5 * * * ? *)` (every 5 minutes)
-- **Minimum Runtime**: Linux 5min, Windows 15min
+- **Minimum Runtime**: Linux 5min, Windows 15min, OSX 20min
 - **Boot Timeout**: Configurable via `runner_boot_time_in_minutes`
```
steveclifton, line 149: Hey @edersonbrilhante 
Super excited for this PR 🚀 

One thing I noticed: `runner_boot_time_in_minutes` still looks like it defaults to 5 minutes. Since the docs here call out mac boot times in the 6-20 minute range, is there a risk the scale-down flow could mark a fresh mac runner as orphaned before it has a chance to register?

edersonbrilhante: I adjusted the minimum time to run in 30 minutes and worked fine in my tests

examples/multi-runner/templates/runner-configs/mac-arm64-26.yaml
npalm, line 1: Suggest to just drop mac os for now from multi runners, eep it with a dedicated example. 

Got this error

```│ Error: "license_specification.0.license_configuration_arn" (<license_configuration_arn>) is an invalid ARN: arn: invalid prefix
│
│   with module.runners.module.runners["mac-arm64-26"].aws_launch_template.runner,
│   on ../../modules/runners/main.tf line 130, in resource "aws_launch_template" "runner":
│  130: resource "aws_launch_template" "runner" {
│
╵
╷
│ Error: "placement.0.host_resource_group_arn" (<host_resource_group_arn>) is an invalid ARN: arn: invalid prefix
│
│   with module.runners.module.runners["mac-arm64-26"].aws_launch_template.runner,
│   on ../../modules/runners/main.tf line 130, in resource "aws_launch_template" "runner":
│  130: resource "aws_launch_template" "runner" {
│
╵

```

npalm (dismissed):
@edersonbrilhante Awesome. 

I would prefer we either comment the mac os config in multi runner examlle or drop it. THe example is now broken for me to a license error.

npalm:
Also bumped in an unexpected change of the windows userdata script, but this is the result of an extra new line. So not an issue at all.

examples/multi-runner/templates/runner-configs/mac-arm64-26.yaml
edersonbrilhante: I used placeholder to show it is needed to pass the your host_resource_group_arn and license_configuration_arn in the region you want to deploy. 
- host_resource_group_arn: "<host_resource_group_arn>"
- license_configuration_arn: "<license_configuration_arn>"

Can be the outputs from examples/dedicated-mac-hosts.

How about if I add a descriptive comment about this in the yaml file

edersonbrilhante:
@npalm can you review?

JFenstermacher:
Hey! We've been testing this branch with **persistent runners on macOS dedicated hosts** (pool-managed, non-ephemeral) and ran into two issues that likely don't surface in ephemeral mode. Sharing in case they're useful:

### 1. `processFleetResult` discards successfully-created instances on partial fleet failure

**File:** `lambdas/functions/control-plane/src/aws/runners.ts` (line 174-221)

When `CreateFleet` is asked for N instances but only creates some (common with dedicated hosts when one host is in pending/scrubbing state), the response contains both `Instances` and `Errors`. The current logic at line 219-220 throws `"Create fleet failed, no instance created."` — even though some instances **were** created.

Because the throw happens before `createStartRunnerConfig()` is reached, the successfully-launched instances never get their SSM token written. They boot, can't find their config in SSM, and self-terminate (orphaned).

This doesn't surface in ephemeral mode because scale-up typically requests 1 instance per job, making partial success unlikely. The pool Lambda requests `poolSize` at once, where partial success is the common case with dedicated hosts.

**Suggested fix** — return partial instances before throwing so `createRunners()` can write their SSM config:

```typescript
// After the on-demand failover check, before the scaleErrors check:
if (instances.length > 0) {
  logger.warn(
    `Partial fleet success: ${instances.length} of ${runnerParameters.numberOfRunners} created. ` +
      `Returning successful instances.`,
    { data: fleet.Errors },
  );
  return instances;
}
```

### 2. `start-runner-osx.sh` — `set -euo pipefail` kills script before SSM retry loop

**File:** `modules/runners/templates/start-runner-osx.sh` (lines 115-127)

The macOS script uses `set -euo pipefail` but the initial `aws ssm get-parameter` call has no error suppression. If the parameter doesn't exist yet (race between Lambda writing and instance booting), the pipeline exits non-zero and the script dies — the `while [[ -z "$config" ]]` retry loop is never reached.

The Linux `start-runner.sh` doesn't have `set -euo pipefail`, so the same code works there by accident (command fails silently, config is empty, retry loop fires).

**Suggested fix:**
```bash
config=$(aws ssm get-parameter \
  --name "$token_path"/"$instance_id" \
  --with-decryption \
  --region "$region" 2>/dev/null | jq -r ".Parameter | .Value") || config=""

while [[ -z "$config" || "$config" == "null" ]]; do
  echo "Waiting for GH Runner config to become available in AWS SSM"
  sleep 1
  config=$(aws ssm get-parameter \
    --name "$token_path"/"$instance_id" \
    --with-decryption \
    --region "$region" 2>/dev/null | jq -r ".Parameter | .Value") || config=""
done
```

docs/mac-runners.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Scaling caveats
+
+Running macOS runners at scale introduces challenges that do not exist with Linux or Windows runners:
+
+1. **1:1 host-to-VM ratio.** Unlike Linux where many instances share underlying hardware, each Mac VM requires its own dedicated host. To run N concurrent macOS jobs, you need at least N dedicated hosts.
+2. **Host recycle delay.** After a Mac instance is terminated, the dedicated host enters a ~50 minute cleanup cycle (scrubbing, software updates). During this window the host is unavailable. For bursty workloads, you need additional hosts to absorb demand while others recycle.
+3. **Capacity planning.** As a rule of thumb, if you expect N peak concurrent macOS jobs and each job takes T minutes, account for the extra ~50 minutes of host downtime per cycle when sizing your host pool.
```
grahamc: This is a really trivial nit, but defining terms (N, T) and then not using them in some calculation to help me size my pool was a letdown 😆.

modules/multi-runner/runners.tf
```diff
@@ -56,6 +56,8 @@ module "runners" {
   credit_specification                 = each.value.runner_config.credit_specification
   cpu_options                          = each.value.runner_config.cpu_options
   placement                            = each.value.runner_config.placement
+  license_specifications               = each.value.runner_config.license_specifications
```
spcaipers-arm, line 60: `license_specifications` only applies to mac right? Maybe the name should include prefix or suffic `mac`?

modules/multi-runner/variables.tf
```diff
@@ -246,6 +250,7 @@ variable "multi_runner_config" {
         vpc_id: "The VPC for security groups of the action runners. If not set uses the value of `var.vpc_id`."
         subnet_ids: "List of subnets in which the action runners will be launched, the subnets needs to be subnets in the `vpc_id`. If not set, uses the value of `var.subnet_ids`."
         idle_config: "List of time period that can be defined as cron expression to keep a minimum amount of runners active instead of scaling down to 0. By defining this list you can ensure that in time periods that match the cron expression within 5 seconds a runner is kept idle."
```
spcaipers-arm, line 255: missing the license agreement info?

modules/runner-binaries-syncer/main.tf
```diff
@@ -1,5 +1,5 @@
 locals {
-  action_runner_distribution_object_key = "actions-runner-${var.runner_os}.${var.runner_os == "linux" ? "tar.gz" : "zip"}"
+  action_runner_distribution_object_key = "actions-runner-${var.runner_os}.${var.runner_os == "windows" ? "zip" : "tar.gz"}"
```
spcaipers-arm: Is this expected?

modules/runners/templates/install-runner-osx.sh
```diff
@@ -0,0 +1,59 @@
+# shellcheck shell=bash
+
+## install the runner (macOS)
```
spcaipers-arm, line 5: set -euo pipefail?

docs/mac-runners.md
edersonbrilhante: Sorry. I fixed the description

modules/runner-binaries-syncer/main.tf
edersonbrilhante: Yes. Because linux and mac use `tar.gz`, so I change the conditional logic

modules/multi-runner/variables.tf
edersonbrilhante: Thanks. Fixed

modules/runners/templates/install-runner-osx.sh
edersonbrilhante: Added it

modules/multi-runner/runners.tf
edersonbrilhante: At this moment yes. But we are using aws naming spec(credit_specification, cpu_options, placement, etc) . I rather not customize just this config

edersonbrilhante:
> ### 1. `processFleetResult` discards successfully-created instances on partial fleet failure
> **File:** `lambdas/functions/control-plane/src/aws/runners.ts` (line 174-221)
> 
> When `CreateFleet` is asked for N instances but only creates some (common with dedicated hosts when one host is in pending/scrubbing state), the response contains both `Instances` and `Errors`. The current logic at line 219-220 throws `"Create fleet failed, no instance created."` — even though some instances **were** created.
> 
> Because the throw happens before `createStartRunnerConfig()` is reached, the successfully-launched instances never get their SSM token written. They boot, can't find their config in SSM, and self-terminate (orphaned).
> 
> This doesn't surface in ephemeral mode because scale-up typically requests 1 instance per job, making partial success unlikely. The pool Lambda requests `poolSize` at once, where partial success is the common case with dedicated hosts.
> 
> **Suggested fix** — return partial instances before throwing so `createRunners()` can write their SSM config:
> 
> ```ts
> // After the on-demand failover check, before the scaleErrors check:
> if (instances.length > 0) {
>   logger.warn(
>     `Partial fleet success: ${instances.length} of ${runnerParameters.numberOfRunners} created. ` +
>       `Returning successful instances.`,
>     { data: fleet.Errors },
>   );
>   return instances;
> }
> ```

Thanks. I just don't have time to test option and compatibility with ephemeral right now.

Brend-Smits (approved):
LGTM :shipit:!

Nice work!

PR #4948, opened 2025-12-12
feat: use prefix variable for POWERTOOLS_SERVICE_NAME in Lambda functions

alexalbu001:
Updates `POWERTOOLS_SERVICE_NAME` to use the `prefix` variable instead of hardcoded values in pool, ami-housekeeper, scale-down, scale-up, webhook and dispatcher Lambda functions.

## Why
- function names already include prefix, `POWERTOOLS_SERVICE_NAME` should too


npalm (changes requested):
@alexalbu001 thx, can you also fix the following places in code as well.

- modules/runners/pool/main.tf
- modules/runners/ssm-housekeeper.tf

PR looks good.

alexalbu001:
@npalm should be good now

npalm:
Sorry for the late catch, but why do you think the service name should include the prefix as well. The initial idea was to set only the name of the service, since it is in context of the lambda the prefix is part of that.

alexalbu001:
I understand the design philosophy of keeping the service name as a logical identifier separate from the deployment context, however, I see a few benefits to including the prefix:
- Consistency with function names: The Lambda function names already include the prefix (`${var.prefix}-scale-up`). Having `POWERTOOLS_SERVICE_NAME` match the function name creates consistency between what users see in the AWS console and what appears in their observability tools.
- Current behaviour: The existing hardcoded values already include "runners-" prefix (e.g., `"runners-scale-up"`), so this change just makes that pattern configurable rather than changing the philosophy

alexalbu001:
Hey @npalm, not sure if you've seen the above^

npalm:
I am fine with this change, @guicaulada do you also agree?

npalm (approved):

PR #4961, opened 2025-12-18
fix: job retry mechanism not triggering

### LINKED ISSUE #4960: job-retry mechanism broken in version 7.0.0
## Description

Yesterday we upgraded from version 6.9.0 to 7.0.0 and we started noticing an increase in job start time. After investigation we noticed that the job-retry mechanism is no longer functioning. Here are some screenshots:

Image

Image

You can see clearly the invocations and logs stopped at 14:00 UTC (when we did the update).

## Investigation

We're not sure yet what has caused it, we're still trying to figure this out. Any help is appreciated.

Edit 09:46:
- Version 6.9.0 is calling `publishRetryMessage(payload)` in the scale-up function.
- Version 7.0.0 is no longer calling the `publishRetryMessage(payload)` in the scale-up function, it's not calling it anywhere except for tests.

Edit 10:06:
- Testing a fix locally which we will make a PR for if it works properly
- Confirmed the fix is working as expected:

Image

stuartp44:
This pull request adds comprehensive tests for the retry mechanism in the `scaleUp` functionality and **re**introduces the `publishRetryMessage` call to the scale-up process. The tests ensure that the retry logic works correctly under various scenarios, such as when jobs are queued, when the maximum number of runners is reached, and when queue checks are disabled. 

**Testing and Retry Mechanism Enhancements:**

* Added a new test suite "Retry mechanism tests" in `scale-up.test.ts` to cover scenarios where `publishRetryMessage` should be called, including: when jobs are queued, when maximum runners are reached, with correct message structure, and when job queue checks are disabled.

**Other code Updates:**

* Fixed logic to skip runner creation if no new runners are needed by checking if `newRunners <= 0` instead of comparing counts, improving clarity and correctness.

<details><summary>Example scenarios for the above bug</summary>
<p>

Scenario 1
- Admin sets RUNNERS_MAXIMUM_COUNT=20
- System scales up to 15 active runners
- Admin reduces RUNNERS_MAXIMUM_COUNT=10 (cost control, policy change)
- Before those 15 runners terminate, new jobs arrive
- Bug triggers: newRunners = Math.min(scaleUp, 10-15) = -5
- Code tries to call createRunners({numberOfRunners: -5}) and fails

Scenario 2
- RUNNERS_MAXIMUM_COUNT=5
- Someone manually launches 8 EC2 instances with runner tags
- New jobs arrive
- Bug triggers: newRunners = Math.min(2, 5-8) = -3
- Code tries to call createRunners({numberOfRunners: -3}) and fails

Scenario 3
- Admin sets RUNNERS_MAXIMUM_COUNT=20
- System scales up to 15 active runners
- Admin reduces RUNNERS_MAXIMUM_COUNT=10 (cost control, policy change)
- Before those 15 runners terminate, new jobs arrive
- Bug triggers: newRunners = Math.min(scaleUp, 10-15) = -5
- Code tries to call createRunners({numberOfRunners: -5}) and fails

</p>
</details> 

We tested this in our staging environment and verified it's working.

Closes #4960


lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -356,6 +357,7 @@ export async function scaleUp(payloads: ActionRequestMessageSQS[]): Promise<stri
       }
 
       scaleUp++;
+      await publishRetryMessage(message as ActionRequestMessageRetry);
```
npalm: Good catch, the call got lost in the PR that introces the btaches.

I doubt a bit if we should publish the message hre. At this point no runner is created. Only the count is increased. But on the other hand we cannot match the created runners with the event.

I think it would be a bit safter to call the publishRetry messages at the end for each message that is not marked as invalid. Since the invalid once are going back to the queue and will be retried. This will lead that more retry messages are created for the same event.

npalm: @copilot what do you think?

Brend-Smits: I'm not sure if it's worth it to add all that additional logic (and an additional for-loop) into this function. It will make it way more complex as well..

In case the message is marked as invalid, worst case it goes through the retry mechanism, ends up in the same scale-up function and it still is an invalid message. Leading to minimal waste. 
IMO a good trade-off to avoid making this function even more complex.

What do you think @npalm?

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -395,7 +397,7 @@ export async function scaleUp(payloads: ActionRequestMessageSQS[]): Promise<stri
       }
 
       // No runners will be created, so skip calling the EC2 API.
-      if (missingInstanceCount === scaleUp) {
+      if (newRunners <= 0) {
```
iainlane, line 443: Would you say this is a separate fix? I think it'd be easier to get in via its own dedicated PR (with tests) if that's possible 👍

iainlane (commented):
Cheers - right now I just had one request!

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
stuartp44: I don't really see the benefit in my opinion, this is a "Boy Scout" rule. But happy if that is what is needed to get things moving!

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
stuartp44: @npalm, what if your feeling about what @Brend-Smits said?

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
```diff
@@ -1667,6 +1674,159 @@ describe('scaleUp with Github Data Residency', () => {
...
+    process.env.RUNNERS_MAXIMUM_COUNT = '0'; // No runners can be created
+
+    const messages = createTestMessages(2);
+
+    await scaleUpModule.scaleUp(messages);
+
+    // publishRetryMessage should still be called even though no runners will be created
+    expect(mockPublishRetryMessage).toHaveBeenCalledTimes(2);
+    expect(createRunner).not.toHaveBeenCalled();
+  });
```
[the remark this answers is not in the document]
stuartp44: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4961#discussion_r2632523665)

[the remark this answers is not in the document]
stuartp44: Resolved

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
npalm: No I do not agree, problem is in case of failure by pushing all messages back the presure on APIs only quickly increase. So I think it will be better to filter. I did an quick experiment on a branch with some copilot refactory. But had not the time to dig in futher.

Brend-Smits: This should now be resolved. Closing this conversation.

Brend-Smits:
Rebased the PR and resolved open conversations. This is now ready to get merged after getting approvals.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
npalm: @Brend-Smits Sorry I do not see how this is solved. All messages got published for retry. In case scaling if failing. It results in still for also the failing onces a retry is triggered. Problem is that in case of errors the presure on the whole system increase. Since the failing of creating is already creating preasur and trigger a retry via SQS. As weel the retry check got fired.

Brend-Smits: Okay sorry it seemed that the fix (that was in a different PR, which we merged), got overwritten likely by a rebase or something. That's why closed this conversation as it should have been fixed by that PR. 
I've gone back and manually merged those changes into the PR and it should not be addressed.

I'll now resolve this conversation, please open up a new conversation if the new solution requires changes.

andrecastro (approved):

andrecastro:
This is a critical issue for me. After upgrading our infrastructure to v7.3.0, we noticed that jobs are no longer being retried, which is resulting in a growing number of jobs stuck waiting indefinitely.

I understand this is an experimental feature, but It’s unclear why this behavior changed. It appears to be a regression and likely a bug.

Could one of the maintainers please review and approve this?

npalm:
I will give this PR priority early next week

npalm (approved):
@Brend-Smits @stuartp44 thx.

PR #4979, opened 2026-01-05
fix: change runner_placement host_resource_group_arn type to be a string instead of a number

phene:
# Description
Fixes bug in https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4929 that assigned the type `number` to `host_resource_group_arn` instead of `string`.


npalm (approved):
Looks good to me.

@edersonbrilhante do I remember correctly you introduced this. Looks to me a bug as well when checking the docs.

edersonbrilhante:
It is a bug indeed. Probably it was due a bad copy paste, when I set type `number` in partition_number. I remember going to aws provider repo to check the types in the go module for placement

I had fixed in my PR too https://github.com/github-aws-runners/terraform-aws-github-runner/pull/4930.

We can merge this PR and I rebase mine later

PR #4994, opened 2026-01-09
fix: cache GitHub App ID to reduce SSM calls

Brend-Smits:
Cache the GitHub App ID at module level to avoid repeated SSM parameter fetches on every rate limit metric update. The value is now fetched once per Lambda execution context and reused across invocations.

Especially when multiple events are badged, repeatedly calling the SSM Parameter API feels wasteful and adds a non-negligible amount of latency to several functions in the scale-up Lambda.


guicaulada (commented):
The implementation is solid and ready to merge from my perspective. 

We should consider how this relates to the broader caching work in PR #4899, these may want to be coordinated.

This change is very low-risk and provides a clear performance benefit for high-volume environments. Work from #4899 can probably wait and be added later integrating this cache into the broader caching work.

There probably isn't any issues, but I will run some tests on our environment and come back to approve.

guicaulada (approved):
Testing showed lower number of calls to the `getParameter` function which resulted in lower lambda duration, from eventual timeouts to a max of 55s for 50 runners batch creation. System seemed more responsive since no orphan runners were created.

PR #4995, opened 2026-01-09
feat: add bypass-removal tag to prevent runner scale-down

Brend-Smits:
Add support for ghr:bypass-removal EC2 tag that allows engineers to manually tag runners to prevent them from being scaled down during debugging or investigation. When this tag is set to 'true', the runner will be skipped during scale-down operations with appropriate logging.

Tested on our staging environment and things are working:
image

This small change makes debugging some issue on the runner significantly easier...

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
@@ -130,6 +130,14 @@ function runnerMinimumTimeExceeded(runner: RunnerInfo): boolean {
 async function removeRunner(ec2runner: RunnerInfo, ghRunnerIds: number[]): Promise<void> {
   const githubAppClient = await getOrCreateOctokit(ec2runner);
   try {
+    const runnerList = ec2runner as unknown as RunnerList;
+    if (runnerList.bypassRemoval) {
+      logger.info(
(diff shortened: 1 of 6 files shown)


lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -130,6 +130,14 @@ function runnerMinimumTimeExceeded(runner: RunnerInfo): boolean {
 async function removeRunner(ec2runner: RunnerInfo, ghRunnerIds: number[]): Promise<void> {
   const githubAppClient = await getOrCreateOctokit(ec2runner);
   try {
+    const runnerList = ec2runner as unknown as RunnerList;
```
npalm, line 133: Can you add a unit test case please?

npalm (changes requested):
Thx

- small request about testing
- would be good the document this option in for example docs/additional_notes.md

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
Brend-Smits: Definitely, next time I will do it by default. Was a bit in a rush this time.

npalm (approved):

PR #4529, opened 2025-04-08
feat: Add feature to enable dynamic instance types via workflow labels

npalm:
## Description

This PR adds an option to define the required instance type via the workflow labels. This could be all tricky, since in the end GH decides which workflo runs on a runner. 

By adding a label to the workflow starting with `ghr-ec2-` and followed by the instance type a runenr for the required instance type is caled (only if the teature is enabled).

For example: `ghr-ec2-m5.xlarge`
image

## Limitiation

- No support for pool (since not scaled by eventts)

### Status

Early draft, just a POC, no tests, no docs, requires rerfactoing.

## Todo

- [ ] Decide if this should enter the main line? If yes
- [ ] Refactor feature config to an object, containing enable and prefix
- [ ] Add docs including limitations
- [ ] Add unit tests
- [ ] Adjust an example, enable feature. Ephemeral case with job retry.
- [ ] Adjust logging

PR #5003, opened 2026-01-19
feat: Add feature to enable dynamic ec2 config via workflow labels

edersonbrilhante:
## Summary

This PR resumes and completes the work started in **#4529**.

It also allows to use any other dynamic labels with prefix `ghr-`. Giving support for unique labels per job or per group of jobs

It ensures that EC2-specific config can be defined via `run-ons`

### How to test:

Use your regular labels, and add ghr-ec2-instance-type and ghr-ec2-image-id

```yaml
run-ons:
  - <regular-labels>
  - ghr-ec2-instance-type:<different-instance-type>
  - ghr-ec2-image-id:<different-ami-id>
```

In this case:

* The runner is resolved from `<regular-labels>`
* The EC2 instance type and AMI ID are taken **exactly** from the provided labels


npalm:
@edersonbrilhante great to see this PR.

andrecastro:
This is a really interesting feature!

Just one suggestion from my side: would it be possible to support a whitelist of allowed instance types?

Also, it could be really powerful to have some kind of feature-flag / policy control over which parts of the configuration are allowed to be dynamic. For example, in my org we don’t want developers to be able to select arbitrary AMIs (only a pre-approved set), but it would be awesome to still allow them to choose the instance type for workflow jobs, as long as it’s constrained to an allowed list.

Maybe the "feature flag" is not even necessary as long as we could define the "allowed values" for each configuration, with this we could list only the pre-approved AMIs.

edersonbrilhante:
@andrecastro I liked and makes a lot of sense to me. I just need more time to think about the implementation. And tbis PR is already big enough XD. I could create a following for adding this restricted values feature

lambdas/functions/webhook/src/runners/dispatch.test.ts
```diff
@@ -181,49 +183,61 @@ describe('Dispatcher', () => {
...
+      expect(canRunJob(workflowLabels, runnerLabels, true, false)).toBe(false);
     });
 
     it('should accept jobs not providing labels and exact match is set to false.', () => {
       const workflowLabels: string[] = [];
       const runnerLabels = [['self-hosted', 'linux', 'x64']];
-      expect(canRunJob(workflowLabels, runnerLabels, false)).toBe(true);
+      expect(canRunJob(workflowLabels, runnerLabels, false, false)).toBe(true);
+    });
+
```
stuartp44, line 231: Reference my previous point, what do I feel is valid here?

lambdas/functions/control-plane/src/aws/runners.test.ts
```diff
@@ -425,6 +426,215 @@ describe('create runner', () => {
       }),
     });
   });
+
```
stuartp44, line 451: I think these tests work with all good values, but because we are in the user space, what about bad values and their effects? Is it maybe worth extending the tests to not trust the user data? I am not sure how the behaviour will be if someone makes a mistake, does it take the whole batch/process out?

stuartp44: An example would be m5.large, could be m5,large. How will this change behaviour?

stuartp44 (commented):
I am happy to approve, but I do have a statement about incorrect labels and the effect on the process.

stuartp44:
I also agree with what was previously mentioned; we probably need a safelist, as we don't want lateral movement when a compromised pipeline is used, especially with the VPC setting. Maybe worth some "allowed_instance_type" setting or something to that effect that can be checked against, and if not in the list, ignored.

lambdas/functions/control-plane/src/aws/runners.test.ts
edersonbrilhante: @stuartp44 

The payload will be sent to aws, and aws will validate it. Example

```
      - self-hosted
      - x64
      - type:small
      - ghr-ec2-instance-type:m5,xlarge
```
 
 ```
{
  "level": "WARN",
  "message": "Create fleet failed, error not recognized as scaling error.",
  "timestamp": "2026-03-09T17:24:59.008Z",
  "service": "test-small-scale-up",
  "sampling_rate": 0,
  "xray_trace_id": "REDACTED_XRAY_TRACE_ID",
  "region": "eu-west-1",
  "environment": "test-small",
  "module": "runners",
  "aws-request-id": "REDACTED_AWS_REQUEST_ID",
  "function-name": "test-small-scale-up",
  "runner": {
    "ephemeral": true,
    "type": "Org",
    "namePrefix": "",
    "n_events": 1
  },
  "data": [
    {
      "LaunchTemplateAndOverrides": {
        "LaunchTemplateSpecification": {
          "LaunchTemplateId": "REDACTED_LAUNCH_TEMPLATE_ID",
          "Version": "22"
        },
        "Overrides": {
          "InstanceType": "m5,xlarge",
          "SubnetId": "REDACTED_SUBNET_ID_1"
        }
      },
      "Lifecycle": "on-demand",
      "ErrorCode": "InvalidFleetConfiguration",
      "ErrorMessage": "Your requested instance type (m5,xlarge) is not supported in your requested Availability Zone (eu-west-1b)."
    },
    {
      "LaunchTemplateAndOverrides": {
        "LaunchTemplateSpecification": {
          "LaunchTemplateId": "REDACTED_LAUNCH_TEMPLATE_ID",
          "Version": "22"
        },
        "Overrides": {
          "InstanceType": "m5,xlarge",
          "SubnetId": "REDACTED_SUBNET_ID_2"
        }
      },
      "Lifecycle": "on-demand",
      "ErrorCode": "InvalidFleetConfiguration",
      "ErrorMessage": "Your requested instance type (m5,xlarge) is not supported in your requested Availability Zone (eu-west-1a)."
    }
  ]
}
```

File: lambdas/functions/webhook/src/runners/dispatch.test.ts (hunk 1 above)
Comment (edersonbrilhante) [reply lambdas/functions/webhook/src/runners/dispatch.test.ts:231]: I removed. `canRunJob` will called by `dispatch` with the github event. It will have the proper workflow labels

**Comment by edersonbrilhante** [comment]
@stuartp44 We can add the safelist, but I think we need a deeper discussion about the implementation

File: lambdas/functions/control-plane/src/aws/runners.test.ts (hunk 2 above)
Comment (edersonbrilhante) [reply lambdas/functions/control-plane/src/aws/runners.test.ts:451]: @stuartp44 can you review again?

**Comment by npalm** [comment]
@edersonbrilhante tried to test the feature but so far not got it wokring

Used labels: `runs-on: [self-hosted, x64, linux, ubuntu-latest, ghr-ec2-instance-type:c5.2xlarge, ghr-ec2-ebs-volume-size:200]` But need to check what is worng. I enabled the feature as well rebuild the lambda stack.

**Comment by edersonbrilhante** [comment]
Can you print the logs from dispatch-to-runner? Check if the labels were accepted or not.

File: lambdas/functions/webhook/src/runners/dispatch.ts [hunk 3]
```diff
@@ -80,14 +89,20 @@ export function canRunJob(
   workflowJobLabels: string[],
   runnerLabelsMatchers: string[][],
   workflowLabelCheckAll: boolean,
+  enableDynamicLabels: boolean,
 ): boolean {
+  // Filter out ghr- and ghr-run- labels only if dynamic labels config is enabled
+  const filteredLabels = enableDynamicLabels
+    ? workflowJobLabels.filter((label) => !label.startsWith('ghr-'))
```
npalm: This is a new convention. And I would prefer we add a comment to the variables that labels starting with `ghr-` will be ingored (main and multi-runenr)

docs/configuration.md
```diff
@@ -328,6 +328,212 @@ Below is an example of the log messages created.
 }
 ```
 
+### Dynamic Labels
+
```
npalm, line 332:
```suggestion

[!WARNING]
**Security implication:** Dynamic labels are extracted from the `runs-on` labels in incoming `workflow_job` webhook events. These labels originate from what 
users define in their workflow files. Any user with permission to create or modify workflows can inject arbitrary EC2 configuration values — including instance types, AMI IDs, subnet IDs, EBS volumes, placement settings, and more. **These values are not sanitized or validated** against an allowlist before being passed to the EC2 CreateFleet API. This means a malicious or careless workflow author could, for example:
- 

- Launch expensive instance types (e.g., `p5.48xlarge`) to inflate costs
- Override the AMI (`ghr-ec2-image-id`) to boot a compromised image
- Target specific subnets (`ghr-ec2-subnet-id`) to escape network boundaries
- Set arbitrarily large EBS volumes (`ghr-ec2-ebs-volume-size:10000`)

**Only enable this feature in repositories where you trust all workflow contributors.** Consider combining it with [GitHub branch protection 
rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-a-branch-rule/about-branch-rules) and required reviews for workflow file changes.

```

lambdas/functions/webhook/src/runners/dispatch.ts
npalm, line 1: I th9ink we can add at least a small safety net here, like:

```
function sanitizeGhrLabels(labels: string[]): string[] {
  const GHR_LABEL_MAX_LENGTH = 128;
  const GHR_LABEL_VALUE_PATTERN = /^[a-zA-Z0-9._\-:\/]+$/;
 
  return labels.map((label) => {
    if (!label.startsWith('ghr-')) return label;
 
    if (label.length > GHR_LABEL_MAX_LENGTH) {
      logger.warn('Dynamic label exceeds max length, stripping', { label: label.substring(0, 40) });
      return null;
    }
    if (!GHR_LABEL_VALUE_PATTERN.test(label)) {
      logger.warn('Dynamic label contains invalid characters, stripping', { label });
      return null;
    }
    return label;
  }).filter((l): l is string => l !== null);
}
```
next someting like
```
const sanitizedLabels = enableDynamicLabels
   ? sanitizeGhrLabels(body.workflow_job.labels)
   : body.workflow_job.labels;
```

npalm (dismissed):
Great feature. Maybe we can add a small saftety net in the disspachter. A clear warning in the docs. And refer in the variable where users can enable to the risk in the docs.

lambdas/functions/webhook/src/runners/dispatch.ts
edersonbrilhante: Good point. I added it.

lambdas/functions/webhook/src/runners/dispatch.ts
edersonbrilhante: I added comment in the tf variables

edersonbrilhante:
@npalm can you review

Brend-Smits (approved):
LGTM, tested it and it seems to work. Documentation looking great.
Thanks a lot! 

Let's improve it from here by adding allow/deny lists 👍🏼

piotrj:
@edersonbrilhante @Brend-Smits I'm working on getting rid of conflicts in #5077 and there is a `Priority` value that kind of clashes with this feature. 

I wonder what's the use case of Priority being set by dynamic label? Priority makes sense mostly (and possibly only) when you define an EC2 fleet that has multiple instance_types and have prioritised allocation type. Then during a scale-up it tries to launch the instance type with the highest priority (and then going down the list in case of no available capacity).

For the case of dynamic labels where you create a separate fleet and that fleet has only 1 instance type it doesn't really make sense to define the priority.

edersonbrilhante:
@piotrj the logic is the same as today: "the first match", then uses the dynamic labels to override the launch template.
Your feature is compatible. The non-dynamic labels will find the first match, using the priority config, then it will override the ec2 configs passed by `ec2-*`

PR #5027, opened 2026-02-11
fix(install-runner.sh): support Debian

gmccue:
Currently, when using the Debian OS, the existing `install-runner.sh` script will take an additional 25s during start up as the script attempts to install `libicu` via `dnf`

This PR adds support for the `debian` OS which is listed as a supported OS in the GitHub docs: https://docs.github.com/en/actions/reference/runners/self-hosted-runners#linux


gmccue:
@npalm @Brend-Smits apologies for the `@mention` here, but any chance I could get a review on this simple PR?

Brend-Smits (approved):

Brend-Smits:
Thanks for the PR @gmccue! 
Approved, :shipit:

PR #4990, opened 2026-01-07
fix: gracefully handle JIT config failures and terminate unconfigured instance

Brend-Smits:
This pull request enhances the robustness and reliability of the GitHub Actions runner scaling logic by improving error handling and retry mechanisms for GitHub API calls. It introduces the `@octokit/plugin-retry` plugin to automatically retry failed API requests, adds detailed logging for retry attempts, and ensures that failures in creating JIT configs for individual runner instances do not halt the entire scaling process. Additionally, new tests are added to verify handling of various API failure scenarios.

**GitHub API client improvements:**

* Added `@octokit/plugin-retry` to dependencies (`package.json`) and integrated it into the Octokit client initialization to enable automatic retries for failed GitHub API requests. [[1]](diffhunk://#diff-37d09418dae74ded5678eabfa3b60993ee491e2fd9e49e11142f639b078ac9b2R41) [[2]](diffhunk://#diff-cf7cdd79fe0ed0e3a2e8928c0c7667a096c47c47abdb2354ddadee67e80a226dR21) [[3]](diffhunk://#diff-cf7cdd79fe0ed0e3a2e8928c0c7667a096c47c47abdb2354ddadee67e80a226dL29-R30)
* Configured the retry plugin to log detailed warnings on each retry attempt, including the HTTP method, URL, error message, and status code.

**Error handling and resilience in JIT config creation:**

* Updated `createJitConfig` in `scale-up.ts` to catch and log errors for individual runner instances when creating JIT configs, allowing the process to continue for remaining instances and logging a summary of failed attempts at the end. [[1]](diffhunk://#diff-fbc68af2a40bf14ad13a80b13958c0b52d1d0fde5f0009416a693fb4b691ceaeR537-R542) [[2]](diffhunk://#diff-fbc68af2a40bf14ad13a80b13958c0b52d1d0fde5f0009416a693fb4b691ceaeR582-R596)
* Instances that failed to generate a configuration, will now be terminated to avoid generating waste.

**Testing improvements:**

* Added comprehensive tests to `scale-up.test.ts` to verify correct behavior when GitHub API calls fail for some instances, including retryable errors (e.g., 5xx), non-retryable errors (e.g., 4xx), and partial failures, ensuring only successful JIT configs are stored.

PR #5029, opened 2026-02-12
fix: retry SQS messages when GitHub API fails after EC2 instance creation

### LINKED ISSUE #5024: Scale-up Lambda silently drops entire SQS batch on non-ScaleError exceptions
## Description

When the scale-up Lambda encounters a non-`ScaleError` exception (such as a transient HTTP 404 from the GitHub App installation token endpoint), it logs the error but returns an **empty `batchItemFailures`** response. This tells SQS that all messages in the batch were processed successfully, causing them to be permanently deleted from the queue. The corresponding `workflow_job` events are silently lost — no EC2 instances are launched, no JIT configs are created, and no retry messages are published. The affected jobs remain `queued` in GitHub until they time out (24 hours).

## Root Cause

In `lambda.ts`, the `scaleUpHandler` catch block distinguishes between `ScaleError` (EC2 fleet creation failures) and all other exceptions:

```typescript
} catch (e) {
    if (e instanceof ScaleError) {
      // ✅ Returns batch failures → SQS retries these messages
      batchItemFailures.push(...e.toBatchItemFailures(sqsMessages));
      logger.warn(`${e.detailedMessage} A retry will be attempted via SQS.`, { error: e });
    } else {
      // ❌ Logs error but returns EMPTY batchItemFailures → SQS deletes all messages
      logger.error(
        `Error processing batch (size: ${sqsMessages.length}): ${(e as Error).message}, ignoring batch`,
        { error: e },
      );
    }
    return { batchItemFailures };
}
```

When a non-`ScaleError` exception occurs (e.g., `HttpError` from Octokit), `batchItemFailures` remains empty, and SQS treats all messages as successfully processed. The phrase "ignoring batch" in the log is misleading — the messages aren't just ignored temporarily, they're permanently lost.

There are two failure modes depending on when the error occurs:

- **Before EC2 instances are created** (e.g., during `createGithubInstallationAuth()`): No orphaned instances, but the webhook events for those jobs are gone and the jobs will never get a runner.
- **After some instances are created but before JIT config is written** (e.g., during `createStartRunnerConfig()`): Instances boot with no JIT config in SSM. The `start-runner.sh` polls SSM for some time, then gives up and self-terminates. 

## Impact

- **Jobs permanently lost**: Affected `workflow_job` events are deleted from SQS with no retry. Jobs remain `queued` in GitHub for up to 24 hours before timing out.
- **No observability**: The error log says "ignoring batch" but there's no metric, alarm, or dead-letter queue to catch this. The `job_retry` mechanism is never triggered because the retry message is only published after a successful scale-up.
- **Wasted EC2 instances**: If the failure occurs after instance creation, those instances boot with no SSM config and eventually self-terminate.

## How to Trigger

Any non-`ScaleError` exception during `scaleUp()` will trigger this. The most common cause we've observed is a transient HTTP 404 from the GitHub App installation token endpoint under concurrent load (see related issue: JWT collisions).

## Environment

- Module version: `~> 7.3`
- GitHub: Enterprise Cloud with Data Residency (`ghe.com`)
- `enable_jit_config = true`, `enable_ephemeral_runners = true`

### Fix 1: Return all messages as batch failures on unhandled errors

The generic catch block should return all messages as `batchItemFailures` instead of acknowledging them:

```typescript
} catch (e) {
    if (e instanceof ScaleError) {
      batchItemFailures.push(...e.toBatchItemFailures(sqsMessages));
      logger.warn(`${e.detailedMessage} A retry will be attempted via SQS.`, { error: e });
    } else {
      logger.error(
        `Error processing batch (size: ${sqsMessages.length}): ${(e as Error).message}, returning batch for retry`,
        { error: e },
      );
      batchItemFailures.push(
        ...sqsMessages.map(({ messageId }) => ({ itemIdentifier: messageId })),
      );
    }
    return { batchItemFailures };
}
```

The SQS redrive policy / maxReceiveCount will eventually move persistently failing messages to the DLQ, preventing infinite retries.

### Fix 2: Enable Octokit rate limit retry

The `onAbuseLimit` and `onRateLimit` handlers in `createAuth()` log the rate limit but don't return `true`, so Octokit never retries:

```typescript
onRateLimit: (retryAfter, options, octokit, retryCount) => {
  logger.warn(`Rate limit hit, retrying after ${retryAfter} seconds`);
  return true; // ← was missing
},
onSecondaryRateLimit: (retryAfter, options, octokit, retryCount) => {
  logger.warn(`Secondary rate limit hit, retrying after ${retryAfter} seconds`);
  return true; // ← was missing
},
```

## Our Workaround

We have patched the generic catch to return all messages as `batchItemFailures` and enabled Octokit rate limit retry by returning `true` from the handlers.

edersonbrilhante:
Closes #5024

When createStartRunnerConfig fails with a GitHub HTTP error (e.g., 401, 404, 422) after EC2 instances have already been created, the SQS messages are consumed and permanently deleted. The workflow jobs are silently lost — runners can't register, and jobs remain queued in GitHub until they time out (24 hours).

This change introduces GHHttpError, a subclass of ScaleError, that catches Octokit HttpError exceptions in createStartRunnerConfig and wraps them. Because GHHttpError extends ScaleError, the existing ScaleError catch in scaleUpHandler handles it via polymorphism — no changes needed in lambda.ts. Unlike ScaleError (which retries only failedInstanceCount messages), GHHttpError overrides toBatchItemFailures() to retry all messages, since the GitHub API failure affects every instance in the batch.

### DIFF
--- a/lambdas/functions/control-plane/src/lambda.test.ts
+++ b/lambdas/functions/control-plane/src/lambda.test.ts
@@ -3,7 +3,7 @@ import { Context, SQSEvent, SQSRecord } from 'aws-lambda';
 
 import { addMiddleware, adjustPool, scaleDownHandler, scaleUpHandler, ssmHousekeeper, jobRetryCheck } from './lambda';
 import { adjust } from './pool/pool';
-import ScaleError from './scale-runners/ScaleError';
+import ScaleError, { GHHttpError } from './scale-runners/ScaleError';
 import { scaleDown } from './scale-runners/scale-down';
 import { ActionRequestMessage, scaleUp } from './scale-runners/scale-up';
 import { cleanSSMTokens } from './scale-runners/ssm-housekeeper';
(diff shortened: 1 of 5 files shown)


Brend-Smits:
This seems similar to #4990, can you please have a look?

edersonbrilhante:
@Brend-Smits your implementation is so much better than mine.

PR #5054, opened 2026-03-03
feat(runner): add source parameter to distinguish between scale-up and pool lambda

### LINKED ISSUE #5053: EC2 ghr:created_by tag is getting set to "pool-lambda" even though there is no pool set for the runner type
We have been noticing that some of the EC2 instances, the control plane spawns, are getting to the ghr:created_by tag set to random values of "scale-up-lambda" and "pool-lambda" even though they are being created by only the "scale-up-lambda" as a pool does not exist for the runner type we have noticed the behaviour in.

Looking at the code, 
https://github.com/github-aws-runners/terraform-aws-github-runner/blob/6dc97d55d7b01c7c197573843b298236f891cda8/lambdas/functions/control-plane/src/aws/runners.ts#L244
Looks like the issue is that the default is set to "pool-lambda": currently, if x.numberofrunners is 1, it's the "scale-up" lambda; otherwise, it sets the default to "pool-lambda". Since https://github.com/github-aws-runners/terraform-aws-github-runner/releases/tag/v7.0.0, we now process in batches, which means we are almost always doing actions in multiple rather than single (scale-up) and multiple (pool).

We really need to allign this with the https://github.com/github-aws-runners/terraform-aws-github-runner/releases/tag/v7.0.0 change

stuartp44:
Fixes #5053

This pull request updates the way runner creation is tracked and tagged by introducing an explicit `source` parameter to distinguish between runners created by the scale-up lambda and those created by the pool lambda. The changes ensure that the runner creation flow and related tests consistently use this new `source` field.

**Runner creation and tagging improvements:**

* Added a `source` field (`'scale-up-lambda' | 'pool-lambda'`) to `RunnerConfig`, `RunnerInputParameters`, and related test interfaces to explicitly track the origin of runner creation. [[1]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR756) [[2]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR777) [[3]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR789-R797)
* Updated the runner creation logic in `createInstances` and `createRunners` to use the new `source` field for the `ghr:created_by` tag, replacing previous logic based on runner count. [[1]](diffhunk://#diff-69076cf041b0bcf70b90ba1dde121bfc3cb3b57bc9d293d29f7da72affcf0041L244-R244) [[2]](diffhunk://#diff-fbc68af2a40bf14ad13a80b13958c0b52d1d0fde5f0009416a693fb4b691ceaeR253-R259)

**Test enhancements:**

* Modified and added tests in `runners.test.ts` to verify correct behaviour when the `source` parameter is specified, including both single and multiple instance fleet creation. [[1]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR321-R330) [[2]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR371-R389) [[3]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR450-R471) [[4]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR480-R487) [[5]](diffhunk://#diff-a41db796ea6eeeb679f8cda3eb91243ffc16e68b8ebbeaea6837fbf85e7241baR595-R603)
* Updated pool adjustment tests to pass the `source` argument and verify its usage in runner creation calls. [[1]](diffhunk://#diff-f39f9287eeda57c37c329d28d6e6463da19db37f71791bf7da0181191b95857aL195-R201) [[2]](diffhunk://#diff-f39f9287eeda57c37c329d28d6e6463da19db37f71791bf7da0181191b95857aL271-R283) [[3]](diffhunk://#diff-f39f9287eeda57c37c329d28d6e6463da19db37f71791bf7da0181191b95857aL286-R304) [[4]](diffhunk://#diff-f39f9287eeda57c37c329d28d6e6463da19db37f71791bf7da0181191b95857aL336-R360)

**Pool and scaling logic:**

* Changed the pool adjustment logic in `pool.ts` to always pass `'pool-lambda'` as the `source` when topping up the pool.
* Ensured scale-up runner logic in `scale-up.test.ts` and related code passes `'scale-up-lambda'` as the `source` for created runners.

These changes collectively improve the clarity and reliability of runner source tracking throughout the control-plane codebase.

### DIFF
--- a/lambdas/functions/control-plane/src/aws/runners.test.ts
+++ b/lambdas/functions/control-plane/src/aws/runners.test.ts
@@ -318,6 +318,8 @@ describe('create runner', () => {
     allocationStrategy: SpotAllocationStrategy.CAPACITY_OPTIMIZED,
     capacityType: 'spot',
     type: 'Org',
+    scaleErrors: [],
+    source: 'scale-up-lambda',
   };
 
   const defaultExpectedFleetRequestValues: ExpectedFleetRequestValues = {
@@ -325,6 +327,7 @@ describe('create runner', () => {
     capacityType: 'spot',
     allocationStrategy: SpotAllocationStrategy.CAPACITY_OPTIMIZED,
     totalTargetCapacity: 1,
+    source: 'scale-up-lambda',
   };
 
   beforeEach(() => {
@@ -361,6 +364,25 @@ describe('create runner', () => {
       ...expectedCreateFleetRequest({
         ...defaultExpectedFleetRequestValues,
         totalTargetCapacity: 2,
+      }),
+    });
+  });
+
+  it('calls create fleet of multiple instances with pool-lambda source when specified', async () => {
+    const instances = [{ InstanceIds: ['i-1234', 'i-5678', 'i-9012'] }];
+
+    mockEC2Client.on(CreateFleetCommand).resolves({ Instances: instances });
+
+    await createRunner({
+      ...createRunnerConfig({ ...defaultRunnerConfig, source: 'pool-lambda' }),
+      numberOfRunners: 3,
+    });
+
+    expect(mockEC2Client).toHaveReceivedCommandWith(CreateFleetCommand, {
+      ...expectedCreateFleetRequest({
+        ...defaultExpectedFleetRequestValues,
+        totalTargetCapacity: 3,
+        source: 'pool-lambda',
       }),
     });
   });
@@ -425,6 +447,28 @@ describe('create runner', () => {
       }),
     });
   });
+
+  it('calls create fleet with source set to scale-up-lambda when source is specified', async () => {
+    await createRunner(createRunnerConfig({ ...defaultRunnerConfig, source: 'scale-up-lambda' }));
+
+    expect(mockEC2Client).toHaveReceivedCommandWith(CreateFleetCommand, {
+      ...expectedCreateFleetRequest({
+        ...defaultExpectedFleetRequestValues,
+        source: 'scale-up-lambda',
+      }),
+    });
+  });
+
+  it('calls create fleet with source set to pool-lambda when source is specified', async () => {
+    await createRunner(createRunnerConfig({ ...defaultRunnerConfig, source: 'pool-lambda' }));
+
+    expect(mockEC2Client).toHaveReceivedCommandWith(CreateFleetCommand, {
+      ...expectedCreateFleetRequest({
+        ...defaultExpectedFleetRequestValues,
+        source: 'pool-lambda',
+      }),
+    });
+  });
 });
 
 describe('create runner with errors', () => {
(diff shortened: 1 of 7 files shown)


edersonbrilhante (approved):

npalm:
@stuartp44 can you make also clear what bug / issue is addressed. I think the batch intruduced to handle scaling is causing runners are marekd with pool instead of scale-up

stuartp44:
@npalm its in the issue attached. TLDR, it's because we have some runner types that do not have pools but are getting tagged with created by "pool-lambda", which is incorrect.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -250,11 +250,13 @@ export async function createRunners(
   ec2RunnerConfig: CreateEC2RunnerConfig,
   numberOfRunners: number,
   ghClient: Octokit,
+  source: 'scale-up-lambda' | 'pool-lambda' = 'scale-up-lambda',
```
guicaulada: I noticed we repeat `'scale-up-lambda' | 'pool-lambda'` in quite a few places.

Maybe we could benefit from a `export type RunnerSource = 'scale-up-lambda' | 'pool-lambda'` so we can update a single place if new sources are added in the future?

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
guicaulada, line 482: Here we call create runners without a source, relying on the default parameter, passing it explicitly would make the intent clearer and be consistent with `pool.ts`.

lambdas/functions/control-plane/src/aws/runners.test.ts
```diff
@@ -546,12 +592,15 @@ describe('create runner with errors fail over to OnDemand', () => {
     capacityType: 'spot',
     type: 'Repo',
     onDemandFailoverOnError: ['InsufficientInstanceCapacity'],
+    scaleErrors: [],
```
guicaulada, line 596: I noticed `scaleErrors` was already required on `RunnerConfig`, but I'm not sure why it's being added just now, is this some pre-existing issue?

guicaulada (dismissed):
Looks good to me! The change is well-scoped and well-tested.

Making source explicit feels better than relying on `numberOfRunners`.

I have a few minor suggestions and questions, but no blockers.

If you could link the issue directly on the PR body (e.g. Fixes https://github.com/github-aws-runners/terraform-aws-github-runner/issues/5053) it would make it easier to find.

lambdas/functions/control-plane/src/aws/runners.test.ts
stuartp44: This is a required property thus it was added, without it, it fails

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
stuartp44: resolved

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
stuartp44: Resolved

npalm (approved):

PR #5056, opened 2026-03-06
fix(lambda): add jti claim to GitHub App JWTs to prevent concurrent collisions

### LINKED ISSUE #5025: Concurrent Lambda invocations generate byte-identical JWTs causing transient 404 on installation token endpoint
## Description

When multiple scale-up Lambda invocations run concurrently (common during burst workloads with 100+ `workflow_job` events), the GitHub App JWT generation produces **byte-identical tokens**. GitHub rejects these duplicates (likely replay protection), causing `POST /app/installations/{id}/access_tokens` to return HTTP 404. This triggers silent batch dropping (see related issue: silent batch drop) and permanently loses the affected jobs.

## Root Cause

The `@octokit/auth-app` library (via `universal-github-app-jwt`) generates JWTs with only `{ iat, exp, iss }` claims — no `jti` (JWT ID). The `iat` uses seconds precision (`Math.floor(Date.now() / 1000)`). When multiple Lambda invocations generate JWTs within the same second using the same App ID and private key, they produce **byte-identical tokens**.

GitHub rejects duplicate JWTs, causing the `POST /app/installations/{id}/access_tokens` request to be treated as unauthenticated. An unauthenticated request to this endpoint returns HTTP 404 (GitHub won't confirm resource existence to unauthorized callers).

The 404 is transient — the same installation ID succeeds on subsequent requests seconds later.

## Observed Error

```json
{
  "level": "ERROR",
  "message": "Error processing batch (size: 4): Not Found, ignoring batch",
  "error": {
    "name": "HttpError",
    "status": 404,
    "request": {
      "method": "POST",
      "url": "https://api.<redacted>.ghe.com/app/installations/<redacted>/access_tokens"
    },
    "response": { "status": 404, "data": { "message": "Not Found" } }
  }
}
```

This repeats for multiple concurrent invocations in rapid succession while other invocations at the same time succeed with the same installation ID.

## Impact

- Combined with the silent batch drop issue, this permanently loses SQS messages and their corresponding jobs.
- The failure rate correlates with burst size and number of runner type configurations. More configurations = more concurrent Lambdas = higher probability of same-second JWT generation.
- Small workloads work fine; large matrix workflows trigger this consistently.

## Environment

- Module version: `~> 7.3`
- GitHub: Enterprise Cloud with Data Residency (`ghe.com`)
- Multi-runner module with 12+ runner type configurations

### Fix 1: Add `jti` claim to JWT generation

Add a unique `jti` (JWT ID) claim to the JWT payload to prevent byte-identical tokens. This can be done via `@octokit/auth-app`'s `createJwt` callback:

```typescript
import { randomUUID } from 'node:crypto';

const auth = createAppAuth({
  appId,
  createJwt: async ({ appId, privateKey }) => {
    const now = Math.floor(Date.now() / 1000);
    const payload = { iat: now - 60, exp: now + 600, iss: appId, jti: randomUUID() };
    // ... sign with privateKey
  },
});
```

This eliminates the root cause of the transient 404s.

Alternatively, this could be fixed upstream in `universal-github-app-jwt` by always including a `jti` claim.

### Fix 2: Retry installation token API call with backoff

As a defense-in-depth measure, treat HTTP 404 on `POST /app/installations/{id}/access_tokens` as a transient error and retry within `createGithubInstallationAuth()`:

```typescript
const installationId = await getInstallationId(githubAppClient, enableOrgLevel, payload);
let ghAuth, retries = 0;
while (retries < 3) {
  try {
    ghAuth = await createGithubInstallationAuth(installationId, ghesApiUrl);
    break;
  } catch (e) {
    if (e.status === 404 && retries < 2) {
      retries++;
      await new Promise(r => setTimeout(r, 1000 * retries));
      continue;
    }
    throw e;
  }
}
```

## Our Workaround

We have applied both fixes locally:
1. Custom `createJwt` callback with `jti` claim via `crypto.randomUUID()`
2. Retry with backoff on 404 in `createGithubInstallationAuth()`

jensenbox:
## Summary

Fixes concurrent JWT collisions that cause silent job loss during burst workloads.

When multiple scale-up Lambda invocations generate GitHub App JWTs within the same second, `universal-github-app-jwt` produces byte-identical tokens (same `iat`, `exp`, `iss`, no `jti`). GitHub rejects the duplicates, returning HTTP 404 on `POST /app/installations/{id}/access_tokens`, which triggers silent batch dropping.

### Root cause

`universal-github-app-jwt` generates JWTs with only `{ iat, exp, iss }` claims. The `iat` uses seconds precision (`Math.floor(Date.now() / 1000)`). With the same App ID and private key, concurrent invocations within the same second produce identical tokens.

### Fix

Replace `privateKey`-based auth with a custom `createJwt` callback — a first-class API in `@octokit/auth-app` v8.x that completely bypasses `universal-github-app-jwt`. The callback:

- Signs JWTs using `node:crypto.createSign` (zero new dependencies)
- Includes a `crypto.randomUUID()` `jti` claim, ensuring every token is unique
- Preserves the existing `iat`/`exp` logic (30s safety margin, 10-minute expiry)
- Properly forwards the `timeDifference` parameter for clock drift correction
- Supports both PKCS#1 and PKCS#8 private key formats (via `node:crypto`)

### Changes

- `lambdas/functions/control-plane/src/github/auth.ts` — replace `privateKey` with `createJwt` callback in `createAuth()`
- `lambdas/functions/control-plane/src/github/auth.test.ts` — update tests to assert `createJwt` instead of `privateKey`, add test verifying unique JWTs with `jti`

### Test coverage

- Existing tests updated to verify `createJwt` callback is passed instead of `privateKey`
- New test generates two JWTs in rapid succession and verifies they differ (proving `jti` uniqueness)
- New test validates JWT structure (header.payload.signature) and verifies `jti`, `iat`, `exp`, `iss` claims are present
- All 343 control-plane tests pass

Fixes #5025

### DIFF
--- a/lambdas/functions/control-plane/src/github/auth.test.ts
+++ b/lambdas/functions/control-plane/src/github/auth.test.ts
@@ -3,6 +3,7 @@ import { StrategyOptions } from '@octokit/auth-app/dist-types/types';
 import { request } from '@octokit/request';
 import { RequestInterface, RequestParameters } from '@octokit/types';
 import { getParameter } from '@aws-github-runner/aws-ssm-util';
+import { generateKeyPairSync } from 'node:crypto';
 import * as nock from 'nock';
 
 import { createGithubAppAuth, createOctokitClient } from './auth';
@@ -77,23 +78,12 @@ describe('Test createGithubAppAuth', () => {
     process.env.ENVIRONMENT = ENVIRONMENT;
   });
 
-  it('Creates auth object with line breaks in SSH key.', async () => {
+  it('Creates auth object with createJwt callback including jti claim', async () => {
     // Arrange
-    const authOptions = {
-      appId: parseInt(GITHUB_APP_ID),
-      privateKey: `${decryptedValue}
-${decryptedValue}`,
-      installationId,
-    };
-
-    const b64PrivateKeyWithLineBreaks = Buffer.from(decryptedValue + '\n' + decryptedValue, 'binary').toString(
-      'base64',
-    );
-    mockedGet.mockResolvedValueOnce(GITHUB_APP_ID).mockResolvedValueOnce(b64PrivateKeyWithLineBreaks);
+    mockedGet.mockResolvedValueOnce(GITHUB_APP_ID).mockResolvedValueOnce(b64);
 
     const mockedAuth = vi.fn();
     mockedAuth.mockResolvedValue({ token });
-    // Add the required hook method to make it compatible with AuthInterface
     const mockWithHook = Object.assign(mockedAuth, { hook: vi.fn() });
     mockedCreatAppAuth.mockReturnValue(mockWithHook);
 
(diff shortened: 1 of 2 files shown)


Brend-Smits (approved):
Hey! 
Thanks for this great contribution 🚀 
Tested this in our testing environment and things seem to work well. Okay from me (and @stuartp44) side.

PR #5064, opened 2026-03-11
fix: prevent EC2 termination when GitHub runner de-registration fails

shivdesh:
## Summary
This PR complements #4990 by ensuring that when GitHub runner de-registration fails (even after automatic retries), the EC2 instance is **not terminated**. This prevents stale runner entries in GitHub org settings.

## Problem
When the scale-down Lambda fails to de-register a runner from GitHub (e.g., due to persistent API errors), the current code still terminates the EC2 instance. This leaves stale/offline runner entries in the GitHub organization settings.

We encountered this issue in production where transient 502 errors during scale-down left 117 stale runner entries in our GitHub organization.

## Relationship to #4990
PR #4990 added `@octokit/plugin-retry` which provides automatic retries at the Octokit client level. This is great for handling transient failures. However, if de-registration **ultimately fails** after all retries, we still need to handle that gracefully by NOT terminating the EC2 instance.

## Solution
- Extract `deleteGitHubRunner()` helper that catches errors per-runner
- Only terminate EC2 instance if **all** GitHub de-registrations succeed
- If any de-registration fails, leave the instance running so the next scale-down cycle can retry

## Changes
- `lambdas/functions/control-plane/src/scale-runners/scale-down.ts`:
  - Added `deleteGitHubRunner()` helper function
  - Modified `removeRunner()` to only terminate EC2 if all de-registrations succeed

## Testing
- Added test verifying EC2 is NOT terminated when de-registration throws an error
- All 124 scale-down tests pass

## Why not custom retry logic?
The `@octokit/plugin-retry` (added in #4990) already handles automatic retries at the client level, so no custom retry logic is needed. This PR focuses solely on the **failure handling** aspect - what to do when de-registration fails after all retries.

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
@@ -631,6 +631,35 @@ describe('Scale down runners', () => {
         await expect(scaleDown()).resolves.not.toThrow();
       });
 
+      it(`Should not terminate instance when de-registration throws an error.`, async () => {
+        // setup - runner should NOT be terminated because de-registration fails
+        const runners = [createRunnerTestData('idle-1', type, MINIMUM_TIME_RUNNING_IN_MINUTES + 1, true, false, false)];
+
+        const error502 = new RequestError('Server Error', 502, {
+          request: {
+            method: 'DELETE',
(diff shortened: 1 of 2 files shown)


shivdesh:
Closing in favor of updating PR #5061 with the same simplified approach.

PR #5078, opened 2026-03-26
feat(pool): Allow to optionally include busy runners into the pool

rvem:
## Description
'pool_config' is useful when one wants to ensure that a certain number
of runners is running within given time periods.
However, when 'pool_config' is used in a combination with persistent runners,
the pool lambda tends to spin up new runner instance when it sees that some
existing runners are online, but busy.

As a result, with frequent pool lambda checks, the total size of the runners
pool tends to grow more than desired (without respecting
'runners_maximum_count', which is only considered by the 'scale_up' lambda).

These changes introduce a new 'pool_include_busy_runners' module variable
to make it possbile to include all online runners (both idle and busy)
into the pool, so that the pool's lambda only tops up the pool if not
enough runners are online.

## Test Plan

Tested on a module instance with persistent runners, also added a few unit tests to the pool lambda itself.


edersonbrilhante (changes requested):
@rvem Can you fix the conflicts?

rvem:
@edersonbrilhante done

edersonbrilhante:
@rvem Can you sign your commit?

rvem:
Done; also rebased on the latest upstream `main`

edersonbrilhante (approved):

PR #5086, opened 2026-03-30
fix(lambda): add post-deregister busy check to prevent terminating active runners

### LINKED ISSUE #5085: Scale-down can terminate a runner that picks up a job between busy check and termination

```
```

3. Terminate the EC2 instance

### Suggested fix

### Environment

JVenberg:
## Summary

Fixes #5085

The scale-down lambda can terminate an EC2 instance while it's actively running a job. This happens because a job can be assigned to a runner between checking its busy state and calling `TerminateInstances`.

We hit this in production: a Helm deploy was killed mid-execution when the scale-down lambda terminated the instance 13 seconds after the job started. CloudTrail confirmed the `TerminateInstances` call came from the scale-down lambda at the exact moment the runner received a shutdown signal.

## The race condition

Current flow in `removeRunner`:

1. Check GitHub API: "Is this runner busy?" -> `false`
2. _A job gets assigned to the runner here_
3. Deregister the runner from GitHub
4. Terminate the EC2 instance
5. The in-flight job is killed

## The fix

Add a post-deregistration busy re-check:

1. **Check busy** (fast-path to skip obviously busy runners)
2. **Deregister from GitHub** (prevents new job assignment server-side)
3. **Re-check busy** (now stable, since no new jobs can be assigned after deregistration)

## Why this is safe

Deregistering a runner does not affect in-flight jobs. The runner worker uses job-scoped OAuth credentials from the job message, not the runner registration:

- [`JobRunner.cs` lines 80-95](https://github.com/actions/runner/blob/main/src/Runner.Worker/JobRunner.cs): the worker creates its own `VssConnection` using `systemConnection` credentials from the job message
- The worker never checks runner registration status during execution
- Deregistration only affects the listener (no new job pickup), not the worker (current job)

## Test plan

- [x] All 130 existing tests pass
- [x] New test: runner that becomes busy between deregister and re-check is NOT terminated
- [x] New test: runner that returns 404 on post-deregister busy check IS terminated (runner fully removed from GitHub)


npwolf:
We independently hit the underlying race condition in production today and posted repro details on the linked issue (#5085 comment). This fix looks correct to us — the reorder (deregister → recheck busy) closes the TOCTOU window since in-flight jobs use job-scoped credentials that don't depend on registration state.

This has been open since March without review — @edersonbrilhante @Brend-Smits @npalm would one of you be able to take a look? Happy to help test if useful.

edersonbrilhante (changes requested):
@npwolf Can you fix the conflicts?

npwolf:
Rebased this on current `main` to resolve the merge conflicts (mainly around the `deleteGitHubRunner` per-runner error handling that landed since this was opened) — opened as #5201 to help get this moving. All credit to @JVenberg for the original fix and tests. @edersonbrilhante check out #5201

npwolf:
🤖 Follow-up on #5201 (the rebase of this PR): ran an automated backend + security review of the mechanism, and both independently flagged the same core issue, which I verified against the code myself.

**The post-deregister busy re-check can't actually observe "busy" in production.** Once step 2 deregisters the runner, a subsequent GET on that runner ID returns 404 — and this codebase's `getGitHubRunnerBusyState` maps 404 → "not busy". So the re-check basically always reports idle and the instance gets terminated regardless of whether a job was assigned during the race window. And in the rare case the re-check *does* see busy (an eventually-consistency artifact), the orphan-cleanup path terminates the runner within ~2 scale-down cycles anyway, since it treats 404 as "safe to kill" too.

Net effect: this fix doesn't close the TOCTOU race it targets — an in-flight job can still be killed, just via a different path (or slightly delayed via orphan cleanup instead of immediately). The new test added for this also mocks a `{ busy: true }` response on the post-deregister check, which the real API can't return after a successful delete (it returns 404), so it doesn't actually exercise the production behavior.

Full findings with file/line detail are posted as inline comments on #5201. Flagging here too since the mechanism gap applies to this PR as originally written, not just the rebase. Would be good to get eyes on whether deregistering *before* confirming idle is fundamentally the wrong order, or whether there's a way to keep the busy signal alive post-deregistration.

edersonbrilhante:
@JVenberg I fixed the conflicts, but I need you to sign your commit.

edersonbrilhante:
@JVenberg Can you fix the conflicts? Heads up, I am working to refactor the module to introduce plugin layout. Ref: #5234

edersonbrilhante:
@JVenberg Can you please fix the conflict?

PR #5096, opened 2026-04-08
feat: Enterprise-level runners

dmitrykiselev27:
## Description

Adds enterprise-level runner registration support. Previously, runners could only be registered at the repository or organization level using GitHub App authentication. This PR introduces a third registration level — enterprise — using PAT-based authentication against the GitHub Enterprise runner management APIs. 

### Key changes
**Terraform**
- New `runner_registration_level` variable ("repo" | "org" | "enterprise") replaces the boolean `enable_organization_runners`
- New `enterprise_slug` and `enterprise_pat` variables for enterprise configuration. Supports comma-separated multiple PATs for rate limit distribution.
- `github_app` no longer requires `key_base64` or `id` when `runner_registration_level = "enterprise"` — only `webhook_secret` is needed. No GitHub App creation required for enterprise runners.

**Lambda functions**
- Enterprise code paths added for runner registration, listing, busy-state checks, de-registration, and termination — all using raw `client.request()` since Octokit's typed helpers don't cover enterprise endpoints
- New `createEnterprisePATClient()` in `auth.ts` reads PAT from SSM and randomly selects one token from a comma-separated list per invocation
- New `resolveRunnerType()` helper reads `RUNNER_REGISTRATION_LEVEL` env var with backward-compatible fallback to the legacy `ENABLE_ORGANIZATION_RUNNERS`

## Test Plan

1. Create PAT with the `manage_runners:enterprise` scope. 
2. Create webhook on Enterprise level. Events: select `Workflow jobs` only
3. Set required Terraform variables:
```hcl
runner_registration_level = "enterprise"
enterprise_slug = "<your-enterprise-slug>"   
enterprise_pat = {
  pat = "<your-PAT>"                                  
}
```
4. Deploy the module 
5. Trigger GHA workflow to test


edersonbrilhante:
What is the rate limit for this type of pat? Maybe allow multiple pat and set some round robin logic. There is an open pr to support multiple apps for this reason.

dmitrykiselev27:
@edersonbrilhante as usual, 5000 requests/hour. For this reason, in this implementation `pat` string can be a comma-separated list.

When a function creates `EnterprisePATClient` it takes a random token from the list. 
It's not the most clever logic, I know, but to properly implement round robin logic, I would need to store some state outside of Lambda function (there is no guarantee Lambda will reuse container for the next execution, also there may be multiple concurrent executions). But I'm open to feedback on this topic

Brend-Smits:
What's the reason for supporting PAT's and not just enterprise github app's?

dmitrykiselev27:
@Brend-Smits unfortunately, GitHub App on enterprise level does not allow to manage runners. 

image

Brend-Smits:
> @Brend-Smits unfortunately, GitHub App on enterprise level does not allow to manage runners.
> 
> image

Okay thanks! I will try to arrange a test enterprise to be able to test this and get back to you 👍🏼

examples/default/main.tf
```diff
@@ -50,8 +50,8 @@ module "runners" {
   # runner_binaries_syncer_lambda_zip = "../lambdas-download/runner-binaries-syncer.zip"
   # runners_lambda_zip                = "../lambdas-download/runners.zip"
 
-  enable_organization_runners = true
-  runner_extra_labels         = ["default", "example"]
+  runner_registration_level = "org"
```
Brend-Smits, line 53: I like this change, but maybe instead of completely removing the 'enable_organization_runners' and making this a breaking change, instead we can just deprecate that option and move people to this new option instead?
In the mean time we can keep the old variable working. If that is used, then it takes priority or something?

edersonbrilhante (changes requested):
@dmitrykiselev27 can you fix the conflicts?

examples/default/main.tf
dmitrykiselev27: Makes sense. Deprecated `enable_organization_runners` instead of breaking it: it still works, and if set to `true`, it takes priority over the new `runner_registration_level` for backwards compatibility.

dmitrykiselev27:
@edersonbrilhante rebased and fixed the conflicts

edersonbrilhante:
> "Commits must have verified signatures." 

Can you fix please?

dmitrykiselev27:
@edersonbrilhante fixed

edersonbrilhante:
I think is good. But I don't have enterprise env to test. Have you tested after fixing the conflicts?

dmitrykiselev27:
Yes, I've deployed my branch to an enterprise I have access to. I'm happy to provide test evidences in case you want to check something in particular 
image

edersonbrilhante:
You moved the policy from json to data. This is a not problem for me, but it created a mixed approach when compared with other modules. I personally prefer data, but I think this could be a following PR.

edersonbrilhante:
Could we move this nonsensitive() change to a separate PR? It appears unrelated to enterprise runner support and adds noise to an already large change. If it is required for this feature, please document the specific Terraform error or sensitive-value propagation it fixes.

edersonbrilhante:
@dmitrykiselev27  Can you fix the conflicts? Heads up, I am working to refactor the module to introduce plugin layout. Ref: #5234

dmitrykiselev27:
@edersonbrilhante both changes (policy and nonsensitive) were attempts to fix this error, which I'm not longer can reproduce in my test environment,  so I've reverted them.
```
  │  122:   policy = templatefile("${path.module}/policies/lambda-scale-up.json", {
  │  123:     arn_runner_instance_role  = aws_iam_role.runner.arn
  │  124:     sqs_arn                   = var.sqs_build_queue.arn
  │  125:     github_app_id_arn         = nonsensitive(var.github_app_parameters.id.arn)
  │  126:     github_app_key_base64_arn = nonsensitive(var.github_app_parameters.key_base64.arn)
  │  127:     ssm_config_path           = local.arn_ssm_parameters_path_config
  │  128:     kms_key_arn               = local.kms_key_arn
  │  129:     ami_kms_key_arn           = local.ami_kms_key_arn
  │  130:     ssm_ami_id_parameter_arn  = local.ami_id_ssm_module_managed ? local.arn_ssm_parameter_ami_id : var.ami.id_ssm_parameter_arn
  │  131:     enterprise_pat_arn        = var.enterprise_pat_parameter != null ? nonsensitive(var.enterprise_pat_parameter.arn) : ""
  │  132:   })
  │ 
  │ Call to function "templatefile" failed: function returned an inconsistent
  │ result.
```

lambdas/functions/control-plane/src/scale-runners/runner-config.ts
```diff
@@ -0,0 +1,26 @@
...
+      case 'repo':
+        return 'Repo';
+      default:
+        throw new Error(
+          `Invalid RUNNER_REGISTRATION_LEVEL: '${registrationLevel}'. Must be 'enterprise', 'org', or 'repo'.`,
+        );
+    }
+  }
+  return 'Repo';
+}
```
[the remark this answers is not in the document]
dmitrykiselev27: Backward compatibility implemented on the Terraform level

[runners module](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5096/changes#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR17)

[multi-runner](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5096/changes#diff-57f00cdf57feef92ffcb35d6618e62e14ad652d5d520f331068332d5f3cade51R40)

edersonbrilhante:
@dmitrykiselev27 Can you fix the conflicts?

dmitrykiselev27:
@edersonbrilhante fixed conflicts

edersonbrilhante:
@dmitrykiselev27 I am about to finish the work to add support for a version 2 of mudule multi-runner. This v2 will decouple ec2, ssm and control plane(webhook, sqs, lambda) from multi-runner.

It will allow to support microvm in addition to ec2, dynamodb as alternative to ssm, and scaleset in addtion to current control plane.

Once it is done, we can revisit it. 

Reference:
https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Aopen+is%3Apr+author%3Aedersonbrilhante+label%3Aexperimental-v2

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Aorchestration-provider%3Ascale-set

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Astorage-provider%3Aaws_dynamodb

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Acompute-provider%3Aaws_microvm

PR #5099, opened 2026-04-09
feat(webhook): support multi-region SQS dispatch in runner webhook

bogdankrasko:
## Description

This PR updates the webhook SQS dispatch logic to select the SQS client region from the target queue URL instead of always using `AWS_REGION`.

This fixes cross-region dispatch scenarios where the webhook needs to send workflow job messages to an SQS queue in a different AWS region than the Lambda runtime. The change also caches SQS clients per region so repeated sends reuse the correct client while still creating separate clients when queues are in different regions.

To support this model, the webhook `runner_matcher_config` must include each runner group's `matcherConfig` together with the region-specific SQS queue URL (`id`) and queue ARN (`arn`) for every runner group in every region. This allows the webhook to match the job and dispatch it to the correct regional queue.

This also requires runner pool labels to be unique across all configured regions so webhook matching remains deterministic and a given job maps to exactly one regional queue.

## Test Plan

- Added and updated unit tests in `lambdas/functions/webhook/src/sqs/index.test.ts` to cover:
  - queue URL region parsing
  - fallback to `AWS_REGION` when the queue URL cannot be parsed
  - behavior when no region can be resolved
  - client reuse for multiple queues in the same region
  - separate clients for different queue regions
- Ran `yarn vitest run functions/webhook/src/sqs/index.test.ts`
- Deployed the webhook stack only in the primary region, `us-east-2`, so a single API Gateway/webhook entrypoint handles workflow dispatch for runner queues across multiple AWS regions
- Deployed validation stacks across multiple AWS regions using `modules/runners`. This setup provisions per-runner-group Lambda functions (scale-up, scale-down, pool, and job-retry), along with the SQS queues defined in [modules/multi-runner/queues.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/multi-runner/queues.tf). This was used to validate both same-region and cross-region dispatch behavior.
To enable it, an additional wrapper was introduced within `modules/runners` to incorporate the required SQS queues.
  - `us-east-2`: Ubuntu 24.04 runner pools for both `amd64` and `arm64`. This region was used to validate that jobs with architecture-specific labels were matched correctly and dispatched successfully to queues in the local region.
  - `us-west-2`: Ubuntu 24.04 `amd64-2` runner pool. This region was used as the secondary-region target to validate that the webhook could dispatch a workflow job to an SQS queue outside the Lambda runtime region.
- Triggered workflow job dispatch in the deployed environment and verified that the webhook sent messages successfully to the expected queue in each case, including the cross-region queue path this change is intended to support


edersonbrilhante (changes requested):
@bogdankrasko Can you please sign the commits?

bogdankrasko:
@edersonbrilhante Sure, signed the commits. Let me know if you have any questions regarding the changes.

edersonbrilhante (dismissed):

edersonbrilhante:
@bogdankrasko Can you this https://github.com/github-aws-runners/terraform-aws-github-runner/actions/runs/28951982062/job/86459632957?pr=5099?

bogdankrasko:
@edersonbrilhante Fixed the linter issues

edersonbrilhante (approved):

PR #5109, opened 2026-04-16
fix: retry all SQS messages on unhandled errors instead of silently dropping them

Snojo:
## Description

When the scale-up lambda encounters an unhandled error (e.g. SSM ThrottlingException during registration token creation), the catch block returned an empty batchItemFailures array. With ReportBatchItemFailures enabled, this tells SQS that all messages were processed successfully, permanently deleting them from the queue.

This causes queued GitHub Actions jobs to be silently lost — they never get a runner and remain stuck in 'queued' state indefinitely.

The fix returns all message IDs as batch item failures on unhandled errors, so SQS retries them after the visibility timeout.

## Related Issues

We ran into issues with our runners in https://github.com/island-is/island.is

### DIFF
--- a/lambdas/functions/control-plane/src/lambda.test.ts
+++ b/lambdas/functions/control-plane/src/lambda.test.ts
@@ -215,7 +215,9 @@ describe('Test scale up lambda wrapper.', () => {
       vi.mocked(scaleUp).mockRejectedValue(new Error('Generic error'));
 
       const result = await scaleUpHandler(multiRecordEvent, context);
-      expect(result).toEqual({ batchItemFailures: [] });
+      expect(result).toEqual({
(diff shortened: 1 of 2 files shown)


edersonbrilhante:
@Snojo AWS exceptions can be handled through scale errors. It makes sense to retry transient errors, but there are some cases where I don’t want retries, for example when the AMI doesn’t exist.

edersonbrilhante:
I created this PR to fix it #5220

edersonbrilhante:
I implemented this idea in #5220

edersonbrilhante:
@Snojo Can we close?

Snojo:
Closed :)

PR #5123, opened 2026-05-13
fix: retry installation auth for regional GitHub App events

bogdankrasko:
## Description

Handle webhook events whose installationId belongs to a different GitHub App by retrying installation authentication after a 404.

Apply the fallback in both the shared Octokit helper and the scale-up flow by re-resolving the installation ID for the current app before retrying auth.

Preserve existing behavior for installationId=0 and non-404 errors, and add regression tests covering the mismatched installation/app scenario.

## Test Plan

Follow the instructions in https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5099 to deploy runners to multiple regions, and set the GitHub App secrets in SSM so that each region has its own GitHub App ID.


jtmthf:
Bumping this for review. I observed an incident yesterday where we received an influx of events with an incorrect organization id that impacted our ability to scale up.

edersonbrilhante (commented):
@bogdankrasko Can you sign your commits?

bogdankrasko:
@edersonbrilhante Sure, signed the commits. Let me know if you have any questions regarding the changes.

edersonbrilhante:
Can you fix the formatting issues?

bogdankrasko:
@edersonbrilhante Fixed the formatting issues

edersonbrilhante (approved):

Issue #5128, opened 2026-05-26
Webhook dispatch uses unstable sort when multiple pools match the same labels


vegardx:
## Description

When `exactMatch: true` is configured (subset-check: all job labels must exist in the runner pool's label set), a workflow job requesting a subset of labels can match multiple runner pools that all contain those labels as a subset of their own.

The current dispatch logic iterates `matcherConfig` without a stable sort, so which pool receives the SQS message is non-deterministic. In practice, jobs can land on the most expensive pool when a cheaper one would have sufficed.

### Pool configuration

Each pool has its own unique labels, but they all share a common subset:

| Pool | Labels | Instance |
|------|--------|----------|
| small-spot | `[self-hosted, linux, x64, small, spot]` | 2 vCPU c5.large spot |
| medium-spot | `[self-hosted, linux, x64, medium, spot]` | 4 vCPU c5.xlarge spot |
| large-ondemand | `[self-hosted, linux, x64, large, ondemand]` | 16 vCPU c5.4xlarge on-demand |

### Workflow A — requests specific pool (works today)

```yaml
runs-on: [self-hosted, linux, x64, small, spot]
```

Only matches `small-spot` because `exactMatch: true` requires all job labels to be present in the pool's label set. ✅ Deterministic.

### Workflow B — requests just the common subset (broken today)

```yaml
runs-on: [self-hosted, linux, x64]
```

All three pools contain `[self-hosted, linux, x64]` in their label set, so all three match. Which pool gets the message depends on iteration order of `matcherConfig` — **non-deterministic**. The job might land on `large-ondemand` (16 vCPU, on-demand pricing) when `small-spot` would have been sufficient.

This pattern is common: teams define a "baseline" `runs-on` for simple CI jobs and only add specific labels (e.g. `large`) when they actually need more resources.

## Key assumption

The omission of labels is intentional — the requester doesn't care which specific pool handles the job, they just want any runner that matches their request. A workflow using `runs-on: [self-hosted, linux, x64]` is saying "I need a Linux x64 runner" without expressing a preference for size or pricing. The `priority` field lets the **operator** decide which pool should be preferred for these ambiguous requests (e.g. cheapest first).

## Why this matters

- **Cost:** Jobs silently route to expensive pools
- **Capacity:** Cheap spot pools sit idle while expensive on-demand pools scale up unnecessarily
- **Predictability:** Same workflow can use different instance types across runs depending on config iteration order

## Suggested fix

[...]

PR #5031, opened 2026-02-13
feat: add bidirectionalLabelMatch option and deprecate exactMatch

Brend-Smits:
## Description

Introduce a new `bidirectionalLabelMatch` option that performs strict two-way label matching (runner labels must equal workflow labels as a set). This preserves the existing `exactMatch` behavior (unidirectional subset check) to avoid breaking changes.

The `bidirectionalLabelMatch` option requires labels to be identical in both directions. Previously, a runner with labels `[A, B, C, D]` would match a job requesting `[A, B, C]` when `exactMatch` was `true`. With `bidirectionalLabelMatch=true`, the runner will only match if the job requests exactly `[A, B, C, D]`.

```
Before: Job [A,B,C] + Runner [A,B,C,D] + exactMatch=true            → Match
After:  Job [A,B,C] + Runner [A,B,C,D] + bidirectionalLabelMatch=true → No Match
```

`exactMatch` was supposed to have this behavior, but to avoid breaking changes, the variable will be deprecated to give users time to migrate.

**Changes:**
- Add `bidirectionalLabelMatch` to `MatcherConfig` interface (optional, defaults to `false`)
- Update `canRunJob` to support bidirectional matching when enabled — now accepts both `bidirectionalLabelMatch` and `enableDynamicLabels` parameters; bidirectional matching correctly respects dynamic label sanitization
- Deprecate `exactMatch` in Terraform variables with migration guidance
- Add `bidirectionalLabelMatch` to multi-runner and webhook variable types
- Add new root variable `enable_runner_bidirectional_label_match`
- Add comprehensive test coverage for bidirectional matching

**Migration:** To migrate, use `bidirectionalLabelMatch` instead of `exactMatch` in your runner configs. Then either:
1. Remove extra labels from runner configurations, or
2. Add the extra labels to your workflow job `runs-on`

Co-authored-by: Stuart Pearson

## Test Plan

- Added/updated unit tests in `dispatch.test.ts` covering bidirectional label matching, interaction with `exactMatch`, case-insensitivity, empty labels, and combined use with `enableDynamicLabels` — 36 tests passing

## Related Issues

PR #5133, opened 2026-05-26
feat(webhook): use priority field as tiebreaker in matcher dispatch

vegardx:
## Problem

When multiple runner pools match the same job labels via `exactMatch: true` (subset matching), dispatch order is non-deterministic. A workflow requesting `[self-hosted, linux, x64]` can land on any pool whose labels contain that subset — including an expensive 16 vCPU on-demand pool when a cheap 2 vCPU spot pool would suffice.

See #5128 for the full discussion and concrete examples.

## Fix

The Terraform variable `runner_matcher_config` already supports an optional `priority` field (default 999) and pre-sorts the config before writing to SSM. However the Lambda re-sorts only by `exactMatch`, discarding the priority ordering. This PR adds `priority` as a secondary sort key:

```typescript
matcherConfig.sort((a, b) => {
  if (a.matcherConfig.exactMatch !== b.matcherConfig.exactMatch) {
    return a.matcherConfig.exactMatch ? -1 : 1;
  }
  return (a.matcherConfig.priority ?? 999) - (b.matcherConfig.priority ?? 999);
});
```

The assumption is that when a workflow omits specific labels (e.g. uses just `[self-hosted, linux, x64]` instead of `[self-hosted, linux, x64, large, ondemand]`), the omission is intentional — the requester does not care which specific pool handles the job. The `priority` field lets the operator decide which pool should be preferred for these ambiguous requests.

## Changes

- `lambdas/functions/webhook/src/sqs/index.ts` — add optional `priority` field to `MatcherConfig`
- `lambdas/functions/webhook/src/runners/dispatch.ts` — sort by priority as tiebreaker
- `lambdas/functions/webhook/src/runners/dispatch.test.ts` — test: multiple pools match common subset, lowest priority wins

## Backwards compatible

- `priority` is optional, defaults to 999
- Existing configs without priority behave identically (all sort equal, first match wins as before)
- The Terraform variable already defines `priority = optional(number, 999)` — no Terraform changes needed

Refs: #5128

### DIFF
--- a/lambdas/functions/webhook/src/runners/dispatch.test.ts
+++ b/lambdas/functions/webhook/src/runners/dispatch.test.ts
@@ -153,6 +153,54 @@ describe('Dispatcher', () => {
       });
     });
 
+    it('should dispatch to lowest priority pool when multiple pools match the same labels', async () => {
+      config = await createConfig(undefined, [
+        {
+          ...runnerConfig[0],
+          id: 'large-ondemand',
+          matcherConfig: {
+            labelMatchers: [['self-hosted', 'linux', 'x64', 'large', 'ondemand']],
+            exactMatch: true,
+            priority: 100,
+          },
+        },
+        {
+          ...runnerConfig[0],
+          id: 'medium-spot',
+          matcherConfig: {
+            labelMatchers: [['self-hosted', 'linux', 'x64', 'medium', 'spot']],
+            exactMatch: true,
+            priority: 5,
+          },
+        },
+        {
+          ...runnerConfig[0],
+          id: 'small-spot',
+          matcherConfig: {
+            labelMatchers: [['self-hosted', 'linux', 'x64', 'small', 'spot']],
+            exactMatch: true,
+            priority: 1,
+          },
+        },
+      ]);
+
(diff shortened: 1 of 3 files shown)


vegardx:
Closing this one — I was wrong about the premise, and main has moved past it.

`priority` already works end-to-end: `modules/webhook/webhook.tf` builds the matcher keys as `format("%03d-%s", priority, k)` and sorts them before serialising to SSM, so the array reaching the lambda is *already* in priority order. Re-sorting in the lambda just redoes Terraform's work. And `Array.sort` has been stable by spec since ES2019, so the tiebreaker this PR adds isn't fixing real nondeterminism.

Worse, the comparator here predates `bidirectionalLabelMatch` (#5031) — merging it as-is would regress bidirectional matching back to `exactMatch` only.

What was worth salvaging is in #5214: declaring `priority` on the `MatcherConfig` type, and not sorting the shared config in place — `ConfigLoader` holds it on a static singleton, so the in-place sort lets one event reorder what the next one sees on a warm container. Idempotent today, but it makes dispatch order depend on invocation history.

Issue #5037, opened 2026-02-18
Support for multiple GitHub Apps to overcome API rate limits at scale


thomasnemer:
### Problem

When running this module at enterprise scale with many runner configurations, GitHub API rate limits become a bottleneck that cannot be solved by tuning the existing setup.

#### How API usage grows with runner configurations

Each runner configuration (in multi-runner mode) gets its own set of Lambda functions (scale-up, scale-down, pool). Each of these makes independent GitHub API calls through a single shared GitHub App:

**Scale-up** (per batch of workflow job events):
- 1 installation token creation per unique owner/repo group
- 1 `actions.getJobForWorkflowRun()` per message (job queued check)
- 1 `actions.generateRunnerJitconfigForOrg/Repo()` per runner created (ephemeral/JIT runners)
- Optionally 1 paginated `GET /orgs/{org}/actions/runner-groups` for runner group lookup

**Scale-down** (per cycle):
- 1 paginated `actions.listSelfHostedRunnersForOrg/Repo()` per owner
- 1 `actions.getSelfHostedRunnerForOrg/Repo()` per runner evaluated
- 1 `actions.deleteSelfHostedRunnerFromOrg/Repo()` per runner removed

**Pool** (per adjustment cycle):
- 1 `apps.getOrgInstallation()` + 1 paginated `actions.listSelfHostedRunnersForOrg()` per pool
- 1 registration token or JIT config call per runner created

All of these calls go through a single GitHub App installation token, sharing the same rate limit bucket.

#### Concrete example

With 10 runner configurations, a moderate burst of 50 workflow jobs, and 100 active runners:

| Component | API calls per cycle | Notes |
|---|---|---|
| Scale-up (10 configs × ~5 jobs each) | ~110 | Auth setup + JIT config per runner |
| Scale-down (10 configs × ~10 runners each) | ~180 | List + status check + deletions |
| Pool (if enabled, 10 configs) | ~60 | List + top-up |
| **Total per cycle** | **~350** | |

With scale-down running every 5 minutes and scale-up triggered on demand, this can easily consume **4,000–5,000+ calls/hour** under sustained load — approaching or exceeding the rate limit.

#### GitHub API rate limits

- **Standard GitHub App**: 5,000 req/hour (scales up to 12,500 max with >20 repos/users)
- **GitHub Enterprise Cloud**: 15,000 req/hour (fixed)

[...]

PR #5136, opened 2026-05-27
fix: enable Octokit throttle retry and SSM adaptive retry under burst

### LINKED ISSUE #5135: fix: Octokit throttle callbacks should retry, and SSM client should use adaptive retry
## Problem

Two resilience gaps that compound under burst load:

### 1. Octokit throttle plugin never retries

The `@octokit/plugin-throttling` callbacks `onRateLimit` and `onSecondaryRateLimit` in `auth.ts` only log a warning — they don't return `true`, so the plugin never retries the request. The plugin's contract is:

- Return `true` → retry after the `retryAfter` delay
- Return nothing / `false` → throw immediately

Current code:
```typescript
onRateLimit: (retryAfter, options) => {
  logger.warn(`GitHub rate limit: Request quota exhausted...`);
  // implicitly returns undefined → plugin throws, request fails
},
```

This means any GitHub API call that hits a rate limit immediately fails and propagates up — potentially causing the entire SQS batch to be dropped (the problem fixed by #5129).

### 2. SSM client uses default retry (standard, 3 attempts)

Under burst (multiple concurrent Lambdas writing JIT configs via `PutParameter`), SSM's per-account rate limit (~40 TPS for PutParameter standard throughput) is easily exceeded. The default SDK retry:

- `retryMode: 'standard'` — exponential backoff without rate-sensing
- `maxAttempts: 3` — gives ~3 seconds of retry budget

This is insufficient. After 3 attempts the SDK throws `ThrottlingException`, which propagates up and (pre-#5129) silently drops the entire SQS batch.

### 1. Return `true` from throttle callbacks (with retry caps)

```typescript
onRateLimit: (retryAfter, options) => {
  logger.warn(`...retrying after ${retryAfter}s`);
  return options.request.retryCount < 2;
},
onSecondaryRateLimit: (retryAfter, options) => {
  logger.warn(`...retrying after ${retryAfter}s`);
  return options.request.retryCount < 1;
},
```

Primary rate limit: retry up to 2 times. Secondary (abuse) rate limit: retry once. These are conservative caps — the plugin handles the `retryAfter` delay automatically.

### 2. Adaptive retry with maxAttempts=10 for SSM

```typescript
const SSM_CLIENT_CONFIG = {
  region: process.env.AWS_REGION,
  maxAttempts: 10,
  retryMode: 'adaptive' as const,
};
```

`adaptive` mode adds client-side rate-sensing via a token bucket — when the SDK sees ThrottlingException it slows further calls to match the observed budget. Combined with 10 attempts this gives ~30s of retry per call without hammering the API.

This is safe because runners take ~30-50s to boot before reading their JIT config from SSM, so longer retry on PutParameter is essentially free.

## Impact

| Scenario | Before | After |
|---|---|---|
| GitHub rate limit during scale-up | Request fails immediately, batch potentially dropped | Retries 1-2 times with backoff, succeeds if rate limit window passes |
| SSM throttle during JIT config write | ThrottlingException after ~3s, batch fails | Adaptive backoff for ~30s, succeeds once throttle clears |
| Burst of 100 jobs with batch_size=10 | High probability of SSM throttle → orphaned instances | Retry absorbs the throttle, instances get their configs |

Refs: #5024, #5037

vegardx:
## Problem

Two resilience gaps that compound under burst load:

1. **Octokit throttle plugin never retries** — `onRateLimit` and `onSecondaryRateLimit` callbacks log a warning but don't return `true`, so `@octokit/plugin-throttling` throws immediately instead of retrying after the `retryAfter` delay.

2. **SSM client uses default retry** (`standard`, 3 attempts, ~3s budget) — under burst with multiple concurrent Lambdas writing JIT configs via PutParameter, SSM's ~40 TPS limit is easily exceeded and `ThrottlingException` propagates up, failing the entire batch.

### Octokit throttle callbacks return `true` with caps

```typescript
onRateLimit: (retryAfter, options) => {
  logger.warn(`...retrying after ${retryAfter}s`);
  return options.request.retryCount < 2; // retry up to 2x on primary rate limit
},
onSecondaryRateLimit: (retryAfter, options) => {
  logger.warn(`...retrying after ${retryAfter}s`);
  return options.request.retryCount < 1; // retry once on secondary/abuse limit
},
```

### SSM adaptive retry with maxAttempts=10

```typescript
const SSM_CLIENT_CONFIG = {
  region: process.env.AWS_REGION,
  maxAttempts: 10,
  retryMode: 'adaptive' as const,
};
```

`adaptive` mode adds client-side rate-sensing — when the SDK sees ThrottlingException it slows further calls to match the observed budget. 10 attempts gives ~30s of retry budget, which is safe because runners take ~30-50s to boot before reading their JIT config.

## Changes

- `lambdas/functions/control-plane/src/github/auth.ts` — throttle callbacks return `true` with retry caps
- `lambdas/libs/aws-ssm-util/src/index.ts` — shared SSM client config with adaptive retry

## Impact

| Scenario | Before | After |
|---|---|---|
| GitHub rate limit during scale-up | Request fails immediately | Retries 1-2x with backoff |
| SSM throttle during JIT config write | ThrottlingException after ~3s | Adaptive backoff for ~30s |
| Burst of 100 jobs, batch_size=10 | SSM throttle → orphaned instances | Retry absorbs throttle |

Fixes #5135
Refs: #5024, #5037

### DIFF
--- a/lambdas/functions/control-plane/src/github/auth.test.ts
+++ b/lambdas/functions/control-plane/src/github/auth.test.ts
@@ -6,7 +6,7 @@ import { getParameters } from '@aws-github-runner/aws-ssm-util';
 import { generateKeyPairSync } from 'node:crypto';
 import * as nock from 'nock';
 
-import { createGithubAppAuth, createOctokitClient } from './auth';
+import { createGithubAppAuth, createOctokitClient, onRateLimit, onSecondaryRateLimit } from './auth';
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 
 type MockProxy<T> = T & {
@@ -295,5 +295,29 @@ describe('Test createGithubAppAuth', () => {
     expect(callArgs.request).toBeDefined();
     expect(mockedAuth).toBeCalledWith({ type: authType });
     expect(result.token).toBe(token);
+  });
+});
+
+describe('Test throttling retry caps', () => {
+  // The plugin passes retryCount as the 4th argument and uses the boolean return
+  // value to decide whether to retry (see @octokit/plugin-throttling wrap-request).
+  const options = { method: 'GET', url: '/repos/o/r' } as never;
+  const octokit = {} as never;
+
+  it.each([
+    [0, true],
+    [1, true],
+    [2, false],
+    [3, false],
+  ])('onRateLimit retries at retryCount=%i -> %s', (retryCount, expected) => {
+    expect(onRateLimit(60, options, octokit, retryCount)).toBe(expected);
+  });
+
+  it.each([
+    [0, true],
+    [1, false],
+    [2, false],
+  ])('onSecondaryRateLimit retries at retryCount=%i -> %s', (retryCount, expected) => {
+    expect(onSecondaryRateLimit(60, options, octokit, retryCount)).toBe(expected);
   });
 });
--- a/lambdas/functions/control-plane/src/github/auth.ts
+++ b/lambdas/functions/control-plane/src/github/auth.ts
@@ -54,11 +54,15 @@ export async function createOctokitClient(token: string, ghesApiUrl = ''): Promi
     throttle: {
       onRateLimit: (retryAfter: number, options: Required<EndpointDefaults>) => {
         logger.warn(
-          `GitHub rate limit: Request quota exhausted for request ${options.method} ${options.url}. Requested `,
+          `GitHub rate limit: Request quota exhausted for request ${options.method} ${options.url}, retrying after ${retryAfter}s`,
         );
+        return (options as unknown as { request: { retryCount: number } }).request.retryCount < 2;
       },
       onSecondaryRateLimit: (retryAfter: number, options: Required<EndpointDefaults>) => {
-        logger.warn(`GitHub rate limit: SecondaryRateLimit detected for request ${options.method} ${options.url}`);
+        logger.warn(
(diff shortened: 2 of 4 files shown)


vegardx:
Pushed a follow-up addressing all of these — the third one was a real bug, thanks Copilot:

- **Casts**: unnecessary. The throttling plugin already passes `retryCount` as the 4th argument to both handlers, so I use that instead of reaching through `options.request`.
- **SSMClient reuse**: correct and it mattered. Adaptive retry keeps its rate-sensing token bucket on the client instance, so constructing a fresh client per call discarded it every time — `retryMode: 'adaptive'` was doing almost nothing. Now memoised per container.
- **Tests**: added retry-cap boundary tests for both handlers, plus coverage of the client's retry config and reuse.

Would appreciate a human pass on the retry caps themselves (2 for primary, 1 for secondary) — those are judgement calls, not derived from anything.

edersonbrilhante:
@vegardx Build is not working

vegardx:
Fixed in f3f65e99: the dist build runs tsc and caught a type error the tests missed, the throttle handlers typed their octokit param as @octokit/rest's Octokit but the plugin's LimitHandler expects @octokit/core's, so I typed it to match. Build and tests pass locally.

edersonbrilhante (approved):

PR #5139, opened 2026-05-28
fix(runner): handle long prefixes in pool scheduler and IAM resource names

alexjurkiewicz:
The pool module's `aws_scheduler_schedule_group` and `aws_iam_role` for the scheduler both use `name_prefix`, which has a 38-character limit (64 max name minus 26 for Terraform's random suffix). When the module prefix is long — for example a multi-runner prefix like `self-hosted-linux-ubuntu2404-x64-2cpu-8gb` — the combined `${prefix}-pool` exceeds 38 characters and Terraform fails.

For long prefixes, the name is now truncated with an md5 hash suffix to stay within limits, matching the existing pattern used elsewhere in the module. Existing deployments with short prefixes see no resource changes.

### DIFF
--- a/modules/runners/pool/main.tf
+++ b/modules/runners/pool/main.tf
@@ -1,3 +1,11 @@
+locals {
+  pool_name_prefix = (
+    length("${var.config.prefix}-pool") <= 38
+    ? "${var.config.prefix}-pool"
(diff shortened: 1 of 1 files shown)


edersonbrilhante (changes requested):
@alexjurkiewicz Please sign the commits

alexjurkiewicz:
Done, and rebased too

edersonbrilhante (approved):

PR #5172, opened 2026-06-12
feat(webhook): enforce per-runner dynamic labels policy in dispatcher

### LINKED ISSUE #5160: Feature: allow enable_dynamic_labels per multi_runner_config entry
### Problem

`enable_dynamic_labels` is currently a global toggle on the multi-runner module. When enabled, it applies to all runner configurations managed by that deployment. There is no way to enable dynamic labels for a specific runner group (e.g., a staging or experimental config) without enabling it for all groups — or deploying an entirely separate stack with its own webhook and GitHub App.

### Proposed Solution

Allow `enable_dynamic_labels` to be specified per `multi_runner_config` entry, so operators can selectively enable dynamic label parsing for specific runner groups without affecting others.

For example:

```hcl
multi_runner_config = {
  "experimental" = {
    matcherConfig = { ... }
    runner_config = {
      enable_dynamic_labels = true
      ...
    }
  }
  "stable" = {
    matcherConfig = { ... }
    runner_config = {
      enable_dynamic_labels = false
      ...
    }
  }
}
```

### Use Case

We operate multiple runner groups in a single deployment and want to trial dynamic labels on a specific group (e.g., staging runners) before rolling it out broadly — without spinning up a full separate deployment with its own webhook endpoint and GitHub App configuration.

### Current Workaround

Deploy a second multi-runner stack with `enable_dynamic_labels = true` and separate label matchers. This works but adds operational overhead (duplicate webhook, separate App config, additional infrastructure).

### LINKED ISSUE #5161: Feature: allow/deny list for dynamic label values
### Problem

When `enable_dynamic_labels` is enabled, there is no mechanism to restrict the EC2 configuration values that workflow authors can request. The webhook layer performs format-level sanitization (character allowlist and max length), and the scale-up layer does structural key parsing — but **no semantic or business-level validation** is applied to the actual values before they reach the EC2 CreateFleet API.

This means any workflow author can specify:
- Expensive instance types (`ghr-ec2-instance-type:p5.48xlarge`)
- Arbitrary AMI IDs (`ghr-ec2-image-id:ami-compromised`)
- Oversized EBS volumes (`ghr-ec2-ebs-volume-size:10000`)
- Any subnet or availability zone

The documentation correctly flags this as a security concern, but the only mitigation today is "trust all workflow contributors" — which doesn't scale in organizations with many teams.

### Proposed Solution

Add a configurable policy that provides **two levels of control**:

#### 1. Key-level allow/deny (which override categories are permitted)

Operators should be able to control which `ghr-ec2-*` keys users are allowed to use at all. This enables scenarios like:

- "Users can ONLY switch AMI, nothing else"
- "Users cannot set EBS volume size, instance type, or AMI"
- "Users can only use instance-type and vcpu/memory requirements"

```hcl
dynamic_labels_policy = {
  # Allowlist approach: only these keys are accepted, everything else is denied
  allowed_keys = ["image-id"]

  # OR denylist approach: these keys are rejected, everything else is accepted
  # denied_keys = ["ebs-volume-size", "instance-type", "image-id", "subnet-id"]
}
```

Labels using a denied (or non-allowed) key would be stripped with a warning, and the job would either fall back to the base runner config or be rejected entirely (configurable).

#### 2. Value-level allow/deny (what values are accepted for permitted keys)

For keys that ARE permitted, operators should be able to further restrict the values:

```hcl
dynamic_labels_policy = {
  allowed_keys = ["instance-type", "image-id", "ebs-volume-size"]

  instance_type = {
    allowed = ["m5.large", "m5.xlarge", "c5.*", "m6a.*"]
    # or: denied = ["p5.*", "x2idn.*"]
  }
  image_id = {
    allowed = ["ami-abc123", "ami-def456"]
  }
  ebs_volume_size = {
    max = 200
  }
  max_price = {
    max = "0.50"
  }
}
```

### Example Scenarios

| Scenario | Configuration |
|----------|--------------|
| Users can only switch AMI | `allowed_keys = ["image-id"]` + optional value allowlist |
| Users can pick instance type from a safe set | `allowed_keys = ["instance-type"]` + `instance_type.allowed = ["m5.*", "c5.*"]` |
| Users can do anything except change networking/AMI | `denied_keys = ["subnet-id", "availability-zone", "image-id"]` |
| Full flexibility (current behavior) | No policy configured |

### Use Cases

1. **Cost control** — Restrict instance types to a known-safe set, cap EBS volume sizes and spot prices
2. **Security boundaries** — Prevent AMI overrides or subnet targeting in organizations that can't fully trust all workflow contributors
3. **Gradual rollout** — Enable dynamic labels broadly but constrain what can actually be overridden, reducing risk without requiring per-contributor trust
4. **Least privilege** — Only expose the specific override capabilities that teams actually need

### Current Workaround

The only option is to not enable the feature, or to accept the full attack surface and rely on GitHub branch protection rules for workflow file changes. Neither is a practical middle ground for larger organizations.

edersonbrilhante:
## Description

Adds an opt-in, per-matcher policy for dynamic EC2 override labels so operators
can declare which `ghr-ec2-*` keys/values each runner queue is allowed to honor.

The webhook lambda is the gatekeeper: when a `workflow_job` arrives, it picks the
first matching runner queue that both opts into dynamic labels
(`matcherConfig.enableDynamicLabels`) and accepts every `ghr-ec2-*` label on the
job per `matcherConfig.ec2DynamicLabelsPolicy`. If no queue qualifies, the
request is returned with `202` and a warning is logged instead of falling back to
a queue with the dynamic labels stripped.

Highlights:

- Adds a small EC2 dynamic labels policy evaluator co-located with `dispatch.ts`.
- Replaces the flat reserved-key schema with:
  `{ blocked_keys?, restricted_keys? }`.
- `blocked_keys` rejects EC2 override keys outright.
- `restricted_keys` applies per-key value rules:
  `{ allowed?, denied?, max? }`.
- EC2 override keys not listed in either `blocked_keys` or `restricted_keys` are
  allowed by default.
- The flag and policy travel inside the existing `runner-matcher-config` SSM
  parameter via `MatcherConfig`. No new SSM parameter or env var is added.
- Root single-runner config exposes `ec2_dynamic_labels_policy`; matcher config
  exposes `ec2DynamicLabelsPolicy`.

### Example

```hcl
multi_runner_config = {
  m5 = {
    matcherConfig = {
      labelMatchers           = [["self-hosted", "linux"]]
      exactMatch              = true
      enableDynamicLabels     = true
      ec2DynamicLabelsPolicy = {
        blocked_keys = ["image-id", "subnet-id"]

        restricted_keys = {
          "instance-type" = {
            allowed = ["m5.*"]
            denied  = ["m5.metal"]
          }

          "ebs-volume-size" = {
            max = 200
          }
        }
      }
    }

    runner_config = { ... }
  }
}
```

A job with `ghr-ec2-instance-type:r5.large` will be rejected with `202` and a
warning for the `m5` queue rather than being silently dispatched without the
label.

## Test Plan

- `yarn exec vitest run functions/webhook/src/runners/dynamic-labels-policy.test.ts functions/webhook/src/runners/dispatch.test.ts`
- `terraform fmt -check -diff main.tf variables.tf modules/webhook/variables.tf modules/multi-runner/variables.tf`

## Related Issues

Fixes #5161
Fixes #5160


Brend-Smits:
@copilot resolve the merge conflicts in this pull request

edersonbrilhante:
@Brend-Smits I tested in my environment and it is working

docs/configuration.md
```diff
@@ -368,10 +367,32 @@ module "runners" {
...
+      "ebs-volume-size" = {
+        max = 200
+      }
+    }
+  }
   ...
 }
 ```
 
+The policy is evaluated by dynamic label key:
```
Brend-Smits, line 388: We should definitely also add an `allowed_keys` instead of just having a `blocked_keys`. But let's do this in a new issue/pr!

Brend-Smits (approved):
Great work! Thanks a lot 🚀

docs/configuration.md
edersonbrilhante: I used this logic:
 
Blocked keys = never allow 
Restricted Keys = allow to use, but with restricted values only
Not blocked or restricted -> allowed to use

PR #5185, opened 2026-06-25
feat: tag runners with github labels

edersonbrilhante:
## Description

Adds EC2 tags for the GitHub runner labels registered through the JIT runner flow.

Now that dynamic labels are supported, runners created from the same multi-runner entry can register with different label sets. Before this change, AWS-side inspection only had the instance/runner identity, so checking the exact labels required going to GitHub.

The labels are taken from `runnerConfig.data.runner.labels`, so the EC2 tags reflect the labels GitHub registered for that runner. The instance create-tags policy was also updated to allow the new label tag keys.

## Test Plan

Deploy the lambda, run workflow and check the new labels

## Related Issues

None

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
@@ -50,8 +50,8 @@ const logger = createChildLogger('scale-up');
 const RUNNER_LABELS_TAG_KEY = 'ghr:runner_labels';
 const RUNNER_LABELS_TAG_VALUE_SEPARATOR = ',';
 const EC2_OVERRIDE_LIST_VALUE_SEPARATOR = ';';
-const EC2_TAG_VALUE_MAX_LENGTH = 256;
-const RUNNER_LABELS_TAG_MAX_COUNT = 5;
+export const EC2_TAG_VALUE_MAX_LENGTH = 256;
+export const RUNNER_LABELS_TAG_MAX_COUNT = 5;
 
 export type LambdaRunnerSource = 'scale-up-lambda' | 'pool-lambda';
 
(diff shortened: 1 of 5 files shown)


lambdas/functions/control-plane/src/scale-runners/scale-up.ts
```diff
@@ -704,11 +768,16 @@ async function createRegistrationTokenConfig(
   return [];
 }
 
-async function tagRunnerId(instanceId: string, runnerId: string): Promise<void> {
+async function tagRunnerMetadata(instanceId: string, runnerId: string, runnerLabels: unknown): Promise<void> {
```
Brend-Smits: We need to make sure that we sanitize the tags now that we are dynamically inputting the values here.
Especially in terms of special characters and the size. There's some ec2 tag limits here that we have to take care of. 
Can you make a reusable function to take care of this?

Brend-Smits (changes requested):
I love the idea. 
I do think long term we should move to some sort of dynamodb table or something as those are a bit more scalable than ec2 tags. However, this looks great! 

Please have a look at my comment, after resolving that we are good to merge.

lambdas/functions/control-plane/src/scale-runners/scale-up.ts
edersonbrilhante: Sure.

edersonbrilhante: The dispatcher already sanitizes the tags, but I missed that the comma-separated value used for the dynamic label can have edge cases that break registration through the GitHub API. I changed it to use a semicolon instead, and updated the tests and docs.

I also added validation for the EC2 tag size. Since the tag count is exposed as a Terraform variable and could be any number, I set the maximum to 5 as a tradeoff for now. Later, we could apply the same idea to DynamoDB, for example.

Brend-Smits (approved):
LGTM :shipit:!

PR #5201, opened 2026-07-06
fix(lambda): add post-deregister busy check to prevent terminating active runners

### LINKED ISSUE #5085: Scale-down can terminate a runner that picks up a job between busy check and termination

npwolf:
## Summary

Rebases #5086 by @JVenberg on top of current `main` to resolve merge conflicts and get this fix moving. All credit for the original design and tests goes to Jack — this PR only adapts the change to the `deleteGitHubRunner`/per-runner error handling that landed on `main` since #5086 was opened.

Fixes #5085

The scale-down lambda can terminate an EC2 instance while it's actively running a job. This happens because a job can be assigned to a runner between checking its busy state and calling `TerminateInstances`.

## The race condition

Current flow in `removeRunner`:

1. Check GitHub API: "Is this runner busy?" -> `false`
2. _A job gets assigned to the runner here_
3. Deregister the runner from GitHub
4. Terminate the EC2 instance
5. The in-flight job is killed

## The fix

Add a post-deregistration busy re-check:

1. **Check busy** (fast-path to skip obviously busy runners)
2. **Deregister from GitHub** (prevents new job assignment server-side)
3. **Re-check busy** (now stable, since no new jobs can be assigned after deregistration)

## Why this is safe

Deregistering a runner does not affect in-flight jobs. The runner worker uses job-scoped OAuth credentials from the job message, not the runner registration:

- [`JobRunner.cs` lines 80-95](https://github.com/actions/runner/blob/main/src/Runner.Worker/JobRunner.cs): the worker creates its own `VssConnection` using `systemConnection` credentials from the job message
- The worker never checks runner registration status during execution
- Deregistration only affects the listener (no new job pickup), not the worker (current job)

## What changed vs #5086

`main` gained a `deleteGitHubRunner` helper with per-runner de-registration error handling (retry-friendly, doesn't terminate on partial failure) after #5086 was opened, which is what caused the merge conflict. This PR keeps that behavior intact and layers the post-deregister busy re-check on top of it.

## Test plan

- [x] All existing tests pass
- [x] New test: runner that becomes busy between deregister and re-check is NOT terminated
- [x] New test: runner that returns 404 on post-deregister busy check IS terminated (runner fully removed from GitHub)

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
@@ -360,6 +360,61 @@ describe('Scale down runners', () => {
         checkNonTerminated(runners);
       });
 
+      it(`Should not terminate a runner that became busy between deregister and post-deregister check.`, async () => {
+        // setup: runner appears idle on the pre-deregister check, deregister succeeds,
+        // but the post-deregister re-check finds it busy (a job was assigned in between)
+        const runners = [
+          createRunnerTestData('race-condition-1', type, MINIMUM_TIME_RUNNING_IN_MINUTES + 1, true, false, false),
+        ];
+
+        mockGitHubRunners(runners);
+        mockAwsRunners(runners);
+
+        // First call (pre-deregister) returns not-busy, second call (post-deregister) returns busy
+        const busyCheckMock =
+          mockOctokit.actions[type === 'Repo' ? 'getSelfHostedRunnerForRepo' : 'getSelfHostedRunnerForOrg'];
+        busyCheckMock
+          .mockImplementationOnce(() => ({ data: { busy: false } }))
+          .mockImplementationOnce(() => ({ data: { busy: true } }));
+
+        // act
+        await expect(scaleDown()).resolves.not.toThrow();
+
+        // assert: runner should NOT be terminated
+        checkTerminated(runners);
+        checkNonTerminated(runners);
+      });
+
+      it(`Should terminate a runner when the post-deregister busy check returns 404.`, async () => {
+        // setup: after deregistration, GitHub API returns 404 (runner fully removed from GitHub)
+        const runners = [
+          createRunnerTestData('deregistered-404', type, MINIMUM_TIME_RUNNING_IN_MINUTES + 1, true, false, true),
+        ];
+
+        mockGitHubRunners(runners);
+        mockAwsRunners(runners);
+
+        const error404 = new RequestError('Runner not found', 404, {
+          request: {
+            method: 'GET',
+            url: 'https://api.github.com/test',
+            headers: {},
+          },
+        });
+
+        // First call (pre-deregister) returns not-busy, second call (post-deregister) throws 404
+        const busyCheckMock =
+          mockOctokit.actions[type === 'Repo' ? 'getSelfHostedRunnerForRepo' : 'getSelfHostedRunnerForOrg'];
+        busyCheckMock.mockImplementationOnce(() => ({ data: { busy: false } })).mockRejectedValueOnce(error404);
+
+        // act
+        await expect(scaleDown()).resolves.not.toThrow();
+
+        // assert: runner should be terminated (404 post-deregister is treated as not busy)
+        checkTerminated(runners);
+        checkNonTerminated(runners);
+      });
+
       it(`Should terminate orphan (Non JIT)`, async () => {
         // setup
         const orphanRunner = createRunnerTestData('orphan-1', type, MINIMUM_BOOT_TIME + 1, false, false, false);
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
@@ -172,6 +172,7 @@ async function removeRunner(ec2runner: RunnerInfo, ghRunnerIds: number[]): Promi
       return;
     }
 
+    // Step 1: Check busy state as a fast-path to skip runners that are obviously busy.
     const states = await Promise.all(
       ghRunnerIds.map(async (ghRunnerId) => {
         // Get busy state instead of using the output of listGitHubRunners(...) to minimize to race condition.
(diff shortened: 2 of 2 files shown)


lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -172,35 +172,60 @@ async function removeRunner(ec2runner: RunnerInfo, ghRunnerIds: number[]): Promi
...
+    }
 
-      const allSucceeded = results.every((r) => r.success);
-      const failedRunners = results.filter((r) => !r.success);
+    // Step 3: Re-check busy state after de-registration. A job may have been assigned between
+    // step 1 and step 2. After de-registration no new jobs can be assigned, so this check is
+    // now stable. If the runner is busy, the in-flight job will complete using its job-scoped
+    // OAuth token (the runner worker uses credentials from the job message, not the runner
+    // registration). We leave the instance running and it will be cleaned up as an orphan.
+    const postDeregisterStates = await Promise.all(
```
npwolf, line 214: 🤖 **[HIGH] The post-deregister busy re-check can never see "busy" in production — the race this PR targets stays open**

Step 2 deletes the runner from GitHub (only proceeds on 204). Step 3 then GETs that same runner ID, which now returns 404. `getGitHubRunnerBusyState` maps 404 → `null` → `false` ("treating as not busy", lines 84–91). So `postDeregisterStates.every(busy => busy === false)` is essentially always `true` in production → the instance is always terminated. If a job was assigned during the real race window (between the Step 1 check at :176 and the delete at :191), it is still running on the box and still gets killed. The re-check reduces to `if (allDeregisterSucceeded) terminate` plus a wasted API round-trip — the only real guard remains the pre-deregister Step 1 check, which is what already existed before this PR.

**Fix:** a busy signal that survives deregistration is required. Either re-check busy *before* deleting, or don't treat a post-delete 404 as authoritative "idle" — after your own successful DELETE, 404 is guaranteed and tells you nothing about job state. This needs a rethink of the mechanism, not a tweak.

(Independently confirmed by both a backend-correctness review and a security review of this diff.)

lambdas/functions/control-plane/src/scale-runners/scale-down.test.ts
```diff
@@ -360,6 +360,61 @@ describe('Scale down runners', () => {
...
+        mockAwsRunners(runners);
+
+        // First call (pre-deregister) returns not-busy, second call (post-deregister) returns busy
+        const busyCheckMock =
+          mockOctokit.actions[type === 'Repo' ? 'getSelfHostedRunnerForRepo' : 'getSelfHostedRunnerForOrg'];
+        busyCheckMock
+          .mockImplementationOnce(() => ({ data: { busy: false } }))
+          .mockImplementationOnce(() => ({ data: { busy: true } }));
+
+        // act
```
npwolf, line 380: 🤖 **[HIGH] This test asserts a response the real GitHub API can't return, masking the issue above**

This "became busy between deregister and re-check" test mocks the *second* (post-deregister) busy check to return `{ busy: true }`. A runner deleted with a 204 returns **404** on a subsequent GET — as the sibling 404 test elsewhere in this file correctly assumes. So this test exercises an impossible production state; it passes green while the production code actually takes the opposite branch (terminate). It gives false confidence that the guard works.

**Fix:** mock the post-deregister check to return 404 (matching real GitHub behavior after a successful delete) and assert what the code actually does in that case, rather than asserting the desired-but-unreachable behavior.

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -172,35 +172,60 @@ async function removeRunner(ec2runner: RunnerInfo, ghRunnerIds: number[]): Promi
...
-        logger.error(
-          `Failed to de-register ${failedRunners.length} GitHub runner(s) for instance '${ec2runner.instanceId}'. ` +
-            `Instance will NOT be terminated to allow retry on next scale-down cycle. ` +
-            `Failed runner IDs: ${failedRunners.map((r) => r.ghRunnerId).join(', ')}`,
-        );
-      }
+    if (postDeregisterStates.every((busy) => busy === false)) {
+      await terminateRunner(ec2runner.instanceId);
+      logger.info(`AWS runner instance '${ec2runner.instanceId}' is terminated and GitHub runner is de-registered.`);
     } else {
```
npwolf, line 223: 🤖 **[MEDIUM] Even if this branch were reachable, orphan cleanup terminates the still-busy runner within ~2 cron cycles anyway**

In the (production-unreachable) branch where Step 3 sees busy and leaves the instance running: next cycle, `listGitHubRunners` no longer includes the deregistered runner → `markOrphan` tags it as an orphan. Following cycle, `terminateOrphan` → `lastChanceCheckOrphanRunner` GETs it → 404 → `state === null` → `isOrphan = true` → `terminateRunner`. The safety check that spares `offline && busy` runners can never fire for a deregistered runner, since it's 404, not offline+busy. So the "leave it as an orphan, the job will complete" promise in the comments above is false — the job dies regardless of which branch is taken.

**Fix:** if you intend to spare busy runners, they need to stay registered (so busy state remains queryable) or be tagged with a marker the orphan-cleanup path explicitly honors as "keep until GitHub reports it idle/gone," rather than relying on 404 as a safe-to-kill signal.

npwolf (commented):
🤖 Automated review summary (via /datagrail:code-review --report)

Ran a backend-correctness pass and a security pass on this diff. Both independently converged on the same core issue, which I verified directly against the code: **the post-deregister busy re-check can't observe "busy" in production**, because deregistering the runner makes it 404 on the next GET, and 404 is mapped to "not busy." So the fix doesn't close the TOCTOU race it targets — whether the re-check sees idle (terminate immediately) or the rare eventually-consistent "busy" (terminate via orphan cleanup ~2 cycles later), the in-flight job dies either way. See the 3 inline comments for details, including a test that mocks a response the real API can't return.

The rebase itself (resolving conflicts against current `main`) is clean and preserves existing behavior — this is a pre-existing design gap inherited from the original #5086, not something introduced by the rebase. Flagging so the mechanism can be revisited before merge.

npwolf:
Closing — the review above surfaced that this mechanism (and the original #5086) doesn't actually close the TOCTOU race: the post-deregister busy re-check can't observe "busy" once the runner is deregistered (it 404s, which is treated as idle), so in-flight jobs can still be killed. Will re-approach with a design that keeps the busy signal alive post-deregistration, or reorders the checks differently, and open a fresh PR.

PR #5203, opened 2026-07-07
refactor(runners): split runner provider layout

edersonbrilhante:
## Description

Refactors the existing EC2 runner implementation behind provider-specific boundaries while preserving EC2 as the only supported/default runner provider in this PR.

This prepares the codebase for future AWS runner providers without adding MicroVM support yet:

- renames the control-plane AWS runner implementation to EC2-specific files (`ec2-runners.ts`, `ec2-runners.d.ts`)
- adds provider interfaces and registries for scale-up, scale-down, and pool flows
- moves EC2-specific scale-up, scale-down, pool, runner creation, and dynamic-label logic into EC2 modules while keeping orchestration code provider-neutral
- extracts shared GitHub runner/JIT config flow and scale-runner types
- decouples webhook dynamic-label dispatch through AWS/provider helpers and keeps the EC2 label policy implementation behind the EC2 provider
- renames dynamic label policy configuration from `ec2_dynamic_labels_policy` / `ec2DynamicLabelsPolicy` to `aws_dynamic_labels_policy` / `awsDynamicLabelsPolicy`
- wires Terraform defaults so scale-up and pool payloads carry the EC2 provider type where needed, while `idle_config` remains provider-neutral
- updates configuration docs and generated module README tables
- keeps the existing test file layout to reduce review noise

## Test Plan

Replace your config to use aws_dynamic_labels_policy and run apply. Everything should work as today

## Related Issues

N/A


lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -174,50 +153,3 @@ function notAccepted(body: WorkflowJobEvent): Response {
   );
   return { statusCode: 202, body: notAcceptedErrorMsg };
 }
-
-function sanitizeGhrLabels(labels: string[]): string[] {
```
edersonbrilhante, line 178: just moved to https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5203/changes#diff-03eb27e722ce50c350e259e1ea49be472c07fb6759d6edb388efa96a60f7dc04R30-R75

lambdas/functions/webhook/src/runners/dispatch.ts
```diff
@@ -87,29 +81,14 @@ async function handleWorkflowJob(
...
-      }
-      for (const v of violations) {
-        logger.warn(`Queue ${q.id}: dynamic label '${v.label}' does not match policy (${v.reason}); trying next match`);
-      }
-    }
-
-    if (compliant) {
-      targets = [compliant];
-      labelsToSend = [...nonGhrLabels, ...sanitizedGhrLabels];
+    // Dynamic labels present: prefer the first provider-compliant queue. The
```
edersonbrilhante, line 84: Logic moved to selectAwsDynamicLabelQueue

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -146,30 +110,6 @@ async function getInstallationId(ghesApiUrl: string, org: string): Promise<numbe
...
-      logger.info(`Runner ${ec2Instance.instanceId} is still booting and counted as part of the pool`);
-    } else {
-      logger.debug(
-        `Runner ${ec2Instance.instanceId} is not idle in GitHub nor booting and not counted as part of the pool`,
-      );
-    }
-  }
-  return numberOfRunnersInPool;
-}
-
```
edersonbrilhante: Moved to https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5203/changes#diff-f2f62735dbcc53c60a207cb885638c52a3788c08eb61ede281bbdfa371be881eR84-R106

lambdas/functions/webhook/src/sqs/index.ts
```diff
@@ -22,13 +24,14 @@ export interface MatcherConfig {
   exactMatch: boolean;
   bidirectionalLabelMatch?: boolean;
   enableDynamicLabels?: boolean;
-  ec2DynamicLabelsPolicy?: Ec2DynamicLabelsPolicy | null;
+  awsDynamicLabelsPolicy?: AwsDynamicLabelsPolicy | null;
 }
```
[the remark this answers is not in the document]
edersonbrilhante: ec2DynamicLabelsPolicy is considered experimental

lambdas/functions/webhook/src/runners/ec2-dynamic-labels.ts
```diff
@@ -0,0 +1,44 @@
...
+  nonGhrLabels: string[],
+  sanitizedGhrLabels: string[],
+): Ec2DynamicLabelDispatchTarget | undefined {
+  for (const queue of matches) {
+    if (!queue.matcherConfig.enableDynamicLabels) {
+      logger.warn(`Queue ${queue.id} matches non-dynamic labels but does not allow dynamic labels; trying next match`);
+      continue;
+    }
+
+    const violations = violationsAgainstPolicy(sanitizedGhrLabels, queue.matcherConfig.awsDynamicLabelsPolicy);
```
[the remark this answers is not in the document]
edersonbrilhante: it is considered experimental

variables.tf
```diff
@@ -729,14 +730,15 @@ variable "enable_dynamic_labels" {
...
 }
 
-variable "ec2_dynamic_labels_policy" {
+variable "aws_dynamic_labels_policy" {
   description = <<-EOT
     Experimental! Can be removed / changed without trigger a major release.
-    Optional policy for dynamic EC2 override labels evaluated by the webhook
-    dispatcher. Only effective when `enable_dynamic_labels = true`.
+    Optional AWS dynamic label policy evaluated by the webhook dispatcher.
+    Only effective when `enable_dynamic_labels = true`.
```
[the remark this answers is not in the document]
edersonbrilhante: It was marked as experimental

Brend-Smits (commented):
I love the idea for this, I need a bit more time to review the changes though.
Have you been running this change in your environment?

My reviews are a little bit more delayed at the moment due to my vacation.

lambdas/functions/control-plane/src/pool/ec2-pool.ts
```diff
@@ -0,0 +1,106 @@
...
+          scaleErrors: config.scaleErrors,
+        },
+        numberOfRunners,
+        githubInstallationClient,
+        'pool-lambda',
+      ),
+  };
+}
+
+export function calculateEc2PoolSize(ec2runners: RunnerList[], runnerStatus: Map<string, RunnerStatus>): number {
```
guicaulada: I believe this function was modified in #5078 and this branch was cut before that change was on main, so this implementation of `calculateEc2PoolSize` is missing the `includeBusyRunners` support that was added on that PR.

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -2,57 +2,39 @@ import { Octokit } from '@octokit/rest';
...
-interface RunnerStatus {
-  busy: boolean;
-  status: string;
+  type: PoolRunnerProviderType;
 }
 
 export async function adjust(event: PoolEvent): Promise<void> {
-  logger.info(`Checking current pool size against pool of size: ${event.poolSize}`);
+  const runnerProviderType = event.type;
```
guicaulada: I noticed PoolEvent.type is now required here, when passed to `createPoolRunnerProviderFromEnv` it throws an error. 

For scale-up we have a `getScaleUpRunnerProviderType` that returns a default type and for scale-down we have `getDefaultScaleDownRunnerProviderType`, shouldn't we also have a `defaultPoolRunnerProviderType` and use: 
```suggestion
  const runnerProviderType = event.type?.trim() || defaultPoolRunnerProviderType();.
```

lambdas/functions/webhook/src/runners/ec2-dynamic-labels.ts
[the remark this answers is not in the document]
guicaulada: An SSM parameter still containing the `ec2DynamicLabelsPolicy` would be silently ignored and `violationsAgainstPolicy` would return no violations and previously rejected `ghr-ec2-*` labels would get dispatched.

Wouldn't be better to read the legacy key as a fallback, even though it's experimental? Or at least log a warning when the legacy key is present without the new one?

lambdas/functions/control-plane/src/scale-runners/scale-up-provider-registry.ts
```diff
@@ -0,0 +1,26 @@
...
+export function getDefaultScaleUpRunnerProviderType(): ScaleUpRunnerProviderType {
+  return scaleUpRunnerProviderStrategies[0].type;
+}
+
+export function createScaleUpRunnerProviderFromEnv(
+  type: ScaleUpRunnerProviderType,
+  environment: string,
+  scaleErrors: string[],
+): ScaleUpRunnerProvider {
+  const strategy = scaleUpRunnerProviderStrategies.find((strategy) => strategy.type === type);
```
guicaulada: nit: should type be case-insensitive? "EC2" vs. "ec2"?

This is more so to protect users from themselves, misconfiguration can only happen on purposeful user action so... It's just a nit.

lambdas/functions/control-plane/src/pool/pool.test.ts
```diff
@@ -414,7 +419,7 @@ describe('Test simple pool.', () => {
     it('Should ignore the maximum when set to -1 (unlimited).', async () => {
       process.env.RUNNERS_MAXIMUM_COUNT = '-1';
       // 2 idle of 4 running, pool of 10 tops up with 8 regardless of how many are already running.
-      await adjust({ poolSize: 10 });
+      await adjust({ poolSize: 10, type: 'ec2' });
```
guicaulada, line 459: I noticed every test here on `pool.test.ts` and `lambda.test.ts` was updated to include `type: 'ec2'`, should we include a test for `adjust({ poolSize: 10 })` to account for the transitive state of an environment mid-upgrade?

lambdas/functions/control-plane/src/scale-runners/github-runner.ts
```diff
@@ -0,0 +1,363 @@
...
+        runnerId,
+        error: error instanceof Error ? error.message : String(error),
+      });
+    }
+  }
+
+  if (failedRunnerIds.length > 0) {
+    logger.error('Failed to create JIT config for some runners', {
+      failedRunnerIds,
+      totalRunners: runnerIds.length,
```
guicaulada: There were a few changes to log keys here that might impact observability setups, for example:
- `instance` -> `runnerId`
- `totalInstances` -> `totalRunners`
- `failedInstanceCount` -> `failedRunnerCount`

I would rather we left these changes to a different PR, observability drift should come with a release note since it might impact dashboards and alerts that rely on them.

lambdas/functions/webhook/src/sqs/index.ts
```diff
@@ -22,13 +24,14 @@ export interface MatcherConfig {
...
 
 export type RunnerConfig = RunnerMatcherConfig[];
 
 export interface RunnerMatcherConfig {
   matcherConfig: MatcherConfig;
+  runnerProvider?: RunnerProvider;
```
guicaulada: I don't think any Terraform writes to this variable, is this here just for future implementation of MicroVM?

lambdas/functions/webhook/src/sqs/index.ts
```diff
@@ -2,10 +2,12 @@ import { SQS, SendMessageCommandInput } from '@aws-sdk/client-sqs';
 import { WorkflowJobEvent } from '@octokit/webhooks-types';
 import { createChildLogger, getTracedAWSV3Client } from '@aws-github-runner/aws-powertools-util';
 
-import { Ec2DynamicLabelsPolicy } from '../runners/dynamic-labels-policy';
+import type { AwsDynamicLabelsPolicy } from '../runners/aws-dynamic-labels-policy';
 
 const logger = createChildLogger('sqs');
 
+export type RunnerProvider = string;
```
guicaulada: Should this type be an union of strings? For example:
```ts
export type RunnerProvider = 'ec2' | 'microvm';
```

Same goes for other "RunnerPovider" types in pool, scale-up and scale-down.

lambdas/functions/control-plane/src/scale-runners/scale-down-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface RunnerInfo extends RunnerList {
+  owner: string;
+  type: string;
+}
+
+export type ScaleDownRunnerProviderType = string;
+
+export interface ScaleDownRunnerProvider {
+  name: string;
```
guicaulada: This one exposes a `name` while `ScaleUpRunnerProvider` and `PoolRunnerProvider` expose a `type`.

lambdas/functions/control-plane/src/scale-runners/scale-down-provider.ts
```diff
@@ -0,0 +1,32 @@
...
+  list(environment: string, orphan?: boolean): Promise<RunnerList[]>;
+  bootTimeExceeded(runner: RunnerInfo): boolean;
+  markOrphan(id: string): Promise<void>;
+  unmarkOrphan(id: string): Promise<void>;
+  terminate(id: string): Promise<void>;
+}
+
+export interface ScaleDownRunnerProviderStrategy {
+  type: ScaleDownRunnerProviderType;
+  create(): ScaleDownRunnerProvider;
```
guicaulada: Both scale-up and pool use `createFromEnv`, this is the only one that uses `create`, it's a bit weird how scale-down has a different provider contract than the other two.

I also noticed a bit of repetition between RunnerProviderType, RunnerProvider and RunnerProviderStrategy, maybe it's something we can extract to a shared library so all contracts follow the same patterns?

ScaleDownRunnerProvider could be an union between a shared RunnerProvider interface.
RunnerProviderType could be shared between scale-up, scale-down and pool.

lambdas/functions/control-plane/src/scale-runners/scale-up-provider-registry.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  if (!strategy) {
+    throw new Error(`Unsupported scale-up runner provider type: ${type}`);
```
guicaulada: I noticed scale-down and scale-up follow this format: `Unsupported scale-up runner provider type: ${type}` while pool uses `Unsupported pool runner provider type '${type}'`.

I think we should standardize on the quoted option in case type is a empty string this would make this clear.

lambdas/functions/control-plane/src/scale-runners/ec2-scale-down.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { RunnerList } from './../aws/ec2-runners.d';
+import type {
+  RunnerList as ScaleDownRunnerList,
+  ScaleDownRunnerProvider,
+  ScaleDownRunnerProviderStrategy,
+} from './scale-down-provider';
+
+export function createEc2ScaleDownProvider(): ScaleDownRunnerProvider {
+  return {
+    name: 'EC2',
```
guicaulada: Shouldn't this be `type: 'ec2'` like the others?

lambdas/functions/control-plane/src/scale-runners/ec2-scale-up.ts
```diff
@@ -0,0 +1,94 @@
...
+          logger.debug('EC2 override config parsed from labels', { ec2OverrideConfig });
+        }
+      }
+
+      return { runnerLabels, state: { ec2OverrideConfig } };
+    },
+    getCurrentRunners: async (_state, { runnerType, runnerOwner }) =>
+      (await listEC2Runners({ environment: config.environment, runnerType, runnerOwner })).length,
+    createRunners: async ({ githubRunnerConfig, numberOfRunners, githubInstallationClient, state }) => {
+      const ec2State = state as Ec2ScaleUpState;
```
guicaulada: This cast is unchecked, on pool we use a type parameter `TRunner` to keep it consistent, maybe we could have a similar TState for ScaleUpRunnerProvider?

For example:
```ts
export interface ScaleUpRunnerProvider<TState = unknown> {
  type: ScaleUpRunnerProviderType;
  prepareGroup(messageLabels: string[]): Promise<{ runnerLabels: string[]; state: TState }>;
  getCurrentRunners(state: TState, input: CurrentRunnersInput): Promise<number>;
  createRunners(input: CreateScaleUpRunnersInput<TState>): Promise<string[]>;
}
```

And we could use `ScaleUpRunnerProvider<Ec2ScaleUpState>` to get rid of this unchecked cast.

guicaulada (commented):
Overall this looks good to me and pretty clean for 2,000+ moved lines!

The provider seams are sensibly placed for future MicroVM work.

Two things I think need to happen before merge: a rebase that ports #5078 (pool_include_busy_runners) into the new pool provider, and a default for the pool event's type so existing scheduler payloads don't silently disable pool top-up mid-upgrade.

The SSM policy fail-open is worth a bridge or a loud warning even under the experimental banner. Happy to re-review after the rebase.

One other thing I'd like to see is a shared contract between scale-up, scale-down and pool. Currently adding a second provider means implementing three subtly different contracts. I'd like to see a shared "RunnerProvider" contract across all three. A lot of the review comments I made seem to point towards a missing opportunity to share some types and implementation. They diverge just slightly and seem hard to maintain, making changes to the RunnerProvider contract currently means changing multiple places where it's kind of re-implemented.

lambdas/functions/control-plane/src/pool/ec2-pool.ts
edersonbrilhante: Fixed

lambdas/functions/control-plane/src/pool/pool.test.ts
edersonbrilhante: Addressed in 742aa8fd. PoolEvent.type is optional and defaults to EC2; pool.test.ts now covers adjust({ poolSize: 10 }) and verifies the EC2 list and top-up behavior for legacy scheduler payloads.

lambdas/functions/webhook/src/sqs/index.ts
edersonbrilhante: It is passing now https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5203/changes#diff-295320c3fa414d9ec364d05c682024342e6b9e5aa61df4a1b957af323e84610eR3-R9

lambdas/functions/control-plane/src/pool/pool.test.ts
edersonbrilhante: Added a test case without a type `ec2`

lambdas/functions/webhook/src/runners/ec2-dynamic-labels.ts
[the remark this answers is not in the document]
edersonbrilhante: You are right. I added back with a comment to be remove in a future release

lambdas/functions/control-plane/src/pool/pool.ts
edersonbrilhante: I refactored it. I create a external lib for runner provider. And a factory for interfaces. So it can be centralised this logic

lambdas/functions/control-plane/src/scale-runners/scale-up-provider-registry.ts
edersonbrilhante: I normalized in a external lib

lambdas/functions/control-plane/src/scale-runners/scale-down-provider.ts
edersonbrilhante: I was applying your suggestion, and I noticed a bit hard to maintain after sometime not remembering the code. So I refactored it, I created a external lib(for now just a interface and simple runner provider check), and a interface registry and a bit more straightforward

lambdas/functions/control-plane/src/scale-runners/scale-down-provider.ts
edersonbrilhante: fixed

lambdas/functions/control-plane/src/scale-runners/scale-up-provider-registry.ts
edersonbrilhante: fixed

lambdas/functions/control-plane/src/scale-runners/ec2-scale-up.ts
edersonbrilhante: I changed it.

lambdas/functions/webhook/src/sqs/index.ts
edersonbrilhante: I refactored this

lambdas/functions/control-plane/src/scale-runners/github-runner.ts
edersonbrilhante: Makes sense. Fixed

edersonbrilhante:
> I love the idea for this, I need a bit more time to review the changes though. Have you been running this change in your environment?
> 
> My reviews are a little bit more delayed at the moment due to my vacation.

Yes I am running. No issue so far

lambdas/functions/control-plane/src/pool/ec2-pool.ts
```diff
@@ -0,0 +1,118 @@
...
+    createRunners: createEc2PoolRunners,
+  };
+}
+
+export function calculateEc2PoolSize(
+  ec2runners: RunnerList[],
+  runnerStatus: Map<string, RunnerStatus>,
+  includeBusyRunners = false,
+): number {
+  // Runner should be considered idle if it is still booting, or is idle in GitHub
```
stuartp44, line 61: Is it worth creating a new "booting" state with a tag added by the scaleup lambda and removed by launch/cloudinit once it's booted? Rather than inferring a state when it's booting? I not a big fan of guesstimating.

lambdas/functions/control-plane/src/pool/pool.test.ts
```diff
@@ -25,7 +28,7 @@ vi.mock('@octokit/rest', () => ({
   }),
 }));
 
-vi.mock('./../aws/runners', async () => ({
+vi.mock('./../aws/ec2-runners', async () => ({
   listEC2Runners: vi.fn(),
   // Include any other functions from the module that might be used
```
stuartp44, line 33: This looks like a cleanup potential?
```suggestion
```

lambdas/functions/control-plane/src/pool/pool.ts
```diff
@@ -68,36 +51,35 @@ export async function adjust(event: PoolEvent): Promise<void> {
     runnerNamePrefix,
   );
 
-  // Look up the managed ec2 runners in AWS, but running does not mean idle
-  const ec2runners = await listEC2Runners({
+  // Look up the managed provider runners, but running does not mean idle.
```
stuartp44, line 54:
```suggestion
  // Look up the instances we manage within this process, but a running instance does not mean it's idle, so we need to check that an idle instance is exactly idle.
```

lambdas/functions/webhook/src/runners/ec2-dynamic-labels.ts
```diff
@@ -0,0 +1,61 @@
...
+export type Ec2DynamicLabelDispatchTarget = AwsDynamicLabelDispatchTarget;
+
+function resolveEc2DynamicLabelsPolicy(queue: RunnerMatcherConfig) {
+  const hasAwsDynamicLabelsPolicy = Object.prototype.hasOwnProperty.call(queue.matcherConfig, 'awsDynamicLabelsPolicy');
+  const hasLegacyEc2DynamicLabelsPolicy = Object.prototype.hasOwnProperty.call(
+    queue.matcherConfig,
+    'ec2DynamicLabelsPolicy',
+  );
+
+  if (!hasAwsDynamicLabelsPolicy && hasLegacyEc2DynamicLabelsPolicy) {
```
guicaulada: The fallback keys on property presence, so a config with `awsDynamicLabelsPolicy: null` alongside a legacy `ec2DynamicLabelsPolicy` silently drops the legacy policy, the new test "prefers the new AWS dynamic labels policy when both keys are present" codifies this. Since the root module always emits `awsDynamicLabelsPolicy` (as `null` when unset), any transitional or hand-written SSM config carrying both keys loses a policy that exists to block AMI/subnet/instance-type injection.

`null` already means "no policy configured", so falling back on nullish value instead of key absence is strictly safer:
```ts
return queue.matcherConfig.awsDynamicLabelsPolicy ?? legacyPolicyOrUndefined(queue);
```

The `dispatch.test.ts` case would flip to expect the legacy policy to be enforced.

lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -342,38 +339,42 @@ export function newestFirstStrategy(a: RunnerInfo, b: RunnerInfo): number {
...
-    environment,
-  });
+async function listRunners(environment: string, runnerProvider: ScaleDownRunnerProvider) {
+  return await runnerProvider.list(environment);
 }
 
-function filterRunners(ec2runners: RunnerList[]): RunnerInfo[] {
-  return ec2runners.filter((ec2Runner) => ec2Runner.type && !ec2Runner.orphan) as RunnerInfo[];
+function filterRunners(runners: RunnerList[]): RunnerInfo[] {
+  return runners.filter((runner) => runner.owner && runner.type && !runner.orphan) as RunnerInfo[];
```
guicaulada, line 349: This changed from `runner.type && !runner.orphan` to `runner.owner && runner.type && !runner.orphan`. 

An instance with a type tag but a missing owner tag was previously evaluated and eventually orphan-marked; now it's skipped forever, so scale-down will never terminate it. Both tags are set together at launch so this likely never happens in practice, but if the stricter filter is intentional, a comment stating the invariant (or a test) would keep it from reading as an accidental behavior change.

lambdas/functions/control-plane/src/pool/ec2-pool.ts
```diff
@@ -0,0 +1,118 @@
...
+  subnets: string[];
+  launchTemplateName: string;
+  ec2instanceCriteria: CreateEC2RunnerConfig['ec2instanceCriteria'];
+  amiIdSsmParameterName?: string;
+  tracingEnabled?: boolean;
+  onDemandFailoverOnError: string[];
+  scaleErrors: string[];
+}
+
+function loadEc2PoolProviderConfig(): Ec2PoolProviderConfig {
```
guicaulada: This duplicates `loadEc2ScaleUpProviderConfig` in `ec2-scale-up.ts` almost field-for-field (same env vars, same JSON casts). A shared loader in `scale-runners/ec2.ts`, which both files already import, with the scale-up variant adding `useDedicatedHost` would prevent the two drifting.

lambdas/functions/control-plane/src/scale-runners/ec2.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Octokit } from '@octokit/rest';
+import type { Tag } from '@aws-sdk/client-ec2';
+
+import { createRunner, tag, terminateRunner } from '../aws/ec2-runners';
+import type { RunnerInputParameters } from '../aws/ec2-runners.d';
+import { createStartRunnerConfig } from './github-runner';
+import type { GitHubRunnerMetadata, StartRunnerConfigOptions } from './github-runner';
+import type { CreateGitHubRunnerConfig, LambdaRunnerSource } from './types';
+
+const logger = createChildLogger('ec2-scale-up');
```
guicaulada: This module also serves the pool flow, so pool-initiated runner creation logs under a misleading child name. Suggest 'ec2' or 'ec2-runners'.

modules/runners/scale-up.tf
```diff
@@ -54,6 +54,7 @@ resource "aws_lambda_function" "scale_up" {
       RUNNER_LABELS                            = lower(join(",", var.runner_labels))
       RUNNER_GROUP_NAME                        = var.runner_group_name
       RUNNER_NAME_PREFIX                       = var.runner_name_prefix
+      RUNNER_PROVIDER_TYPE                     = "ec2"
```
guicaulada, line 57: Scale-up gets this env var explicitly while scale-down relies on the unset-defaults-to-ec2 path in `resolveRunnerProviderType`. Both work, but setting it on the scale-down lambda too would make the contract visible in Terraform rather than implicit in the lib default.

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
```diff
@@ -2729,50 +2729,144 @@ describe('Retry mechanism tests', () => {
...
+      return '1';
+    } else if (name === `${process.env.PARAMETER_GITHUB_APP_ID_NAME}`) {
+      return `${process.env.GITHUB_APP_ID}`;
+    } else {
+      throw new Error(`ParameterNotFound: ${name}`);
+    }
+  });
+}
+
 describe('parseEc2OverrideConfig', () => {
```
guicaulada, line 2826: Understood the test layout is kept intentionally to reduce review noise, can we track the follow-up (move these ~90 cases to `ec2-labels.test.ts`, rename `runners.test.ts` → `ec2-runners.test.ts`) in an issue so the split doesn't get forgotten?

guicaulada (changes requested):
Thank you for addressing all my concerns!

I have a few other comments, but besides the fallback suggestion on `ec2-dynamic-labels.ts` I don't think think there's anything blocking. I would like the fallback to use legacy by default on `null` before merging just to be extra safe though.

I also think we should either change the title or add a breaking change block to the PR description.

The rename `ec2_dynamic_labels_policy` -> `aws_dynamic_labels_policy `(and `ec2DynamicLabelsPolicy` -> `awsDynamicLabelsPolicy` in `multi_runner_config` / r`unner_matcher_config`) is a hard break for anyone setting the old name.

Terraform object types reject unknown attributes, so upgrades fail at plan. That's allowed under the "Experimental" contract, but with a `refactor:` title release-please will ship this without any breaking-change signal.

Suggest either a `BREAKING CHANGE:` footer or an explicit migration note in `docs/configuration.md` ("rename X to Y before upgrading"). The JS-side legacy fallback covers the SSM config window but not the plan failure, which is what users hit first.

guicaulada (approved):
Thank you! LGTM!

lambdas/functions/control-plane/src/scale-runners/ec2-scale-down.ts
stuartp44, line 1: Should there be a logger child here?

lambdas/libs/runner-provider/src/index.ts
```diff
@@ -0,0 +1,28 @@
+// TODO: Add MicroVM when its webhook and control-plane provider implementations are available.
```
stuartp44, line 1: > // TODO: Add MicroVM when its webhook and control-plane provider implementations are available.

This left over?

stuartp44 (approved):

lambdas/functions/control-plane/src/pool/pool.test.ts
edersonbrilhante: yes

edersonbrilhante: a following pr

lambdas/libs/runner-provider/src/index.ts
edersonbrilhante: I mark for a following PR. I can remove

lambdas/functions/control-plane/src/scale-runners/ec2-scale-down.ts
edersonbrilhante: I need to check. I don't remember

lambdas/functions/control-plane/src/pool/ec2-pool.ts
edersonbrilhante: Hum. I will think about

lambdas/functions/control-plane/src/pool/pool.ts
edersonbrilhante: Let me use this in a following pr

PR #5204, opened 2026-07-09
feat: warm pool — stop/start idle runners for fast restart

Brend-Smits:
## Description

Adds a warm pool feature that reduces runner startup latency from 2–5 minutes to 10–30 seconds by stopping idle instances instead of terminating them.

Three-tier instance lifecycle:
- **Hot** (running) — 0s startup, full compute cost
- **Warm** (stopped) — 10–30s startup, EBS-only cost
- **Cold** (terminated) — 2–5 min startup, zero cost

Key components:
- DynamoDB state table for warm pool tracking
- Scale-down: stops instances instead of terminating (with AMI staleness eviction)
- Scale-up: queries DynamoDB and starts warm instances before cold-launching
- Pool lambda: `pool_strategy = "warm"` for zero-idle-compute deployments
- Full spot instance support via persistent spot requests
- CloudWatch metrics: `WarmPoolStartLatency`, `WarmPoolEvicted`, etc.

See [docs/warm-pool.md](docs/warm-pool.md) and [ADR-001](docs/adr/001-warm-pool-hibernation.md) for full design.

## Test Plan

- 540 unit tests passing (vitest)
- 6 Terraform integration tests (`warm-pool.tftest.hcl`)
- Manual validation of DynamoDB operations and EC2 stop/start in dev environment

## Related Issues

N/A — new feature


edersonbrilhante:
Great idea.

Brend-Smits:
@copilot+claude-opus-4.8:high resolve the merge conflicts in this pull request. There is a major refactor ongoing on several parts in the control plane. Take this into account and make the needed changes

Brend-Smits:
## Bug: Persistent spot requests leak unmanaged instances

Found this while investigating ~10 orphaned instances in our dev account running the warm pool branch. The instances had **zero tags**, no runner name, and kept respawning after termination.

### Root cause

When a spot instance is created via `CreateFleet` and then **stopped** (for the warm pool), the underlying spot request transitions to a **persistent** state. Later, when the warm pool eviction logic calls `terminateRunner()`, it terminates the EC2 instance but does **not cancel the associated spot request**. The persistent spot request then automatically launches a **new replacement instance** — without tags, without user data context, and without runner registration.

### Evidence from CloudTrail

The `RunInstances` events for the orphaned instances show:
```
userAgent: "spot.amazonaws.com"
invokedBy: "spot.amazonaws.com"
role: "AWSServiceRoleForEC2Spot"
tagSpecificationSet: null
```

This confirms the instances are being launched autonomously by the EC2 Spot service to fulfill persistent spot requests

### Required fixes

1. **Cancel spot requests on eviction/termination**: `retireEc2Runner()` and `evictStaleWarmInstances()` must look up and cancel any associated spot instance request (`CancelSpotInstanceRequests`) before or after terminating the instance. Otherwise the spot request just respawns a new instance.

2. **Set `ValidUntil` on spot requests**: If persistent spot requests are intentional, they should have an expiry aligned with `maxWarmAgeHours` so they self-expire even if the eviction logic fails.

3. **Propagate tags to spot requests**: The spot requests and any instances they spawn should carry the standard `ghr:*` tags so the scale-down Lambda can identify and clean them up.

Brend-Smits:
### Follow-up: warm pool support for the experimental multi-runner v2 config

Note for reviewers: warm pool + persistent spot are fully wired for the **v1 multi-runner path** and the standalone `modules/runners`. The **experimental v2 multi-runner** config (the `orchestration_provider` / `compute_provider` schema that routes through `modules/runner-config`) does **not** yet provision the warm pool resources through its provider contract.

The `warm_pool_config`, `pool_strategy`, and `instance_target_capacity_type` inputs are accepted by the normalized v2 schema (so v2 configs validate), but enabling them on a v2 runner is currently a **no-op** until the provider contract in `modules/runner-config` emits:
- the warm pool DynamoDB table,
- the `WARM_POOL_CONFIG` / `WARM_POOL_TABLE_NAME` / `POOL_STRATEGY` / `ENABLE_PERSISTENT_SPOT` env vars on the scale-up/scale-down/pool functions,
- and the DynamoDB + EC2 stop/start + spot describe/cancel IAM.

Tracking this as a follow-up so the v1 path can land first. Happy to do it in this PR if reviewers would prefer v2 parity before merge.

tehlers320:
nice, is this potentially close to coming out soon?

Brend-Smits:
> nice, is this potentially close to coming out soon?

Not yet, I will work on it again next week. It's a bit more challenging then anticipated due to the persistent spot instances that I have to use.
Ran into some problems recently, planning to start fresh again with learnings from this iterations 👍🏼

PR #5220, opened 2026-07-22
fix(scale-up): return partial runner creation results

edersonbrilhante:
## Description

Refactor runner creation to return structured partial-success results instead of throwing `ScaleError`.

- Return created instance IDs with `retryableErrorCount` and `nonRetryableErrorCount`.
- Preserve successfully created EC2 instances when Fleet or dedicated-host requests only partially succeed.
- Classify configured scale errors, transient AWS errors, network failures, throttling, and server failures as retryable.
- Treat unclassified AWS errors as non-retryable.
- Propagate the result contract through EC2, pool, and scale-up providers.
- Convert unexpected provider exceptions into retryable result counts.
- Treat GitHub runner registration failures as retryable and terminate the affected EC2 instances.
- Retry only the corresponding number of SQS messages.
- Remove the obsolete `ScaleError` implementation and handler logic.
- Skip unsupported event types without retrying them.

## Test Plan
TBD

## Related Issues

None.


lambdas/functions/control-plane/src/lambda.ts
```diff
@@ -58,6 +57,12 @@ export async function scaleUpHandler(event: SQSEvent, context: Context): Promise
   try {
     const rejectedMessageIds = await scaleUp(sqsMessages);
```
npalm: would rename here the const name, since the return value is now the result. So a more logical name woudl be createRunnerResult

edersonbrilhante: scaleUp still returning the same rejectedMessageIds. That is a list of messageId that will send to ReportBatchItemFailures.

wadherv:
@npalm This fix is quite critical for us. The recent GitHub outages have significantly impacted our workflows, resulting in much higher queue times. Would it be possible to get a release out with these changes as soon as possible?

Thanks!

edersonbrilhante:
This PR depends on the change I did on #5203.

lambdas/functions/control-plane/src/lambda.ts
edersonbrilhante: I renamed to be less confusing

stuartp44 (approved):

PR #5224, opened 2026-07-23
fix: expose selected output attributes to avoid deprecated source warnings

### LINKED ISSUE #5159: Webhook depreciation warning
I just wanted to give some feedback. Since version 7.6.0 I started receiving this warning on `terraform apply`

```vim
╷
│ Warning: Deprecated value used
│ 
│   on .terraform/modules/github-runner/modules/webhook/direct/outputs.tf line 10, in output "webhook":
│   10:     role      = aws_iam_role.webhook_lambda
│ 
│   The deprecation originates from module.github-runner.module.webhook.module.direct[0].aws_iam_role.webhook_lambda.inline_policy
│ 
│ inline_policy is deprecated. Use the aws_iam_role_policy resource instead. If Terraform should exclusively manage all inline policy associations (the current behavior of this
│ argument), use the aws_iam_role_policies_exclusive resource as well.
│ 
│ (and one more similar warning elsewhere)
```

From what I can work out it is related to what is in the generated `.terraform/github-runner/modules/webhook/outputs.tf`

```terraform
output "role" {
  value = !var.eventbridge.enable ? module.direct[0].webhook.role : module.eventbridge[0].webhook.role
}
```

to 

```terraform 
role = {
  arn  = aws_iam_role.main.arn
  name = aws_iam_role.main.name
  id   = aws_iam_role.main.id
}
```

Not sure about making changes to the generated file though.

colachg:
## Problem

With Terraform 1.12+ and AWS provider v6, every plan/apply prints `Warning: Value derived from a deprecated source` (38 times in our deployment). See #5159.

Cause: outputs that export whole resource objects. `aws_iam_role` (deprecated `inline_policy`, `managed_policy_arns`) in `lambda`, `runners`, `runners/pool`, `runner-binaries-syncer`, `webhook/direct`, `webhook/eventbridge`, `ami-housekeeper`; `aws_s3_bucket` (deprecated `acl`, `policy`, `website_*`, `acceleration_status`, `request_payer`) in `runner-binaries-syncer`.

## Fix

Those outputs now list the resource attributes explicitly, as suggested by @edersonbrilhante, so the output contract is preserved. Only the deprecated keys are dropped — reading them is what triggers the warning. Everything else keeps the same name and type, so `module.runners.role_runner[0].arn`, `webhook.role.name` and `bucket.id` still work. No consumer changes needed: in-repo consumers only read `id` / `arn` / `name`.

## Why not wait for Terraform 1.16?

hashicorp/terraform#38778 fixes this upstream, but only in 1.16 (`v1.16.0-rc1`). This module allows `>= 1.3.0`, so users on 1.12–1.15 keep the warnings until they can upgrade. These outputs can go back to whole resources once 1.16 is common.

## Testing

`terraform validate` passes for the root module and `multi-runner`, `webhook`, `termination-watcher`, `lambda`. `terraform fmt -check` clean. Rebased on `main`. Only output expressions change, so a plan against existing state shows no resource changes.

Fixes #5159


edersonbrilhante:
This would be a breaking change in the output contract. Let's say someone uses another property that is not deprecated.

edersonbrilhante:
If going forward with this fix, I think we should keep the same output contract
example
```
      arn                   = aws_iam_role.dispatcher_lambda.arn
      assume_role_policy    = aws_iam_role.dispatcher_lambda.assume_role_policy
      create_date           = aws_iam_role.dispatcher_lambda.create_date
      description           = aws_iam_role.dispatcher_lambda.description
      force_detach_policies = aws_iam_role.dispatcher_lambda.force_detach_policies
      id                    = aws_iam_role.dispatcher_lambda.id
      max_session_duration  = aws_iam_role.dispatcher_lambda.max_session_duration
      name                  = aws_iam_role.dispatcher_lambda.name
      name_prefix           = aws_iam_role.dispatcher_lambda.name_prefix
      path                  = aws_iam_role.dispatcher_lambda.path
      permissions_boundary  = aws_iam_role.dispatcher_lambda.permissions_boundary
      tags                  = aws_iam_role.dispatcher_lambda.tags
      tags_all              = aws_iam_role.dispatcher_lambda.tags_all
      unique_id             = aws_iam_role.dispatcher_lambda.unique_id
```

colachg:
@edersonbrilhante done as you suggested — the outputs now list the full attribute set, so the contract is kept. Only the deprecated keys are gone: inline_policy and managed_policy_arns on the roles, acl / policy / website_* / acceleration_status / request_payer and the deprecated nested blocks on the bucket. Reading those is what triggers the warning, so they can't stay.

PR #5234, opened 2026-07-31
refactor(runners): introduce provider plugin framework

edersonbrilhante:
## Description

Introduce a shared runner-provider plugin framework so a provider is enabled once and its webhook, scale-up, scale-down, pool, and runner-resource capabilities are available through one registry facade.

- Move the EC2 provider implementation into the shared `runner-providers` library.
- Add a single provider configuration and capability registry for webhook and control-plane consumers.
- Add reusable scale-up, scale-down, and pool contract tests that run for every enabled provider lane.
- Add a provider skeleton that documents the interfaces required to implement another lane.

This reduces the number of control-plane and webhook locations that must be changed when adding a runner provider.

## Follow-up work

Some tests still need to move to the control-plane package instead of remaining with the EC2 plugin. The affected files contain groups of tests, so this cleanup requires moving individual test cases rather than relocating an entire block or file. It will be handled in a follow-up PR to keep this refactor reviewable.

This is why some EC2 provider tests currently import from `../../../../../../functions/control-plane`. Those dependencies should be removed in the follow-up PR.

## Test Plan

- Ran the `runner-providers:test` Nx target.
- Ran the `control-plane:test` Nx target.
- Ran Prettier checks for the runner-provider library.
- Ran ESLint for the runner-provider library.

## Related Issues

Follow-up: #5236.


lambdas/libs/runner-providers/providers.config.ts
```diff
@@ -0,0 +1,8 @@
+import { provider as ec2 } from './aws/ec2';
+import type { RunnerProviderModule } from './contracts';
+
+/**
+ * Provider plugins enabled in webhook and control-plane bundles.
+ * Installing and enabling a provider requires adding it only to this list.
+ */
+export const enabledRunnerProviders = [ec2] as const satisfies readonly RunnerProviderModule[];
```
guicaulada: This single list is imported (via `core/index.ts` line 3) by both lambdas, and it constructs the provider object whose `createControlPlanePlugin` references the pool/scale-up/scale-down implementations. 

Now the webhook bundle contains the entire EC2 control-plane implementation plus `@aws-sdk/client-ec2`, `aws-ssm-util`, and `moment`, none of which it needs or had before this PR. 

The webhook is a latency-sensitive lambda since it answers GitHub's delivery, so cold-start and bundle growth could cause issues.

Should we split the config per lane? So `providers.config.webhook.ts` imports `./aws/ec2/webhook` and `providers.config.control-plane.ts` imports `./aws/ec2/control-plane` and split `registry.ts` the same way. 

Enabling a provider becomes two one-line entries in the same directory, and each lambda bundles only its lane.

lambdas/libs/runner-providers/aws/ec2/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { RunnerProviderModule } from '../../contracts';
+
+import { createEc2ControlPlanePlugin } from './control-plane';
+import { createEc2WebhookPlugin } from './webhook';
```
guicaulada: Same issue as `providers.config.ts`, this entry point fuses both lanes, so any importer gets both.

The `export *` re-exports at lines 14–21 additionally make every internal module public API of the plugin. If the config splits per lane, this file can become two entry points (`webhook.ts` / `control-plane.ts` already exist and are the right shape) and the wildcard re-exports can go.

lambdas/libs/runner-providers/core/index.ts
```diff
@@ -0,0 +1,169 @@
+import type { Octokit } from '@octokit/rest';
+
+import { enabledRunnerProviders } from '../providers.config';
```
guicaulada: This makes core depend on the top-level plugin list and `core` depends on `providers.config` which depends on `aws/ec2` which depends on `core` types. It stays safe only while every provider module imports nothing but types from core. When trying to import a value (e.g. `resolveRunnerProviderType` inside a provider) creates a runtime circular import which fails when the module loads.

Can we move `normalizeRunnerProviderType`/`resolveRunnerProviderType` into `registry.ts` (which depends on the config), or have them take the provider-type list as a parameter.

lambdas/libs/runner-providers/core/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const providerTypes = enabledRunnerProviders.map(({ type }) => type);
+
+export type RunnerProviderType = (typeof import('../providers.config').enabledRunnerProviders)[number]['type'];
+
+export function normalizeRunnerProviderType(type: unknown): RunnerProviderType | undefined {
+  if (type === undefined) return providerTypes[0];
```
guicaulada: The default provider is now implicitly "first entry in `providers.config.ts`". The old lib had an explicit `defaultRunnerProvider = 'ec2'`. 

Once a second provider exists, reordering the config list silently changes the default for every existing deployment. Could we make the default an explicit named export rather than positional?

lambdas/libs/runner-providers/contracts.ts
```diff
@@ -0,0 +1,63 @@
...
+  denied?: string[];
+  max?: number | string;
+}
+
+export interface AwsDynamicLabelsPolicy {
+  blocked_keys?: string[];
+  restricted_keys?: Record<string, AwsDynamicLabelsValueRule>;
+}
+
+export interface AwsRunnerMatcherConfig {
```
guicaulada: This duplicates the webhook's `RunnerMatcherConfig`/`MatcherConfig` in `sqs/index.ts`. They'll drift the first time a matcher field is added. Should the webhook re-export this type (as it now does for `AwsDynamicLabelsPolicy`) or vice versa, so there's one definition?

guicaulada (changes requested):
LGTM! The only change in the relocated EC2 modules is the createStartRunnerConfig injection, the orchestrators changed only their registry call sites, and no test cases were lost with two additions to replace the deleted registry test. The contract-test lanes and the compile-checked template are nice.

Requesting changes for one reason: the shared `providers.config.ts`/`core` import chain pulls the full EC2 control-plane implementation and the EC2 SDK into the `webhook` bundle.

Splitting the config and registry per lane fixes it while keeping the "enable a provider in one place" goal. The core/config layering and the positional default provider are strong suggestions, the rest is minor.

guicaulada (approved):
Thank you for addressing my concerns! LGTM!

PR #5251, opened 2026-08-05
feat(multi-runner): add experimental EC2 provider stack

### LINKED ISSUE #5252: refactor(multi-runner): decouple EC2 Terraform logic into provider modules
## Summary

Decouple EC2-specific Terraform logic from the common multi-runner orchestration so future lane providers can plug into the module without copying or bending EC2 assumptions.

The current Terraform layout mixes common runner, queue, pool, webhook, IAM, and EC2 implementation details. As we add experimental `multi_runner_config_v2` and prepare for providers such as microVM or CodeBuild, the module should have a provider-neutral common layer plus provider-specific modules.

## Proposed Direction

- Keep common lane orchestration in a shared multi-runner layer.
- Move EC2-specific runner logic into a provider module, for example `modules/multi-runner/providers/ec2` or a similar repo-convention-aligned path.
- Split runner concerns into common inputs/outputs and provider-specific inputs/outputs.
- Split pool logic the same way: common scheduling/pool contract in the shared layer, provider-specific pool implementation in the EC2 provider module.
- Split webhook matcher/routing config from provider execution details.
- Split IAM/policies between common responsibilities and per-provider/per-lane permissions.
- Keep queue ownership and lane keys stable so existing Terraform addresses do not churn unnecessarily.
- Preserve EC2 behavior while moving implementation behind the provider boundary.

## Scope Notes

- This issue is about Terraform module structure, not implementing microVM or CodeBuild Terraform resources yet.
- EC2 should remain the only active Terraform-managed provider until new provider modules are implemented.
- The experimental `multi_runner_config_v2` shape can guide the boundary, but it should remain flexible until the module split settles.

## Acceptance Criteria

- Common multi-runner logic no longer directly owns EC2-specific runner and pool implementation details.
- EC2 provider module owns EC2 runner resources, EC2 pool resources, and EC2-specific policies.
- Common module owns shared queue, webhook matcher/routing, lane normalization, and shared policy pieces.
- Existing `multi_runner_config` EC2 lanes keep working without behavior changes.
- `multi_runner_config_v2` can route an EC2 lane through the provider module without introducing active microVM or CodeBuild support.
- Terraform validation and tflint pass for the affected modules.
- Documentation explains the common/provider split and marks non-EC2 providers as future work.

edersonbrilhante:
## Description

Adds the opt-in provider-oriented `experimental.multi_runner_config` path while preserving stable `multi_runner_config` behavior and existing `module.runners["configuration"]` addresses.

- Normalizes stable and experimental inputs into one canonical configuration with explicit global-to-lane precedence.
- Routes experimental lanes through the provider-neutral `runner-stack` and typed EC2 provider module.
- Keeps EC2 resources, validation, IAM policy fragments, instance profiles, and trust policies behind the provider boundary.
- Applies translated globals to shared queues, SSM, webhook, runner-binary syncer, AMI housekeeper, and termination-watcher callers.
- Adds focused routing, validation, plan-known input, and documentation coverage.

EC2 is the only Terraform-managed compute provider enabled here. The generic termination-watcher environment/KMS fix is excluded: `modules/termination-watcher` matches `main`, and #5274 owns that fix.

## Test Plan

- Multi-runner provider-routing tests: 30 passed, 0 failed.
- Multi-runner computed-input test: 1 passed, 0 failed.
- Focused runner-stack child tests: 5 passed, 0 failed.
- `git diff --check` passed.
- Verified `modules/termination-watcher` is identical to `main`.
- No live AWS apply was performed; Terraform tests use mocked providers.

## Related Issues

- Closes #5252.
- Depends on #5274 for termination-watcher environment-variable and SSM KMS behavior.
- #5254 is intentionally excluded; this PR targets `main` directly.


modules/multi-runner/variables.experimental.tf
```diff
@@ -0,0 +1,209 @@
+variable "multi_runner_config_v2" {
+  description = <<EOT
+Experimental runner lane configuration keyed by lane name. This v2 shape separates common runner routing from provider-specific backend configuration. The schema can change while the provider model is being finalized. When set, this variable takes precedence over stable `multi_runner_config`.
```
Brend-Smits: You might want to explain the concept of a 'lane' an what it is.

Brend-Smits (commented):
I like the general direction where this is going. Wondering if it makes sense to introduce some kind of schema for people to consume as well now that we are making these changes. 

This will be a much needed change seeing as we are moving to a new architecture that supports multiple compute variants. Please do keep in mind though that the people that are reading this documentation (like in your variables.tf), might now have all the context that you have. So try to keep the easy to consume for those folks as well (explain jargon, etc).

Do I understand it correctly that this change will be `opt-in` for this version while we test it further and in the future we will make a breaking change version update to change the default to this new config?

Great work!

modules/multi-runner/variables.experimental.tf
edersonbrilhante: Lane I mean provider. Sorry, I use internally the term "lane" to explain if a runner is deployed in arc(k8s runner) and or ec2 runner. I will rename to fix this confusion

edersonbrilhante:
@Brend-Smits I changed my first implementation. I will not touch runners. Instead I created a whole cloned runners, refactoring it to add support for a decoupled ec2 module.

PR #5265, opened 2026-08-11
feat: support native SSM expiration policies for runner token parameters

jvrplmlmn:
## Description

Registration tokens and JIT configs written to SSM are orphaned when an instance fails to boot or is terminated before registering. The only cleanup today is the scheduled SSM housekeeper lambda, which cannot delete parameters younger than a day, so orphaned parameters (billed hourly when on the Advanced tier) can live for up to ~2 days.

The new opt-in `ssm_token_ttl_seconds` attaches a native SSM Expiration policy to the token / JIT config parameters so SSM deletes leftovers itself once the TTL passes. Parameter policies require the Advanced tier, which incurs additional cost per parameter, hence disabled by default. Expiration is enforced asynchronously by SSM and is a cleanup mechanism rather than a security boundary; the housekeeper lambda remains as a backstop.

## Related Issues

- https://github.com/github-aws-runners/terraform-aws-github-runner/issues/5243


edersonbrilhante:
Nice idea. Just a small note, I am working ro refactor the module to make easier to maintain and add new features. 
I have some idea to nest some variables to organize. It is a bit hard to keep track the features. I don't mind adding this new variable, but I would propose to deprecate in following major version and move it to a nested ssm variable. 

@Brend-Smits @npalm @guicaulada what do you think?

guicaulada:
@jvrplmlmn would mind rebasing and fixing conflicts on this one :)

modules/multi-runner/variables.tf
```diff
@@ -403,6 +404,7 @@ variable "multi_runner_config" {
     }), {})
 
     ssm = optional(object({
+      token_ttl_seconds = optional(number, null)
```
edersonbrilhante: Here I would use ttl_seconds.tokens. This way opens space for extend ttl for other configs in the future

edersonbrilhante:
@guicaulada Here I would like a second opinion. How about add support for this feature in global level(v2) too. So can be configure in global level and allow override per runner entry.

https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/multi-runner/variables.experimental.storage-provider.tf#L31

Also @guicaulada the last commit not signed

guicaulada:
Weird.. why is the docs commit not signed? I thought we had fixed that already.

EDIT: Apparently we didn't fix it for forks.

edersonbrilhante (approved):
Let's just create an issue to add support in v2 as well

PR #5266, opened 2026-08-12
feat: add optional DynamoDB runner config storage

wadherv:
## Summary
  - Related to: https://github.com/github-aws-runners/terraform-aws-github-runner/issues/5243
  - Add backward-compatible `runner_config_storage` support with `ssm` as the default backend and optional `dynamodb` storage for label-scoped runner config.
  - Create one DynamoDB runner config table per runner label/module with configurable billing mode, capacity, autoscaling, TTL, PITR, SSE, deletion protection, tags,
  and CloudWatch alarms.
  - Move label-scoped bootstrap config, runner group cache, and one-time token/JIT config to DynamoDB when enabled; common GitHub App/webhook config remains in SSM.
  - Update scale-up, pool, and runner startup scripts for Linux/macOS/Windows to use the selected backend.
  - Add DynamoDB Lambda client reuse with configurable retry mode, keep-alive, max sockets, and keep-alive timeout.
  - Add `create` flag for SSM housekeeping resources and align DynamoDB token TTL with SSM housekeeping retention.
  - Scope IAM permissions by backend and DynamoDB leading key prefixes.
  - Remove unnecessary DynamoDB item metadata fields.

  ## Testing

  - `node_modules/.bin/vitest run`
  - `node_modules/.bin/tsc -p functions/control-plane/tsconfig.json --noEmit`
  - `node_modules/.bin/eslint ...`
  - `node_modules/.bin/ncc build functions/control-plane/src/lambda.ts -o /private/tmp/control-plane-ncc-check`
  - `git diff --check`


edersonbrilhante:
@wadherv interesting idea. Since you added `count` in ssm, please add a file moved to move the resource state.

yalafe:
+1 - had the same pain. I independently built the same approach (per-instance token/config in DynamoDB with a TTL, read-then-delete mirroring the SSM lifecycle) and it held up in a live run with no SSM token writes.

modules/runners/variables.tf
```diff
@@ -654,6 +654,130 @@ variable "ssm_paths" {
...
+      var.runner_config_storage.dynamodb.partition_key_name,
+      var.runner_config_storage.dynamodb.value_attribute_name,
+      var.runner_config_storage.dynamodb.ttl_attribute_name,
+    ])) == 3
+    error_message = "`runner_config_storage.dynamodb` attribute names must be distinct."
+  }
+
+  validation {
+    condition     = var.runner_config_storage.dynamodb.token_key_prefix == null ? true : trimspace(var.runner_config_storage.dynamodb.token_key_prefix) != ""
+    error_message = "`runner_config_storage.dynamodb.token_key_prefix` must be null or non-empty."
```
guicaulada, line 752: A custom token prefix changes the IAM condition to `<prefix>*`, allowing any runner using this role to read or delete another runner’s token. Please retain instance-specific access. A safe interim fix is to reject custom prefixes:

```suggestion
    condition     = var.runner_config_storage.dynamodb.token_key_prefix == null
    error_message = "Custom token_key_prefix values are not supported because runner token access must remain instance-scoped."
```

modules/runners/templates/start-runner.sh
```diff
@@ -112,6 +112,49 @@ cleanup() {
...
+  key_json=$(jq -cn --arg key_attr "$runner_config_dynamodb_partition_key_name" --arg id "$item_id" '{($key_attr):{S:$id}}')
+  expression_attribute_names=$(jq -cn --arg value_attr "$runner_config_dynamodb_value_attribute_name" '{"#value":$value_attr}')
+  if [[ "$runner_config_dynamodb_consistent_read" == "true" ]]; then
+    consistent_read_args=(--consistent-read)
+  fi
+  item_json=$(aws dynamodb get-item \
+    --table-name "$runner_config_dynamodb_table_name" \
+    --key "$key_json" \
+    "$${consistent_read_args[@]}" \
+    --projection-expression "#value" \
```
guicaulada, line 140: DynamoDB does not delete a record as soon as its TTL timestamp passes. Until that background deletion happens, `GetItem` can still return it. This read requests only the token value, so the startup script cannot tell that the stored token record has expired and will try to use it.

Please check the configured TTL attribute before using a token, and treat expired records as unavailable. Apply the same check to the Linux, macOS, and Windows startup scripts.

One option is to consume the token with a conditional `DeleteItem`: require its expiry timestamp to be later than the current time and use `ReturnValues: ALL_OLD` to return the deleted record. This checks expiry and removes the token in one operation, so two consumers cannot both retrieve it successfully. Apply this only to short-lived token records; durable bootstrap configuration has no expiry.

guicaulada (commented):
LGTM!

Minor comment on the DynamoDB TTL, not sure if it's actually relevant here.

On the prefix, I noticed if we set a token_key_prefix it allows runners to read and delete any other token from other runners which is less safe from the instance-specific access we have at SSM.

edersonbrilhante:
@wadherv your idea is great. I decoupled ssm from control plane, now ssm is a storage provider. I started a draft to make   dynamodb a second storage provider in https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5394

I can close my PRs, if you want to adjust your code to implement it.

PR #5038, opened 2026-02-18
feat(github-app): allow several github apps to be used

thomasnemer:
## Summary

  - Add support for configuring **additional GitHub Apps** to distribute API calls across multiple rate limit buckets, making the module scalable for large enterprise deployments
  - New `additional_github_apps` variable (optional, no breaking changes) accepts a list of extra apps with `id`, `key_base64`, and optionally `installation_id` (all support direct values or SSM references)
  - Lambda functions randomly select an app per authentication flow and thread the selection through the entire JWT → installation token → API call chain for consistency
  - Installation ID resolution is optimized: primary app reuses the webhook payload, additional apps use a pre-configured `installation_id` when provided, falling back to an API lookup only when needed

  ## Context

  At enterprise scale with many runner configurations, all Lambda functions (scale-up, scale-down, pool, job-retry) share a single GitHub App's rate limit bucket (15,000 req/hour on GitHub Enterprise Cloud). This ceiling cannot be raised per-app — the only way to scale is to spread load across multiple apps. This change makes the effective limit `N × 15,000` req/hour, without introducing breaking changes.

  Closes #5037

PR #5269, opened 2026-08-14
feat(github-app): allow several github apps to be used

### LINKED ISSUE #5037: Support for multiple GitHub Apps to overcome API rate limits at scale

guicaulada:
## Description

Supersedes #5038 by @thomasnemer, whose branch lives in an org-owned fork that maintainers cannot push to. This PR rebases that work onto current `main` (porting it to the compute-provider architecture introduced in #5234/#5267) and addresses the review feedback on #5038. All of Thomas's work is preserved with co-authorship.

From the original PR:

- New `additional_github_apps` variable (optional, no breaking changes) accepts extra GitHub Apps with `id`, `key_base64`, and optionally `installation_id` (direct values or SSM references).
- The control-plane lambdas (scale-up, scale-down, pool, job-retry) select one app per invocation and thread the selection through the JWT → installation token → API call chain, spreading load across N × 15,000 req/hour rate-limit buckets.
- Installation ID resolution: pre-configured `installation_id` wins, the primary app reuses the webhook payload's id, API lookup is the fallback.

Changes on top of #5038:

- **Ported to the compute-provider layout**: the `appIndex` selection lives in the orchestrators (`scale-up.ts`, `pool.ts`, `scale-down.ts`) and flows to providers via `CreateGitHubRunnerConfig.appIndex`, so provider interfaces are unchanged.
- **Restored the 404 stale-installation fallback** ([review feedback](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5038#pullrequestreview-4855605507)): when installation auth fails with 404 (app uninstalled/reinstalled while messages were in flight), the flow re-resolves the installation via the API and retries once with the same app. This now also covers stale pre-configured installation ids, and has regression tests in `octokit.test.ts`.
- Updated env wiring for the `COMPUTE_PROVIDER_TYPE` era and regenerated module docs.

## Test Plan

- `vitest`: control-plane 348 passed, compute-providers 270 passed, webhook 89 passed.
- New regression tests: stale-payload 404 retry, same-id rethrow, non-404 rethrow.
- ESLint and Prettier clean on all touched packages.
- `terraform fmt -check -recursive`, `terraform validate` (root, runners, multi-runner, ssm), `terraform test` in `modules/runners` (1 passed).
- READMEs regenerated with terraform-docs.

## Related Issues

Closes #5037. Supersedes #5038.

Co-authored with @thomasnemer.


Brend-Smits (approved):
LGTM :shipit:! Thanks for picking this up! Also thanks to @thomasnemer 🚀

PR #5275, opened 2026-08-16
feat: Make runner trust policy extendable

chris-griffin:
## Description

Adds a configuration option so that the runner trust policy can be extended. Currently, it is hardcoded via `modules/runners/policies/instance-role-trust-policy.json`

## Test Plan

- `terraform test`, `terraform validate` and `terraform fmt -check` all passed
- `tflint` passed although it did raise a warning unrelated to this change
- We have not deployed this exact version yet, but happy to do so if helpful or required

## Related Issues

Would address previously requested use cases like https://github.com/github-aws-runners/terraform-aws-github-runner/issues/4247 and https://github.com/github-aws-runners/terraform-aws-github-runner/issues/1315


edersonbrilhante:
@chris-griffin We are working in a v2 multi-runner module. In this version I added support for extendable trust policy. We have plan to deprecate v1. I am not sure if we should keep adding new features in v1.

What do you think @Brend-Smits ?

PR #5279, opened 2026-08-19
feat(storage): add shared DynamoDB runner storage

edersonbrilhante:
## Description

Adds an opt-in `aws_dynamodb` runner-config storage provider for experimental multi-runner v2 deployments.

- Creates exactly two shared tables for the whole multi-runner deployment:
  - `<prefix>-config` for durable global and entry-scoped configuration, including GitHub App credentials, webhook secret, matcher config, runner bootstrap config, and runner-group cache.
  - `<prefix>-runner-state` for TTL-backed one-time runner configuration plus provider-neutral runner lifecycle/inventory records.
- Uses `scope` and `id` keys to isolate global, runner-entry, lifecycle, and compute-resource records without creating a table per runner entry.
- Adds DynamoDB implementations for runner config, runner-group cache, matcher config, GitHub App credentials, webhook secret, and lifecycle state.
- Makes scale-up, pool, and scale-down reconcile capacity through the durable inventory while preserving provider discovery as a recovery source.
- Lets EC2 runners atomically consume only their own unexpired runner configuration and retains EC2 tags for compatibility.
- Wires least-privilege DynamoDB IAM, encrypted tables, TTL, seed readiness, and opaque configuration-version rotation into webhook and control-plane Lambdas.
- Keeps v1 and omitted v2 configuration on `aws_ssm`. The termination watcher intentionally remains on SSM in this PR.

DynamoDB selection requires direct GitHub App values; externally managed `*_ssm` references continue using the SSM provider so Terraform does not copy those external SecureStrings into state.

## Test Plan

- Lambda Vitest suite: 68 files / 1,037 tests passed.
- TypeScript `tsc --noEmit`: control-plane, storage-providers, and webhook passed.
- ESLint and Prettier checks passed for the changed Lambda packages.
- Terraform and OpenTofu tests for the new DynamoDB provider: 4/4 passed in each CLI.
- Multi-runner v2 Terraform tests: 36/36 passed.
- Stable v1 Terraform tests: 2/2 passed.
- `terraform validate`: storage provider, multi-runner, and webhook modules passed.
- Targeted TFLint for the new DynamoDB provider passed.
- `terraform fmt -check -recursive` and `git diff --check` passed.
- No deployment was performed.

## Related Issues

Credits to @wadherv #https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5266

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-down.ts
@@ -3,6 +3,11 @@ import { Endpoints } from '@octokit/types';
 import { RequestError } from '@octokit/request-error';
 import { createChildLogger } from '@aws-github-runner/aws-powertools-util';
 import { resolveComputeProviderType } from '@aws-github-runner/compute-providers/provider-types';
+import {
+  getRunnerStateStore,
+  type RunnerStateRecord,
+  type RunnerStateStore,
+} from '@aws-github-runner/storage-providers';
 import moment from 'moment';
 
 import {
@@ -22,7 +27,19 @@ const logger = createChildLogger('scale-down');
 
 type OrgRunnerList = Endpoints['GET /orgs/{org}/actions/runners']['response']['data']['runners'];
 type RepoRunnerList = Endpoints['GET /repos/{owner}/{repo}/actions/runners']['response']['data']['runners'];
-type RunnerState = OrgRunnerList[number] | RepoRunnerList[number];
+type GitHubRunnerState = OrgRunnerList[number] | RepoRunnerList[number];
+
+interface InventoryRunnerInfo extends RunnerInfo {
+  inventory?: RunnerStateRecord;
+  providerPresent?: boolean;
+}
+
+type RestorableRunnerState = Parameters<RunnerStateStore['cancelTermination']>[1];
+
+interface TerminationClaim {
+  runnerStateId?: string;
+  restoreState?: RestorableRunnerState;
+}
 
 async function getOrCreateOctokit(runner: RunnerInfo): Promise<Octokit> {
   const key = runner.owner;
@@ -67,7 +84,7 @@ async function getGitHubSelfHostedRunnerState(
   client: Octokit,
   runner: RunnerInfo,
   runnerId: number,
-): Promise<RunnerState | null> {
+): Promise<GitHubRunnerState | null> {
   try {
     const state =
       runner.type === 'Org'
@@ -161,6 +178,12 @@ async function deleteGitHubRunner(
     }
     return { ghRunnerId, status: response.status, success: response.status === 204 };
   } catch (error) {
+    if (error instanceof RequestError && error.status === 404) {
+      logger.info(
+        `GitHub runner ${ghRunnerId} for runner '${runner.id}' is already de-registered; treating cleanup as complete.`,
+      );
+      return { ghRunnerId, status: error.status, success: true };
+    }
     logger.error(
       `Failed to de-register GitHub runner ${ghRunnerId} for runner '${runner.id}'. ` +
         `Error: ${error instanceof Error ? error.message : String(error)}`,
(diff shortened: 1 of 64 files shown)


lambdas/functions/control-plane/src/scale-runners/scale-down.ts
```diff
@@ -305,37 +379,189 @@ async function lastChanceCheckOrphanRunner(runner: RunnerInfo): Promise<boolean>
...
+              computeProvider,
+              orphanEvaluation.githubRunnerExists ? parseInt(runner.githubRunnerId) : undefined,
+            );
+          } else {
+            // Preserve the provider-only recovery behavior used by the legacy path.
+            await computeProvider.terminate(runner.id);
+          }
         } else {
-          await unMarkOrphan(runner.id, computeProvider);
+          await unMarkOrphan(runner, computeProvider);
```
guicaulada, line 475: A runner can get stuck if scale-down sets its state to `terminating` and then crashes before removing it from GitHub. On the next run, the runner is still registered in GitHub, so this branch calls `unMarkOrphan`. That function does not clear `terminating`, and normal scale-down skips runners in that state—even after the termination claim expires.

Please let a later scale-down run take over an expired termination claim and retry cleanup. Before removing the runner, check whether it is now busy so we do not interrupt a job.

I reproduced this with an online, idle runner whose termination claim was one hour old: scale-down never called `beginTermination` to retry cleanup.

modules/storage-providers/aws/dynamodb/capabilities.tf
```diff
@@ -0,0 +1,243 @@
...
+
+  scale_up_iam_policy_json = {
+    for entry_id in var.entry_ids : entry_id => jsonencode({
+      Version = "2012-10-17"
+      Statement = [
+        local.github_app_read_statement,
+        local.entry_runner_group_statements[entry_id],
+        local.runner_config_write_statements[entry_id],
+        local.runner_state_write_statements[entry_id],
+      ]
```
guicaulada: The generated policies allow DynamoDB access but do not grant access to the configured table KMS keys. With a key policy that delegates permissions to IAM, the Lambda and runner roles can fail to read the encrypted tables.

Please include scoped KMS permissions for each table’s consumers and add policy tests using non-null custom key ARNs.

guicaulada (commented):
Overall looks good!

One suggestion on scale-down related to orphan runners: it would be nice if we could retry cleanup. Could be a subject for a follow-up PR.

Another point is related to KMS, I don't think the Lambda and runner roles have access to KMS to read encrypted tables.

Feel free to correct me if I'm missing something here!

Brend-Smits:
Nice work on this , the `scope`/`id` composite key plus native TTL is a clean way to avoid a table-per-entry, and reusing `computeResourceId`/`computeProvider` to keep things provider-neutral sets this up well for future compute providers.

One question on the isolation model: `RunnerConfigRecord.accessScope` (e.g. the EC2 instance ARN) is used as the partition key for the JIT/registration secret, but the interface only exposes `create()`. 
Is there a plan for the runner's own read path to be scoped by an IAM condition tied to its own compute identity (e.g. something like `dynamodb:LeadingKeys` conditioned on the instance's own ARN for EC2), so a runner can only ever fetch its own row? Right now the isolation looks like it depends on the control-plane always writing the correct scope, rather than being enforced by DynamoDB/IAM itself. Worth confirming/documenting how that's intended to work before this is broadly adopted, since it's the difference between "app convention" and "hard boundary."

On extensibility: this scope/id design also seems like a natural fit for replacing per-instance tagging (job context, etc.) with DynamoDB attributes on the same runner-state row, and eventually giving runners a place to read their own operational metadata back. If that's on the roadmap, it might be worth keeping the secret JIT-config item and any future metadata item under separate `id`s within the same `scope` from the start, so a metrics/read path never has access to the secret payload.

For instance, internally we have a similar setup which we protect using IAM policies with something like this:
```
"Condition": {
  "ForAllValues:StringEquals": { "dynamodb:LeadingKeys": ["${ec2:SourceInstanceARN}"] }
}
```

This way the instance itself is only able to read it's own data.

modules/storage-providers/aws/dynamodb/capabilities.tf
edersonbrilhante: Resolved

PR #5280, opened 2026-08-19
feat(microvm): add lifecycle hook service

edersonbrilhante:
## Description

Adds the MicroVM lifecycle-hook service and the CI, release, and integration-test changes needed to exercise the complete provider flow.

- Adds the `lambdas/services/microvm-lifecycle-hooks` service with strict lifecycle-payload validation, runner-configuration consumption, process management, graceful shutdown, and the documented Node 24/s6 runtime contract.
- Registers, builds, tests, packages, and attests the lifecycle-hook Lambda in the Lambda and release workflows.
- Replaces the separate MicroVM and multi-runner-v2 MiniStack paths with the `multi-runner-webhook` example, which configures EC2 and MicroVM lanes behind one webhook deployment.
- Adds an extensible Python smoke harness with shared webhook lifecycle orchestration and provider-specific EC2 and MicroVM checks, including scale-up, pool, and scale-down coverage.
- Updates the example documentation, Terraform/OpenTofu lockfiles, and MiniStack workflow to use the unified smoke entrypoint.

## Test Plan

- Lifecycle-hook unit tests for payload validation, lifecycle handling, process management, server behavior, storage, and timing.
- Lambda service formatting, linting, test, build, packaging, and artifact inspection.
- Terraform formatting, validation, TFLint, conflict checks, and `git diff --check`.
- Terraform and OpenTofu `init -lockfile=readonly` for the unified `multi-runner-webhook` example.
- No live AWS deployment was performed.

## Related Issues

- Stacked on #5255.
- Foundation and Terraform provider layers: #5413 and #5377.


.github/workflows/ministack.yml
```diff
@@ -143,7 +141,7 @@
     timeout-minutes: 30
     services:
       ministack:
-        image: ghcr.io/ministackorg/ministack:1.5.10@sha256:706b2b83c6be7e4f4dbb6a0dc28ffdebb500c6c80b64cf7938f45040fb2158e8
+        image: ghcr.io/ministackorg/ministack-preview-build:pr-1790-f3a550cf
```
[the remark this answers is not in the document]
edersonbrilhante: It is temporary until ministack has a new release

lambdas/services/microvm-lifecycle-hooks/src/payload.ts
```diff
@@ -0,0 +1,101 @@
...
+  if (typeof request.microvmId !== 'string' || !MICROVM_ID_PATTERN.test(request.microvmId)) {
+    throw new HookRequestError('microvmId is missing or invalid');
+  }
+  if (typeof request.runHookPayload !== 'string') {
+    throw new HookRequestError('runHookPayload must be a JSON string');
+  }
+
+  const payload = parseObject<VersionedRunPayload>(request.runHookPayload, 'runHookPayload must contain valid JSON');
+  if (payload.version === 1) {
+    if (!hasExactKeys(payload, ['version', 'runnerConfigSsmPath'])) {
```
guicaulada: I think we should align this parser with the payload produced by #5255. I tried passing the actual `createMicrovmRunHookPayload` output into `parseRunRequest`, and it was rejected both with and without a pinned image version: the producer always includes `runnerTokenSsmPath`, and can also include `imageArn` and `imageVersion`.

There also seems to be a path distinction we should preserve: the producer uses `runnerConfigSsmPath` for bootstrap settings and `runnerTokenSsmPath` for the JIT token, whereas this maps the former to `SSM_TOKEN_PATH`. Maybe a test that feeds the producer output directly into this parser would help keep the two contracts aligned. Is there a translation step I'm missing?

edersonbrilhante: Fixed. SSM_TOKEN_PATH is used by the lambdas, and it passes via payload to microvm

lambdas/services/microvm-lifecycle-hooks/src/processes.ts
```diff
@@ -0,0 +1,274 @@
...
+  ) {
+    if (!isAbsolute(this.runnerRoot)) {
+      throw new Error('RUNNER_ROOT must be an absolute path');
+    }
+  }
+
+  public launch(bootstrap: RunnerBootstrap, microvmId: string): ManagedProcess {
+    const runner = join(this.runnerRoot, 'run.sh');
+    const arguments_ = ['--jitconfig', bootstrap.jitConfig];
+    const child = spawn(runner, arguments_, {
```
guicaulada: The direct runner launch looks simpler, but I think we may need to keep the internal-service startup step as well. The supplied image creates `/etc/services.d/cloudwatch-agent/down`, and `start-services.sh` is the only place that enables that service with `s6-svc -u`. I could not find a caller of that script now that we spawn `run.sh` directly and filter out `INTERNAL_SERVICES`. With CloudWatch enabled, the runner can start while its diagnostic logs never reach the configured log groups. Maybe we should start the configured services with the MicroVM ID and bootstrap config path before handing off to the runner, and cover that path in an image-level test. Am I missing another startup hook that does this?

Issue #4710, opened 2025-08-13
Looking up number of active runners can be slow and cause API rate limit


npalm:
## Problem

For a setup supporting a large amount of runners, about 20K+ (a day) the method the scale up function is suing to decide the nmaximum number of runner is reached can be slow.

The current implementation is looking for active and pending ec2 istnaces with relevant tags using the DescribeIntance API. For large enivorments we see that this method can become slow and even cause RateLimits on the EC2 API.

https://github.com/github-aws-runners/terraform-aws-github-runner/blob/6d3af4ca4b3e7efb4f4384b728f4e8b253aa8f2e/lambdas/functions/control-plane/src/aws/runners.ts#L65-L79

This method was chosen back in the days to keep the complexity low. And avoid managing any state. But negative side effect could be result in a lambda requiring over 15 seconds more to just see if the limit is reached. And also can casue a API rate limit.

## Solutions

We looking for a slotion that keeps the complexity low but have not the negative side effects as mention above. Here a first short with ideas:

- Cache the result in the lambda for a certain time frame, this will avoid that under high load the call will be made frequent. Drawback.. The module can scale beyond the set maximum
- Try to keep track on the number of instances by listining to events for creation and termination. But this requires a complex setup.
- ...

PR #5281, opened 2026-08-19
feat: event-driven runner count cache to reduce DescribeInstances

yalafe:
## Description

Reduces the `DescribeInstances` load in the scale-up path (#4710) with an opt-in, event-driven runner count cache. Builds on the approach in #4983 by s1v4-d@, reworked onto current `main` (plugin layout) with a few additions. Opt-in via `runner_count_cache = { enable = true }`; default is unchanged behaviour.

**Approach** : 
Why: scale-up currently counts active + pending runners via `DescribeInstances` on every decision, which gets slow and rate-limited at scale. We built a fix, then found it lines up with #4710 (npalm@) and #4983 (s1v4-d@) - so bringing it
here rather than maintaining it separately.

What: replace that per-decision call with a DynamoDB counter kept up to date by an EventBridge EC2 state-change Lambda, a short in-memory TTL cache in scale-up, and a `DescribeInstances` fallback when the counter is cold/stale - combining both mitigations from #4710.

**What this adds on top of #4983:**
- **Idempotent counter** - per-instance marker (`INSTANCE#<id>`, COUNTED/TERMINATED) written with the counter update in one `TransactWriteItems`; `+1` once, `-1` only when counted, can't go negative. Lets it count on the first active event (`pending` or `running`) - matching the original "active and pending" count and avoiding boot-window over-provisioning.
- **DLQ + retry + alarm** on the EventBridge target - no silent event loss.
- **Provider-agnostic read path** - a decorator around the `getCurrentRunners` contract (no EC2 specifics). The counter Lambda is the only EC2-aware piece.

**Consistency trade-off (upfront):** the counter is eventually consistent (updated a few seconds after each state change via the event path), so it can briefly under-report just-launched runners - the "can scale beyond max" drawback noted in #4710. Bounded by in-process accounting within a scale-up invocation and the staleness fallback to `DescribeInstances`. The marker leaves room for a synchronous `+1` at launch as a follow-up if needed.

## Test Plan

- Unit tests, all passing:
- runner-count-cache Lambda - 11 tests: +1/-1 transitions, duplicate `running`/`terminated` are no-ops (idempotency), out-of-order `terminated` doesn't underflow, env-filter/tag guards, and non-cancellation errors rethrow for retry.
runner count cache classes (in-memory + DynamoDB) - 21 tests.
- getCurrentRunners decorator - 5 tests: in-memory hit, fresh DDB counter, stale/miss → provider fallback, reset-after-createRunners.
- Existing control-plane scale-up suite still green (113 tests) - decorator is a transparent pass-through when the feature is disabled.
- ESLint + Prettier pass on the new lambda code.
- `terraform validate` clean; `terraform plan` shows only the expected new resources when `runner_count_cache.enable = true`, and no changes when disabled (opt-in).
- Deployed to a test account and triggered live workflow runs: the counter tracked the fleet from 0 up to ~100 runners and back to 0, with `DescribeInstances` kept off the scale-up hot path and the DLQ staying empty. Note: single-repo test setup, so this is functional evidence rather than a 20K/day-scale benchmark.

## Related Issues

Addresses #4710. Builds on the approach in #4983 . Read path is written to align with the storage/plugin-layout refactor in #5277.


edersonbrilhante:
I liked the idea, but this should be added in compute provider ec2. We want to keep the controle plane neutral provider, so we can add new compute provider.

yalafe:
> I liked the idea, but this should be added in compute provider ec2. We want to keep the controle plane neutral provider, so we can add new compute provider.

Done, thanks. Moved the count-cache read into the EC2 provider's getCurrentRunners (in-memory → DynamoDB counter → DescribeInstances fallback) and removed the control-plane decorator, so the control plane is provider-neutral. Opt-in unchanged; tests updated.

edersonbrilhante:
@yalafe we are working to support multi computer provider(ec2, microvm, etc), multi storage provider(ssm, dynamodb, etc) and orchestration provider(GH webhook, GH scaleset) . Your code is a great idea, we need to adapt to be compatible with multi storage provider

Reference:
https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Aopen+is%3Apr+author%3Aedersonbrilhante+label%3Aexperimental-v2

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Aorchestration-provider%3Ascale-set

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Astorage-provider%3Aaws_dynamodb

https://github.com/github-aws-runners/terraform-aws-github-runner/pulls?q=is%3Apr+is%3Aopen+label%3Acompute-provider%3Aaws_microvm

PR #5282, opened 2026-08-19
feat(github-app): use ssm manifest for extra apps

guicaulada:
## Description

Follow-up to #5269, as promised in the [#5038 review](https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5038#pullrequestreview-4855605507) (items 1, 3, 4, and 5). We run this design in production.

- **Deliver additional GitHub App credentials via an SSM manifest parameter** instead of colon-joined parameter names in the lambda environment. Lambda's total environment limit is 4 KB, which the colon-joined lists hit at roughly 15–20 apps with typical parameter paths; the manifest keeps the environment constant regardless of app count (Intelligent-Tiering covers manifests beyond the 4 KB standard parameter tier).
- **Removes the positional list alignment** between id/key/installation-id parameter names, which could silently shift installation ids across apps if the lists drifted.
- **Rate-limit metric reads app ids from already-loaded credentials** (`getLoadedAppId`) instead of re-reading SSM per app index; out-of-range indexes no longer throw inside the swallowed catch.
- **Documents that additional apps must be installed on the same orgs/repos** as the primary app, in a new "Distributing load across multiple GitHub Apps" section of the rate-limits guide.

Internal contract only: Terraform and the lambdas deploy together, so the env transport change is invisible to module users. `additional_github_apps` is unchanged.

## Test Plan

- control-plane: 346 tests passed, including rewritten manifest-based credential loading tests.
- `terraform fmt`/`validate` clean on root, runners, multi-runner, ssm; `terraform test` in `modules/runners` passes.
- ESLint + Prettier clean; READMEs regenerated with terraform-docs.

## Related Issues

Follow-up to #5269 / #5038.


Brend-Smits (dismissed):
Reviewed and dev-deployed on a sandbox multi-runner stack. The env→manifest transport works end-to-end: scale-up/scale-down load credentials via the new gh-auth path with no errors, a runner scaled up and self-terminated cleanly. 

The branch is behind main and needs a rebase, the only conflicts are in `modules/multi-runner/{main.tf,termination-watcher.tf}` from the `effective_config` refactor (#5342), resolved by keeping the effective-config wiring and swapping `github_app_parameters.id[0]/key_base64[0]` → the new scalar shape. 

LGTM once rebased.

Brend-Smits (approved):

PR #5284, opened 2026-08-19
feat(multi-runner): add experimental config translation

edersonbrilhante:
## Description

- Introduce one `multi_runner_config` input that accepts either the stable v1 `runner_config` shape or the experimental v2 provider-boundary shape.
- Translate v1 lanes into the canonical v2 model and resolve global, provider, and per-lane settings without losing plan-known map keys.
- Add typed global configuration inputs and documentation for GitHub, Lambda, orchestration, SSM, observability, and compute-provider settings.
- Route resource-ready configuration through effective/resolved locals while preserving v1 compatibility and adding v1/v2 translation, resolution, and effective-config tests.

## Test Plan

- `terraform fmt -recursive`
- `terraform validate` in `modules/multi-runner`
- `terraform test` in `modules/multi-runner` — 9 passed, 0 failed
- Pre-commit formatting, validation/TFLint, and merge-conflict checks


Brend-Smits (commented):
Really like the direction here, and the ADR is excellent. One design worry: baking `experimental_` into every variable name means anyone who helps test this, or adopts it once a migration path exists, has to rename their whole config again at graduation. That penalizes exactly the early adopters we want.

Could we instead treat instability as a lifecycle property, not a naming one?

- Give the config its intended permanent names now (`global_config`, `global_config_observability`, `multi_runner_config`, …).
- Gate opt-in behind a single acknowledgement flag, e.g. `experimental_features = ["multi-runner-v2"]`, with a validation that errors if the v2 config is set without the ack.
- Communicate "this may break" via the docs banner, description warnings, and semver on the schema during the experimental window.

Then graduation is just dropping the flag (kept as a deprecated no-op for one release), the user's actual config blocks never change. It also fits the ADR's goal of a stable provider boundary: the boundary shouldn't be spelled `experimental_` in every consumer's HCL.

Fair trade-off: the prefix does make instability impossible to miss and keeps v1/v2 unambiguous in locals/state, the gate approach reproduces that via validation instead of naming. I think that's worth it to avoid the double migration. 

What do you think?

edersonbrilhante:
@Brend-Smits Done. Can you re-review

Brend-Smits (approved):
Changes look good, thank you for acting on my feedback.
Please note that there's nothing preventing people from trying to use experimental v2 options when using v1 multi runner. Yes the options might not get picked up, but there's no mixing check that errors when v2 attributes get used when v1 is active or vice versa.
This is not a show stopper now, but definitely something we should consider adding in the future.

Great work though, happy to see this!

PR #5293, opened 2026-08-24
test(terraform): apply examples directly against MiniStack

edersonbrilhante:
## Description

- Add a MiniStack lifecycle workflow that runs literal `terraform init`, `terraform apply`, and `terraform destroy` for the `base`, `prebuilt`, and `termination-watcher` examples.
- Keep the checked-in examples unchanged and pass MiniStack-only values through test-specific tfvars files.
- Create only test-local Lambda ZIP fixtures and, for `prebuilt`, a synthetic AMI through MiniStack's AWS-compatible EC2 API; cleanup removes only fixtures created by the test.
- Prevent real AWS or GitHub changes with a local endpoint, synthetic credentials, and test-only GitHub App values.
- The `default`, `ephemeral`, and `multi-runner` examples are intentionally outside this workflow for now. Their AMI filters depend on public or foreign-account AMI owners, while MiniStack v1.5.0 cannot describe locally registered images as `amazon` or another account owner.
- A follow-up PR will either consume MiniStack support for cross-account/public AMI description or provide mocked AMI metadata for MiniStack while retaining real AMI discovery for Terraform applies against AWS.

## Test Plan

- MiniStack lifecycle checks for `base`, `prebuilt`, and `termination-watcher`.
- Verified `terraform fmt -recursive -check -diff`.
- Verified `shellcheck tests/ministack/run-example.sh`.
- Verified `sh -n tests/ministack/run-example.sh`.
- Verified the workflow and tfvars-only runner use only the local MiniStack endpoint and test credentials.

## Related Issues

None.


Brend-Smits:
Really like the direction here! Running the examples through a full `init` / `apply` / `destroy` against a local AWS emulator is a solid step up from `validate`/`plan` for catching wiring and lifecycle regressions in CI. I tried the workflow locally (MiniStack 1.5.0 via Docker) and `base`, `prebuilt`, and `termination-watcher` all applied and destroyed cleanly.

Since this introduces a new tool and a new testing approach into the framework, would you consider adding a short ADR (e.g. under `docs/adr/`) to capture the decision? It would help future contributors understand:

- **Why MiniStack** — the problem it solves and why it was chosen over alternatives (LocalStack, real-AWS integration tests, etc.).
- **Goal / scope** — what this layer is meant to validate (Terraform resource shape, provider schema, apply/destroy lifecycle) versus what it explicitly does *not* cover (real IAM enforcement, instance boot, GitHub webhook flow).
- **Out of bounds** — the current limitations, e.g. why `default` / `ephemeral` / `multi-runner` are excluded (AMI owner/description constraints in MiniStack 1.5.0) and any other known emulator gaps.
- **Roadmap** — how we plan to grow this to improve the framework's overall testing posture (the follow-up for cross-account/public AMI support, broader example coverage, etc.).

Not a blocker for this PR, I just think an ADR would make the intent durable as the approach expands.

edersonbrilhante:
@Brend-Smits Since I drafted this PR, ministack added support to cross account AMI. I updated the PR to use the new version and support  `default` / `ephemeral` / `multi-runner` as well.

Brend-Smits (approved):
Lovely, I tested it locally and it's working as advertised. Good work.

PR #5295, opened 2026-08-24
refactor(storage): extract SSM provider capabilities

edersonbrilhante:
## Description

- add a storage-provider runtime package with separate `RunnerBootstrapStore` and `RunnerGroupCacheStore` capability contracts
- keep `aws_ssm` as the only implementation while preserving existing parameter paths, `SecureString`/`String` behavior, tags, throughput pacing, IAM, KMS behavior, and Terraform resource addresses
- select SSM independently for bootstrap handoff and runner-group caching in stable and experimental scale-up/pool Lambdas
- rename compute-provider metadata hooks so shared contracts no longer expose SSM-specific tag names
- document the usage-based storage boundary and explicitly defer credentials, persistent configuration, AMI resolution, runner-side reads, and DynamoDB

## Test Plan

- `yarn vitest run --config libs/storage-providers/vitest.config.ts` — 49 files / 819 tests passed
- TypeScript compile checks passed for `storage-providers` and `control-plane`
- ESLint and Prettier checks passed for the changed TypeScript sources
- the control-plane Lambda bundled successfully with `ncc`
- `tofu test -filter=tests/pool.tftest.hcl` passed for the stable runners module
- the changed primary scenarios passed in the webhook scale-runners and pool Terraform tests; later pre-existing override scenarios remain blocked locally by OpenTofu resolving `var.config` as unavailable inside their run-level variable blocks
- staged pre-commit checks passed: Terraform fmt, Terraform validate with TFLint, and merge-conflict detection

## Related Issues

None.


lambdas/libs/storage-providers/core/index.ts
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Short-lived, sensitive handoff from orchestration to one runner.
+ *
+ * Reading and deletion happen in the runner bootstrap implementation. This
+ * writer deliberately exposes no generic key/value operations.
+ */
+export interface RunnerBootstrapStore {
+  readonly provider: StorageProviderType;
+  readonly maxWritesPerSecond?: number;
+  put(record: RunnerBootstrapRecord, options?: RunnerBootstrapWriteOptions): Promise<void>;
```
guicaulada, line 31: This introduces `RunnerBootstrapStore.put`, while the other stack introduces `RunnerConfigStore.create`. Please align this with the chosen storage interface, or identify it as a replacement for that stack. Merging both would leave overlapping abstractions for the same operation.

guicaulada (commented):
LGTM! Just think we need to align the interface on `put` vs. `create`.

PR #5296, opened 2026-08-24
feat: Switch to on-demand for runners 

dgignac-tii:
## Description
To address Serverless gitaction verify failing with Error.   I'm trying on-demand instances rather than spot 

I will coordinate with team for a good time to test the switch over.


Brend-Smits:
We do not want to switch to on-demand runners for the open source framework.

PR #5299, opened 2026-08-26
feat(scale-set): add scale set orchestration in Terraform

edersonbrilhante:
## Description

Adds Terraform support for experimental GitHub Actions runner scale-set orchestration in the multi-runner v2 interface.

This PR includes:

- A standalone `modules/orchestration-providers/scale-set` module that deploys one ECS Fargate controller service per resolved controller group, including the ECS cluster/service/task definition, private networking, security groups, CloudWatch logging, health checks, deployment rollback, and hardened task settings.
- Plan-known grouping by compute-provider type (the default), runner configuration, or explicit custom membership. Grouping keeps each GitHub scale set and message session independent while packing multiple reconcilers into one controller task.
- Versioned non-secret reconciler configuration delivery through SSM Parameter Store and the controller manifest. GitHub App credentials remain SSM parameter references, with task and compute-role permissions restricted to the configured resources.
- Validation for GitHub scope and scale-set name ownership, grouping coverage, plan-time `for_each` inputs, provider contracts, configuration and task-definition size limits, reserved environment variables, wildcard IAM actions, and AWS inline-policy quotas.
- A compute-provider capability boundary that carries provider-owned runtime configuration, environment variables, and IAM statements. The EC2 adapter supplies scale-set launch, tagging, termination, JIT-configuration storage, AMI, and related permissions with ownership conditions.
- Multi-runner and runner-config wiring, the `examples/multi-runner-scale-set` example, MiniStack fixtures, and CI updates for the new Terraform path.

The module adopts scale sets that exist in GitHub by name; it does not create or delete GitHub scale-set resources. The TypeScript controller that consumes this Terraform contract is implemented in #5300.

## Test Plan

- Added focused Terraform tests for the scale-set module, computed-input fixtures, grouping, ownership validation, configuration delivery, IAM policy construction, and quota checks.
- Added the multi-runner scale-set example and MiniStack variables/runner coverage.
- Terraform formatting, validation/TFLint, and the repository CI workflows cover the updated modules and example paths.

## Related Issues

- Builds on the multi-runner v2 interface from #5367.
- Provides the Terraform foundation consumed by #5300.


modules/orchestration-providers/scale-set/locals.tf
```diff
@@ -0,0 +1,254 @@
+locals {
+  github_config_urls = {
+    for runner_name, runner_config in var.runner_configs : runner_name => format(
+      "%s%s",
+      trimsuffix(coalesce(runner_config.github.enterprise_server.url, "https://github.com"), "/"),
+      runner_config.github.runner_owner == null ? "" : "/${runner_config.github.runner_owner}",
```
guicaulada, line 6: With the accepted enterprise defaults (`runner_owner = null` and no `enterprise_server.url`), this generates `https://github.com`. The controller rejects that URL before startup because it requires an organization, repository, or enterprise path. I reproduced this with the manifest parser.

We should let callers provide their enterprise slug or full registration URL, and require it when enterprise registration is selected. That would let the module generate a scoped URL without assuming a default enterprise. We should also keep hosted GitHub routing separate from the GHES override; putting a GitHub.com enterprise URL in `enterprise_server.url` also sets `forceGhes = true`.

modules/orchestration-providers/scale-set/locals.tf
```diff
@@ -0,0 +1,254 @@
...
+      ]
+    } :
+    {}
+  )
+
+  group_resource_names = {
+    for group_name in keys(local.controller_groups) : group_name => format(
+      "%s-ss-%s-%s",
+      var.prefix,
+      substr(replace(lower(group_name), "/[^a-z0-9_-]/", "-"), 0, 20),
```
guicaulada: An allowed 20-character prefix and 20-character group name produce a 70-character compute-role name after `-compute-<8-character hash>` is appended. The AWS provider rejects the plan because IAM role names cannot exceed 64 characters; a boundary test reproduced this.

We could limit the readable group portion to 14 characters to keep the longest role at 64 while retaining the hash. Both the role resource and its generated ARN use this shared name.

```suggestion
      substr(replace(lower(group_name), "/[^a-z0-9_-]/", "-"), 0, 14),
```

guicaulada (commented):
Nice work on this! The overall approach looks good to me, but we should address the enterprise scope configuration and IAM role-name limits before merging. I’ve added details inline.

modules/orchestration-providers/scale-set/locals.tf
edersonbrilhante: Lets remove the enterprise support then. We can add back in this https://github.com/github-aws-runners/terraform-aws-github-runner/pull/5096

So the implementation can work in scaleset and webhook.

guicaulada (dismissed):

guicaulada (dismissed):

Brend-Smits (dismissed):

edersonbrilhante:
@npalm when the time permits, please review this PR. 
It was multi-PRs merged into 1 final PR. 
Also I added integration tests to give more confidence that is working.

npalm:
@edersonbrilhante will check the PR tonight, can you already update the conflict (readme)

edersonbrilhante:
@npalm Done

edersonbrilhante:
@npalm I created the issue for to update the docs in a follow up PR https://github.com/github-aws-runners/terraform-aws-github-runner/issues/5470

modules/orchestration-providers/scale-set/iam.tf
```diff
@@ -0,0 +1,217 @@
...
+    actions = [
+      "ssm:GetParameter",
+      "ssm:GetParameters",
+    ]
+    resources = [for parameter in local.group_github_parameters[each.key] : parameter.arn]
+  }
+
+  statement {
+    sid       = "AssumeComputeProviderRoles"
+    effect    = "Allow"
```
npalm, line 71: The controller writes two IDs back to SSM when they are discovered dynamically: the GitHub App installation ID in `credentials.ts:189` and the runner group ID in `reconciler.ts:261`. Both use `parameterStore.put?.(...)` and `defaultParameterStore` in `parameter-store.ts` defines `put`, so the calls run at runtime. The task role here grants only `ssm:GetParametersByPath`, `ssm:GetParameter`, and `ssm:GetParameters`, so the first cache write will fail with AccessDenied.

Two ways forward, either works:
1. Add a scoped `ssm:PutParameter` statement targeting only the installation ID and runner group ID parameter ARNs.
2. Remove `put` from `defaultParameterStore` and drop the write-back path if caching is not required in this iteration.

Option 1 keeps the no-restart-on-rotation benefit; option 2 keeps IAM smaller. Happy to go either way.

modules/orchestration-providers/scale-set/locals.tf
```diff
@@ -0,0 +1,254 @@
...
+  group_resource_names = {
+    for group_name in keys(local.controller_groups) : group_name => format(
+      "%s-ss-%s-%s",
+      var.prefix,
+      substr(replace(lower(group_name), "/[^a-z0-9_-]/", "-"), 0, 14),
+      substr(sha256(group_name), 0, 8),
+    )
+  }
+
+  official_container_image = "ghcr.io/github-aws-runners/terraform-aws-github-runner-scale-set-service:latest"
```
npalm, line 43: Nice job enabling `versionConsistency` on the container definition, that already stops mid-deploy tag drift. To also give plan-time protection to users who take the default, consider replacing `:latest` here with a released digest, or accept `null` and add a validation that fails when both `var.container.image` is null and no default digest is compiled in. The README already tells production callers to pin a digest, so this closes the gap for the convenience default too.

modules/orchestration-providers/scale-set/iam.tf
```diff
@@ -0,0 +1,217 @@
...
+    ]
+    resources = ["*"]
+  }
+
+  statement {
+    # ECR does not support resource-level permissions for authorization tokens.
+    sid       = "AuthorizePrivateEcrPull"
+    effect    = "Allow"
+    actions   = ["ecr:GetAuthorizationToken"]
+    resources = ["*"]
```
npalm, line 207: The `execution` role always grants `ecr:BatchCheckLayerAvailability`, `ecr:BatchGetImage`, `ecr:GetDownloadUrlForLayer` on `*`, and the resource-unscoped `ecr:GetAuthorizationToken`. The README already notes that public registry images need no pull permission, so callers using the GHCR default get permissions they never use.

A small `dynamic` block gated on something like `local.uses_private_ecr = startswith(local.resolved_container_image, "${account}.dkr.ecr.")` would give the default deployment a smaller IAM surface, while private ECR users still get exactly what they need.

modules/orchestration-providers/scale-set/variables.tf
```diff
@@ -0,0 +1,200 @@
...
+      ipv6_cidrs = optional(set(string), [])
+    }), {})
+  })
+  nullable = false
+}
+
+variable "logging" {
+  description = "CloudWatch Logs configuration. CloudWatch encrypts logs at rest with an AWS-owned key by default; set `kms_key_arn` to use a customer-managed key."
+  type = object({
+    retention_in_days = optional(number, 30)
```
npalm, line 186: KMS handling matches the rest of the repo cleanly: `null` default (AWS-owned key) with an opt-in customer-managed key. The retention default is out of step though. Lambda modules across the repo default to 180 days (see root `variables.tf:418`, `modules/lambda/variables.tf:41`, `modules/webhook/direct/variables.tf:19`, `modules/webhook/eventbridge/variables.tf:19`), and this controller defaults to 30. Since the controller runs continuously and its logs are the main forensic trail when a reconciler misbehaves, aligning on 180 by default (or reading the top-level `logging_retention_in_days` when wired through `multi-runner`) gives operators a consistent baseline and avoids losing history when they had not thought to override.

Minor detail: the rest of the repo uses `kms_key_id`; here it is `kms_key_arn`. Both work on `aws_cloudwatch_log_group`, and the ARN form is arguably clearer, so this is just a naming heads-up, not a change request.

lambdas/services/scale-set/src/reconciler.ts
```diff
@@ -0,0 +1,743 @@
...
+}
+
+function errorLogAttributes(error: unknown, depth = 0): Record<string, unknown> {
+  if (!(error instanceof Error)) return { message: String(error) };
+  if (depth >= 3) return { name: error.name, message: error.message, cause: '[TRUNCATED]' };
+
+  const errorWithMetadata = error as Error & { code?: unknown; status?: unknown; cause?: unknown };
+  return {
+    name: error.name,
+    message: error.message,
```
npalm, line 738: Defense in depth suggestion, not fixing an active leak. `errorLogAttributes` flattens Errors into plain objects before they reach the logger. The logger's `sanitize()` path already has an `Error`-specific branch that runs `sanitizeErrorMessage` over `error.message` and preserves `status` and `code`. By pre-flattening, we skip that branch and the redaction regex never fires.

Two clean options:
1. Pass the raw `Error` at the call sites (`error: error`) and let `sanitizeLogAttributes` handle it, keeping the extra `httpErrorLogAttributes` spread for HTTP status.
2. Route `message` through the same redaction regex inside `errorLogAttributes`.

No exploitable case today since the endpoints that return secrets only do so on 2xx, but restoring the second layer is cheap and matches the intent of the logger.

lambdas/services/scale-set/src/main.ts
```diff
@@ -0,0 +1,144 @@
...
+    groupName: manifest.groupName,
+    revision: manifest.revision,
+    reconcilerCount: manifest.reconcilers.length,
+    runnerConfigNames: manifest.reconcilers.map(({ runnerConfigName }) => runnerConfigName),
+  });
+  logger.debug('scale_set_controller_reconcilers_loaded', {
+    groupName: manifest.groupName,
+    reconcilerCount: manifest.reconcilers.length,
+    runnerConfigNames: manifest.reconcilers.map(({ runnerConfigName }) => runnerConfigName),
+  });
```
npalm, line 32: The `info` log at line 22 and the `debug` log at line 27 emit the same `reconcilerCount` and `runnerConfigNames`. Consider dropping the debug duplicate, or moving the list to debug-only and keeping only the count on info. Cheap win for log volume when debug is enabled.

modules/orchestration-providers/scale-set/variables.tf
```diff
@@ -0,0 +1,200 @@
...
+
+variable "network" {
+  description = <<-EOT
+    Private Fargate networking. Tasks never receive public IP addresses and the managed security groups have no ingress. HTTPS egress defaults to IPv4 Internet access because GitHub endpoints cannot be represented as security-group destinations; route it through controlled NAT, firewall, or proxy infrastructure when required.
+  EOT
+  type = object({
+    vpc_id     = string
+    subnet_ids = set(string)
+    https_egress = optional(object({
+      ipv4_cidrs = optional(set(string), ["0.0.0.0/0"])
```
npalm, line 176: The default of `["0.0.0.0/0"]` is convenient but not the only option. GitHub publishes its egress ranges at `https://api.github.com/meta` (`actions`, `api`, `web`), and adopters can pin them via `data "http"`, an out-of-band refresher, or NAT/proxy allowlists. Consider reframing the variable description and the README from "GitHub endpoints cannot be expressed as security-group destinations" to something like:

> Defaults to full IPv4 Internet for convenience. GitHub publishes its outbound ranges at `https://api.github.com/meta` and adopters should restrict egress to those ranges, a NAT gateway, a firewall, or a proxy when their security posture requires it. Adopters need to make this decision consciously; the default trades reachability for a wider egress surface.

That way the security trade-off is explicit and adopters know the tighter option exists without digging.

.github/workflows/lambda.yml
```diff
@@ -32,21 +33,69 @@ jobs:
...
           name: coverage-reports
           path: ./**/coverage
           retention-days: 5
+
+  scale-set-container:
+    name: Build scale-set service container
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
```
npalm, line 66: Small consistency nudge: the new job here pins `step-security/harden-runner` to `9af89fc7...` (v2.19.4), while the new job in `.github/workflows/smoke-tests.yml` line 106 uses `e14015d5...` (v2.21.1). Same repo, two versions in the same PR. Bumping this one to `e14015d5...` keeps the pinned SHA set easy to audit and lines up with Dependabot's next round.

.github/workflows/smoke-tests.yml
```diff
@@ -83,3 +83,78 @@ jobs:
...
+  scale_set_integration_smoke:
+    name: Run scale-set ECS smoke test against MiniStack and MockServer
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    services:
+      ministack:
+        image: ghcr.io/ministackorg/ministack:1.5.12@sha256:41fe1ce2e666c6cc410c6047a9db8bf1df69cd0028ebc0a6c6e5517c3a83d6e0
+        ports:
+          - 4566:4566
+        options: >-
```
npalm, line 96: Minor, awareness only, no change requested.

Mounting `/var/run/docker.sock` into the `ministack` service gives that container root-equivalent access to the runner's Docker daemon. That is the standard LocalStack pattern and safe here because the workflow does not trigger from forks and the ministack image is digest-pinned, so nothing untrusted rides along. Consider a short in-workflow comment above the mount so future readers see the risk envelope without having to reconstruct it.

npalm (dismissed):
Solid foundation. The security posture stands out: container hardening in `task.tf`, tag-conditioned EC2 IAM in the compute provider, credential values kept out of Terraform entirely (only SSM parameter names threaded through), and workflow-level supply chain hygiene at release (`sbom: true`, `provenance: mode=max`, `actions/attest`, SHA-pinned actions, `harden-runner`). Nice work.

**One thing to fix before merge**

The task role in `modules/orchestration-providers/scale-set/iam.tf` grants only SSM read actions, but the controller writes discovered IDs back to SSM at `credentials.ts:189` (installation ID) and `reconciler.ts:261` (runner group ID) through `parameterStore.put`. First cache write will fail with AccessDenied at runtime. Either add a scoped `ssm:PutParameter` on those two ARNs or drop the write-back path. Details in the inline comment.

**Docs missing from this PR**

This change introduces a new orchestration mode, a new deployment shape (ECS Fargate controller alongside the existing Lambdas), a new grouping model, and a new set of experimental variables. The only doc touched is `docs/security.md`. Please add:

- `docs/index.md` and `docs/configuration.md`: extend the architecture description and the configuration reference to cover the scale-set orchestration provider, the controller group model, and the new experimental variables.
- `docs/multi-runner-v1-to-v2-configuration.md` and `docs/multi-runner-v1-v2-migration.md`: add the scale-set option so v2 adopters can find it and understand when to pick it.
- A new ADR under `docs/adr/` capturing the "adopt existing scale sets by name, do not manage them" contract, alongside `docs/adr/002-runner-orchestration-provider-boundary.md`.
- `modules/orchestration-providers/scale-set/README.md` and `variables.tf`: make the security trade-off on the default `0.0.0.0/0` egress explicit (GitHub's outbound ranges at `https://api.github.com/meta` are pinnable), so adopters make the call consciously.

The remaining inline comments are suggestions and small nudges. Happy to iterate on any of them.

edersonbrilhante:
@npalm I created an issue to follow up your review. 
https://github.com/github-aws-runners/terraform-aws-github-runner/issues/5471

This PR is too big after the PRs got merged one after another. If you don't mind approving as compromise, I will handle this fix in a new PR, so the review is easier and cleaner

modules/orchestration-providers/scale-set/iam.tf
edersonbrilhante: Fixed by removing the controller SSM write-back path. `ParameterStore` now exposes only `get`, and the default implementation only reads parameters: [credentials.ts](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/lambdas/services/scale-set/src/credentials.ts#L14-L16) · [parameter-store.ts](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/lambdas/services/scale-set/src/parameter-store.ts#L21-L23)

modules/orchestration-providers/scale-set/locals.tf
edersonbrilhante: Fixed by removing the mutable `:latest` default. `container.image` is now required and validated before deployment: [locals.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/locals.tf#L43-L48) · [validations.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/validations.tf#L265-L274)

modules/orchestration-providers/scale-set/iam.tf
edersonbrilhante: Fixed by emitting ECR permissions only for private ECR images. Repository-layer permissions are scoped to the detected repository, while the authorization-token permission remains resource-unscoped as required by ECR: [locals.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/locals.tf#L43-L54) · [iam.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/iam.tf#L191-L216)

modules/orchestration-providers/scale-set/variables.tf
edersonbrilhante: Fixed by aligning the default log retention with the repository standard of 180 days. The public logging input is also consistently named `kms_key_id`: [variables.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/variables.tf#L183-L189) · [logging.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/logging.tf#L1-L7)

lambdas/services/scale-set/src/reconciler.ts
edersonbrilhante: Fixed by passing raw `Error` objects to the logger so the logger's Error-specific sanitization and redaction path remains active: [reconciler.ts](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/lambdas/services/scale-set/src/reconciler.ts#L141-L147) · [reconciler.ts](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/lambdas/services/scale-set/src/reconciler.ts#L502-L510)

lambdas/services/scale-set/src/main.ts
edersonbrilhante: Fixed by keeping the reconciler count in the info log and moving the detailed runner-config list to the debug log: [main.ts](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/lambdas/services/scale-set/src/main.ts#L22-L30)

modules/orchestration-providers/scale-set/variables.tf
edersonbrilhante: Fixed by documenting the `0.0.0.0/0` default as a reachability trade-off and pointing adopters to GitHub Meta ranges, NAT gateways, firewalls, and proxies: [variables.tf](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/variables.tf#L168-L180) · [README.md](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/modules/orchestration-providers/scale-set/README.md#L165-L172)

.github/workflows/lambda.yml
edersonbrilhante: Fixed by updating this workflow to the same pinned `step-security/harden-runner` v2.21.1 SHA used by the smoke-test workflow: [lambda.yml](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/.github/workflows/lambda.yml#L65-L69)

.github/workflows/smoke-tests.yml
edersonbrilhante: Added the requested awareness comment documenting the root-equivalent Docker socket access and the narrowly scoped risk envelope: [smoke-tests.yml](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/.github/workflows/smoke-tests.yml#L94-L100)

edersonbrilhante:
Implemented the documentation requested in this review:

- [architecture and scale-set lifecycle](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/index.md)
- [configuration and scale-set orchestration](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/multi-runner-v1-to-v2-configuration.md#L134-L147)
- [v1-to-v2 migration guidance](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/multi-runner-v1-v2-migration.md#L13-L29)
- [security boundaries and image provenance](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/security.md#L27-L60)
- [multi-runner scale-set example](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/examples/multi-runner-scale-set.md)
- [scale-set ownership ADR](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/fix/scale-set-review-followups/docs/adr/003-scale-set-resource-ownership.md#L23-L42)

guicaulada (approved):

PR #5300, opened 2026-08-26
feat(scale-set): add service controller

edersonbrilhante:
## Description

Adds the TypeScript implementation of the GitHub Actions runner scale-set control plane on top of the Terraform orchestration introduced by #5299.

This PR includes:

- A reusable GitHub Actions scale-set client that supports GitHub.com, GHES, and data-residency endpoints; GitHub App access tokens; runner-group and scale-set discovery/registration; JIT runner configuration; runner removal; and message-session polling, acknowledgement, job acquisition, refresh, and close.
- A long-running scale-set controller service for ECS. It loads versioned runner configuration from SSM Parameter Store, runs independent reconcilers for each runner configuration, exposes liveness/readiness endpoints, and performs bounded shutdown and session recovery.
- A scale-set compute-provider contract and EC2 implementation. The provider reconciles capacity from `totalAssignedJobs`, generates JIT configurations for launches, tracks provider-owned instances with tags, preserves busy or unknown runners during scale-down, and supports task-role or assumed-role credentials.
- Configuration, authentication, transport, retry, cancellation, ownership, lifecycle, health, and reconciliation safeguards. Sensitive tokens, message bodies, and JIT configurations are excluded from manifests and logs; disabling TLS verification is scoped to the relevant client instead of changing global Node.js state.
- Unit-test coverage for the scale-set client and HTTP/session behavior, service configuration and credentials, controller lifecycle and health, and EC2 provider configuration, inventory, reconciliation, scale-up, and scale-down.

The Terraform documentation and workflow changes are tracked separately in #5347.

## Test Plan

- Added focused TypeScript tests alongside the client, service, controller, and compute-provider implementations.
- The scale-set service package provides `typecheck`, `build`, `format-check`, `lint`, and `test` targets for CI validation.
- End-to-end ECS deployment is not part of this PR; the service consumes the scale-set configuration and IAM contracts supplied by the Terraform stack.

## Related Issues

- Stacked on #5299.
- Builds on the EC2 orchestration boundary from #5312.
- Follow-up documentation and workflow changes: #5347.


lambdas/services/scale-set/src/credentials.ts
```diff
@@ -0,0 +1,209 @@
...
+): Promise<number> {
+  const appRequest = request.defaults({ baseUrl: apiBaseUrl, request: { fetch: fetchImplementation } });
+  const appAuth = createAppAuth({
+    appId: credentials.appId,
+    privateKey: credentials.privateKey,
+    request: appRequest,
+  });
+  const appAuthentication = await appAuth({ type: 'app' });
+  for (let page = 1; page <= 100; page += 1) {
+    const url = new URL('/app/installations', `${apiBaseUrl}/`);
```
guicaulada: The leading slash replaces the base URL's pathname. With a GHES base such as `https://github.example.com/api/v3`, this requests `/app/installations` instead of `/api/v3/app/installations`, so discovery fails when no installation ID is supplied. We could use a relative path here to preserve `/api/v3` while keeping GitHub.com working.

```suggestion
    const url = new URL('app/installations', `${apiBaseUrl}/`);
```

lambdas/libs/github-actions-scale-set/src/http.ts
```diff
@@ -0,0 +1,327 @@
...
+  const aborted = new Promise<never>((_, reject) => {
+    onAttemptAbort = () => reject(abortReason(attemptController.signal));
+    attemptController.signal.addEventListener('abort', onAttemptAbort, { once: true });
+  });
+
+  try {
+    return await Promise.race([fetchImplementation(input, { ...init, signal: attemptController.signal }), aborted]);
+  } finally {
+    clearTimeout(timeout);
+    callerSignal?.removeEventListener('abort', forwardAbort);
```
guicaulada, line 235: `fetch` resolves when response headers arrive, so this cleanup runs before `executeRequest` finishes reading the body. If the body stalls, its `reader.read()` calls have neither the configured timeout nor caller-abort forwarding. I reproduced that aborting the caller after headers leaves the transport signal un-aborted.

We should keep the deadline and abort forwarding active until the body is consumed or cancelled, then clean them up. It would also help to add a test that returns headers followed by a stalled body and verifies both timeout and shutdown cancellation.

guicaulada (commented):
Thanks for the work on this! The overall approach looks good, but we should cover two edge cases around GHES installation discovery and HTTP cancellation. I’ve added details inline.

guicaulada (approved):

PR #5301, opened 2026-08-26
refactor(compute-providers): reuse EC2 runner clients

edersonbrilhante:
## Description

Depends on #5308.

Introduces an `Ec2RunnerClient` that owns a reusable EC2 client and exposes request-bound EC2 runner operations through `forRequest({ signal })`. The control-plane registry creates the client once for the warm Lambda execution environment, and the pool, scale-up, and scale-down capabilities share those operations while keeping cancellation scoped to each request.

The EC2 capability boundary now distinguishes resource operations from provisioning operations, uses consistent capability naming, and removes the legacy default-operation wrappers. Launch-template block-device lookup uses the same bound EC2 client, and runner creation is named separately from runner configuration.

## Test Plan

- Ran Prettier, ESLint, and the control-plane TypeScript type-check.
- Ran the complete compute-provider suite: 286 tests passed.
- Ran the complete control-plane suite: 345 tests passed.
- Added three runner-client tests covering launch-template lookup with the bound client and request signal, rejection of an already-cancelled request before AWS is called, and isolation between cancelled and active requests sharing one client.
- Updated pool, scale-up, and scale-down tests to exercise the returned capability methods and their dependency wiring.
- Ran `git diff --check`.

## Related Issues

None.

### DIFF
--- a/lambdas/libs/compute-providers/aws/ec2/control-plane.ts
+++ b/lambdas/libs/compute-providers/aws/ec2/control-plane.ts
@@ -4,23 +4,23 @@ import { EC2Client } from '@aws-sdk/client-ec2';
 
 import type { ControlPlaneProviderCapabilities, ControlPlaneProviderModule } from '../../contracts';
 import type {} from './src/environment';
-import { createEc2PoolProvider } from './src/control-plane/pool';
-import { createEc2ScaleDownProvider } from './src/control-plane/scale-down';
-import { createEc2ScaleUpProvider } from './src/control-plane/scale-up';
+import { createEc2PoolCapability } from './src/control-plane/pool';
+import { createEc2ScaleDownCapability } from './src/control-plane/scale-down';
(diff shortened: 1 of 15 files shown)


Brend-Smits (dismissed):
Tested it locally and all is good, working as expected.

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
stuartp44, line 950:
```suggestion
    it('non-retryable partial runner creation failures are not retried', async () => {
```

stuartp44 (approved):

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
edersonbrilhante: I will do in a follow up PR

Issue #2904, opened 2023-01-21
Error: InvalidFleetConfig The fleet configuration contains duplicate instance pools.


dgokcin:
Helloo

Trying to deploy the module to my infrastructure and having some problems with the scale-up lambda.

Any idea why I might have an invalid fleet config? Thanks!

```hcl
include {
  path = find_in_parent_folders()
}

locals {
  common_vars = yamldecode(file(find_in_parent_folders("common_vars.yaml")))
}

terraform {
  source = "../../..//modules/terraform-aws-github-runner"
}

inputs = {
  aws_region            = "eu-west-1"
  vpc_id                = dependency.vpc.outputs.vpc_id
  subnet_ids            = dependency.vpc.outputs.private_subnets
  lambda_s3_bucket      = "myorg-pg-gh-actions-lambdas"
  runners_lambda_s3_key = "runners.zip"
  syncer_lambda_s3_key  = "runner-binaries-syncer.zip"
  webhook_lambda_s3_key = "webhook.zip"

  prefix = "gh-ci"
  github_app = {
    key_base64     = "key"
    id             = "id"
    webhook_secret = "webhook_secret"
  }

  create_service_linked_role_spot = true

  enable_organization_runners = true
  runner_extra_labels         = "default,example"

  instance_types = ["m5.large", "c5.large"]

  # override delay of events in seconds
  delay_webhook_event   = 5
  runners_maximum_count = 1

  # set up a fifo queue to remain order
  enable_fifo_build_queue = true

  # override scaling down
  scale_down_schedule_expression = "cron(* * * * ? *)"
  # enable this flag to publish webhook events to workflow job queue
  # enable_workflow_job_events_queue  = true

  enable_user_data_debug_logging_runner = true

  tags = local.common_vars.tags
}

dependency "vpc" {
  config_path = "../vpc"
}
```

<img width="1122" alt="image" src="https://user-images.githubusercontent.com/33603535/213879941-0745e751-dfd7-4992-b071-af9e1ab339b3.png">

dgokcin:
update: this was the default configuration. I also tried deploying the module with multiple_runners. I believe the following code is failing and I have no idea what is going on. This is my first time working with spot instances so I really appreciate any help.

Thanks

```js
    let fleet;
    try {
        // see for spec https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateFleet.html
        fleet = await ec2
            .createFleet({
[...]

PR #5304, opened 2026-08-27
fix: support runner subnets in the same availability zone

### LINKED ISSUE #5297: Support multiple subnets in the same AZ
## Problem

We would like to use multiple subnets in the same AZ so that our runners can use the additional IP address capacity. Extending the existing subnet ranges is not an option in our environment. Requiring one subnet per AZ therefore prevents us from making use of the additional capacity.

At the moment, adding a second subnet in an AZ already represented in `subnet_ids` causes runner creation to fail. Terraform accepts and deploys the configuration successfully, but the scale-up Lambda fails later when it calls EC2 Fleet.

We saw the following error with the `multi-runner` module on v7.10.1:

```json
{
  "level": "ERROR",
  "message": "Error processing batch (size: 1): The fleet configuration contains duplicate instance pools., ignoring batch",
  "timestamp": "2026-08-06T14:50:36.440Z",
  "service": "Proj-ubuntu-2404-arm64-ephemeral-scale-up",
  "region": "eu-west-1",
  "environment": "Proj-ubuntu-2404-arm64-ephemeral",
  "module": "lambda.ts",
  "error": {
    "name": "InvalidFleetConfig",
    "message": "The fleet configuration contains duplicate instance pools.",
    "Code": "InvalidFleetConfig",
    "Error": {
      "Code": "InvalidFleetConfig",
      "Message": "The fleet configuration contains duplicate instance pools."
    }
  }
}
```

## Example

For example, consider three configured subnets:

```text
subnet-a1 -> eu-west-1a
subnet-a2 -> eu-west-1a
subnet-b1 -> eu-west-1b
```

Configuring all three subnet IDs causes the EC2 Fleet request to fail because two of the subnets are in `eu-west-1a`.

## Cause

The control-plane Lambda generates EC2 Fleet overrides for each combination of subnet and instance type:

```text
subnet-a1 x instance-type-1
subnet-a2 x instance-type-1
subnet-b1 x instance-type-1
```

EC2 Fleet treats a capacity pool as a combination of instance type and AZ. The first two overrides therefore represent the same capacity pool, even though they use different subnet IDs.

AWS rejects the complete `CreateFleet` request with `InvalidFleetConfig`, so no runner is launched.

## Expected behaviour

It should be possible to configure multiple subnets in the same AZ and have the module use their combined IP address capacity. A single EC2 Fleet request should contain at most one subnet from each AZ.

## Possible approach

The module could:

1. Use `DescribeSubnets` to determine the AZ of each configured subnet.
2. Group the subnets by AZ.
3. Create subnet sets containing no more than one subnet from each AZ.

This would require adding `ec2:DescribeSubnets` to the control-plane Lambda permissions.

## Related

This appears to be related to #2904. That issue describes the same `InvalidFleetConfig: The fleet configuration contains duplicate instance pools` error, but the documented workaround of requiring every subnet to be in a different AZ prevents additional same-AZ subnet capacity from being used.

Quindrix:
## Description

This PR allows runner subnets to exist in the same AZ. It groups the configured subnets into safe combinations and distributes runner capacity across separate fleet requests. If a request encounters a retryable capacity failure, the remaining capacity is carried forward to the next available subnet combination.

It also adds the required `ec2:DescribeSubnets` permissions and updates the documentation to remove the previous requirement that every subnet must be in a different AZ. Dedicated-host and explicit subnet override behaviour remains unchanged.

## Test Plan

- Unit tests were added for subnet grouping, capacity distribution, and retryable failures. All 71 EC2 runner tests pass, along with the ESLint and Prettier checks.
- Used these changes in our environment and were able to provide two subnets in the same AZ without issue.

## Related Issues

Fixes #5297, and follows on from #2904


edersonbrilhante:
@Quindrix we finished a refactor in the main. Now we can add your code fix. Can you fix the conflicts. 
Also add the new permission in https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/compute-providers/aws/ec2/control-plane.tf#L136 and https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/modules/compute-providers/aws/ec2/control-plane.tf#L26

For more context: https://github.com/github-aws-runners/terraform-aws-github-runner/blob/main/docs/adr/002-runner-orchestration-provider-boundary.md

PR #5306, opened 2026-08-27
refactor(compute-providers): share EC2 runner operations

edersonbrilhante:
## Description

Moves the shared EC2 runner implementation and types out of the control-plane directory and exposes them through the EC2 provider package boundary so multiple orchestration entry points can use the same runner operations.

Requires callers to provide a `RunnerSource`, preserving the existing `scale-up-lambda` and `pool-lambda` values. Both EC2 launch paths now use that caller-provided source for the `ghr:created_by` tag.

PR #5301 is the follow-up and will be stacked on this PR.

## Test Plan

- Ran ESLint for the compute-provider and control-plane projects.
- Ran the Lambda Vitest suite: 50 test files and 819 tests passed.
- Ran `git diff --check` against `refactor-ec2-provider-isolation`.

## Related Issues

None.

### DIFF
--- a/lambdas/functions/control-plane/src/pool/pool.test.ts
+++ b/lambdas/functions/control-plane/src/pool/pool.test.ts
@@ -3,7 +3,7 @@ import moment from 'moment-timezone';
 import * as nock from 'nock';
 
 import { createRunners } from '@aws-github-runner/compute-providers/aws/ec2/control-plane/runner-config';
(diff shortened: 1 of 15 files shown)


Brend-Smits (dismissed):
Looks good to me, tested it in our dev environment and confirmed the created_by value is correct.

stuartp44 (approved):
LGTM, nice work!

PR #5307, opened 2026-08-27
refactor(control-plane): decouple pool tests from EC2

edersonbrilhante:
## Description

Depends on #5306.

Refactors the control-plane pool tests to use the `PoolComputeProvider` capability boundary instead of importing and mocking the EC2 runner implementation.

EC2-specific assertions for running-instance filters, busy and booting runner counts, and the `pool-lambda` source now live in the EC2 pool provider tests. Generic pool orchestration coverage remains in the control-plane test.

## Test Plan

- Ran the targeted control-plane pool suite: 17 tests passed.
- Ran the targeted EC2 provider pool suite: 7 tests passed.
- Ran ESLint and Prettier checks for both changed files.
- Ran the control-plane TypeScript type-check.
- Ran `git diff --check` against `refactor-ec2-runner-operations`.
- Verified the generic control-plane pool test contains no EC2 references.

## Related Issues

None.

PR #5308, opened 2026-08-27
refactor(control-plane): remove EC2 assumptions

edersonbrilhante:
## Description

Depends on #5307.

Removes EC2-specific assumptions from generic control-plane tests and the local pool invocation. Scale-up and scale-down test doubles now model provider capability factories without owning a provider type, partial creation tests use provider-neutral runner terminology, and pool events rely on the optional provider type's default behavior.

The scale-up provider-selection test remains as integration coverage, but asserts the public `defaultComputeProvider` contract instead of the concrete EC2 value. The provider-type unit tests remain responsible for the concrete default mapping.

## Test Plan

- Ran the affected scale-up, scale-down, and Lambda wrapper suites: 253 tests passed.
- Ran the complete Lambda Vitest suite: 819 tests passed.
- Ran the control-plane TypeScript type-check.
- Ran ESLint and Prettier checks for all changed files.
- Ran `git diff --check`.
- Verified the four scoped files contain no EC2 references.

## Related Issues

None.

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
@@ -16,6 +16,7 @@ import type {
   CreateScaleUpRunnersInput,
   ScaleUpComputeProvider,
 } from './types';
+import { defaultComputeProvider } from '@aws-github-runner/compute-providers/provider-types';
 import { getParameter } from '@aws-github-runner/aws-ssm-util';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import type { Octokit } from '@octokit/rest';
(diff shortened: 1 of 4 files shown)


Brend-Smits (dismissed):
LGTM! All tests are green, no further comments.

stuartp44 (approved):

lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
```diff
@@ -947,7 +947,7 @@ describe('scaleUp with GHES', () => {
       expect(rejectedMessages).toEqual(['message-0']);
     });
 
-    it('does not retry partial EC2 instance creation failures that are not retryable', async () => {
+    it('does not retry non-retryable partial runner creation failures', async () => {
```
stuartp44, line 950:
```suggestion
    it('non-retryable partial runner creation failures are not retried', async () => {
```

edersonbrilhante: Let me add an issue for that. I am trying to merge this PR the whole day XD

PR #5312, opened 2026-08-27
refactor(ec2): decouple creation result from control plane

edersonbrilhante:
## Description

Separates the result returned by shared EC2 runner creation from the control-plane `CreateRunnerResult` contract.

`createEc2Runner()` and `Ec2RunnerResourceOperations.create()` now return `Ec2RunnerCreateResult`, containing created instance IDs, the failed-instance count, and bounded AWS failure identifiers.

The control-plane adapter converts that result into its existing `CreateRunnerResult` and applies retryable or non-retryable classification using `scaleErrors`. The core contract and existing pool and scale-up behavior remain unchanged.

## Test Plan

- Compute-provider suite: 14 files, 295 tests passed.
- Control-plane suite: 14 files, 345 tests passed.
- Control-plane TypeScript type-check passed.
- ESLint and Prettier passed for changed EC2 TypeScript files.
- `git diff --check` and pre-commit checks passed.

## Related Issues

Depends on #5301.

### DIFF
--- a/lambdas/libs/compute-providers/aws/ec2/src/control-plane/create-result.test.ts
+++ b/lambdas/libs/compute-providers/aws/ec2/src/control-plane/create-result.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest';
+
+import type { Ec2RunnerFailureCode } from '../runner-create-result';
+import { toControlPlaneCreateRunnerResult } from './create-result';
+
+function result(failureCodes: Ec2RunnerFailureCode[] = []) {
+  return {
+    instances: [],
+    failedInstanceCount: 2,
+    failureCodes,
+  };
+}
+
+describe('control-plane EC2 create result', () => {
+  it.each([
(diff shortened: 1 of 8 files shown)


Brend-Smits (dismissed):
Approving — clean refactor, and the classification logic all carried over faithfully (retryable sets, `$fault`/HTTP checks, the fleet-error path, and the spot→on-demand failover). Bounding the identifiers is a nice touch.

One side note, purely observability: the failure warns now log `failureCodes` instead of the error. Since `createEc2Runner` swallows AWS errors and returns a result (only `AbortError` rethrows), the raw error never reaches the outer `logger.error` either. Recognised AWS errors still surface as `aws-name:`/`aws-code:`, but anything that yields no codes — a non-AWS error, or a name/code filtered out by `SAFE_FAILURE_IDENTIFIER` — just logs `failureCodes: []` with nothing to go on. Might be worth keeping `error.message` alongside the codes so the odd unexpected failure stays debuggable.

Tiny thing: a CreateFleet failure logs "Create fleet request failed." twice now.

stuartp44 (approved):

PR #5314, opened 2026-08-28
fix(runners): support instance requirements in EC2 Fleet overrides

### LINKED ISSUE #5315: Dynamic instance requirements produce invalid EC2 Fleet overrides
## Problem

Dynamic EC2 labels can populate `InstanceRequirements` so a workflow can request a CPU and memory shape while EC2 Fleet chooses among multiple eligible instance families.

`generateFleetOverrides` currently iterates the statically configured instance types and adds an `InstanceType` to every override before spreading the dynamic override configuration. When dynamic labels provide `InstanceRequirements`, each resulting Fleet override contains both fields.

AWS requires `InstanceType` and `InstanceRequirements` to be mutually exclusive within an override, so `CreateFleet` rejects the request. The labels parse successfully, but no runner launches.

## Expected behavior

- Fixed instance selection should continue to produce one override per subnet and configured instance type.
- Attribute-based selection should produce one `InstanceRequirements`-only override per subnet, allowing Fleet to choose an eligible type.
- A request that explicitly contains both selection modes should fail before calling AWS.

## Reproduction

Enable dynamic labels and submit a job with valid vCPU and memory requirement labels, for example:

```yaml
runs-on:
  - self-hosted
  - <base-runner-labels>
  - ghr-ec2-vcpu-count-min:4
  - ghr-ec2-vcpu-count-max:4
  - ghr-ec2-memory-mib-min:8192
  - ghr-ec2-memory-mib-max:8192
```

The generated Fleet override combines a static `InstanceType` with the parsed `InstanceRequirements`, resulting in an invalid request.

## Additional validation

A requirements-only Spot Fleet request using the same launch-template boundary was accepted by AWS and selected an eligible instance family, confirming that omitting the static `InstanceType` is the required request shape.

slavab89:
## Description

Dynamic EC2 labels allow a workflow to request attribute-based instance selection using `InstanceRequirements`. This is useful when a workload needs a specific CPU and memory shape while allowing EC2 Fleet to choose among several eligible instance families for Spot capacity.

Today, `generateFleetOverrides` combines every configured static `InstanceType` with the dynamic `InstanceRequirements` object. AWS treats these fields as mutually exclusive within one Fleet override, so `CreateFleet` rejects the request. The dynamic labels parse successfully, but no runner can launch.

This change gives the two selection modes separate request paths:

- Fixed selection continues to produce one override per subnet and instance type.
- Attribute-based selection produces one `InstanceRequirements`-only override per subnet, allowing EC2 Fleet to choose from the eligible families.
- Configurations containing both `InstanceType` and `InstanceRequirements` are rejected before an AWS request is sent.

Existing fixed-instance behavior is unchanged.

## Test Plan

- Added regression coverage proving requirements-only overrides omit `InstanceType`.
- Added regression coverage proving conflicting fixed and attribute-based selection sends no `CreateFleet` request.
- Ran the focused EC2 runner suite: 71 tests passed.
- Ran the compute-provider suite: 10 files, 272 tests passed.
- Ran the compute-provider Prettier and ESLint checks.
- Ran `git diff --check`.

Validated the request against AWS using a temporary launch template without runner userdata or an IAM profile. A requirements-only Spot Fleet request for 4 vCPU / 8 GiB across several allowed compute families completed successfully and selected `c7i-flex.xlarge` with no Fleet errors. The instance and temporary launch template were deleted afterward.

The project typecheck currently reports existing Vitest 4 configuration errors in untouched configuration files. Those unrelated formatting and type changes are intentionally excluded from this PR.

## Related Issues

Fixes #5315. Related to #1400 and follows the dynamic EC2 label support introduced in #5003.

### DIFF
--- a/lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-up.test.ts
@@ -16,6 +16,7 @@ import type {
   CreateScaleUpRunnersInput,
   ScaleUpComputeProvider,
 } from './types';
+import { InvalidRunnerLabelsError } from '@aws-github-runner/compute-providers/core';
 import { defaultComputeProvider } from '@aws-github-runner/compute-providers/provider-types';
 import { getParameter } from '@aws-github-runner/aws-ssm-util';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
@@ -736,6 +737,41 @@ describe('scaleUp with GHES', () => {
           throw new Error(`Unexpected labels combination: ${labels.join(',')}`);
         }
       }
+    });
+
+    it('discards an invalid label group and continues processing valid groups', async () => {
+      const invalidMessage = {
+        ...TEST_DATA_SINGLE,
+        labels: ['self-hosted', 'ghr-provider-size:invalid'],
+        messageId: 'invalid-message',
+      };
+      const validMessage = {
+        ...TEST_DATA_SINGLE,
+        labels: ['self-hosted', 'ghr-provider-size:large'],
+        messageId: 'valid-message',
+      };
+      mockResolveLabelsForRunners.mockImplementation(async (labels) => {
+        if (labels.includes('ghr-provider-size:invalid')) {
+          throw new InvalidRunnerLabelsError('Invalid runner labels');
+        }
+        return {
+          runnerLabels: labels.filter((label) => label.startsWith('ghr-')),
+          state: testProviderState,
+        };
+      });
+
+      await expect(scaleUpModule.scaleUp([invalidMessage, validMessage])).resolves.toEqual([]);
+
+      expect(mockCreateRunners).toHaveBeenCalledTimes(1);
+      expect(mockCreateRunners).toHaveBeenCalledWith(
+        expect.objectContaining({
+          githubRunnerConfig: expect.objectContaining({
+            runnerLabels: 'base-label,ghr-provider-size:large',
+          }),
+        }),
+      );
+      expect(mockPublishRetryMessage).toHaveBeenCalledTimes(1);
+      expect(mockPublishRetryMessage).toHaveBeenCalledWith(validMessage);
     });
 
     it('preserves base RUNNER_LABELS for each group without mutation', async () => {
--- a/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
+++ b/lambdas/functions/control-plane/src/scale-runners/scale-up.ts
@@ -1,4 +1,5 @@
 import { addPersistentContextToChildLogger, createChildLogger } from '@aws-github-runner/aws-powertools-util';
+import { InvalidRunnerLabelsError } from '@aws-github-runner/compute-providers/core';
 import { resolveComputeProviderType } from '@aws-github-runner/compute-providers/provider-types';
 import { Octokit } from '@octokit/rest';
 import yn from 'yn';
@@ -20,6 +21,7 @@ import type {
   ActionRequestMessageSQS,
   CreateGitHubRunnerConfig,
   CreateRunnerResult,
+  RunnerLabelResolution,
 } from './types';
 
 const logger = createChildLogger('scale-up');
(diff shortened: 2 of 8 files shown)


lambdas/libs/compute-providers/aws/ec2/src/control-plane/runners.ts
```diff
@@ -220,11 +220,24 @@ function generateFleetOverrides(
 ): FleetLaunchTemplateOverridesRequest[] {
   const result: FleetLaunchTemplateOverridesRequest[] = [];
 
+  if (ec2OverrideConfig?.InstanceType && ec2OverrideConfig.InstanceRequirements) {
```
edersonbrilhante: Good catch. This works fine here for now. But if we need more validations like this one, this will get smelly.
To prevent a refactor in the future, lets create a validation function in `/lambdas/libs/compute-providers/aws/ec2/src/control-plane/dynamic-labels.ts`, there we can add the config validations, including this one

And in `lambdas/libs/compute-providers/aws/ec2/src/control-plane/scale-up.ts` we call the validation
```
    ec2OverrideConfig = parseEc2OverrideConfig(dynamicEC2Labels, defaultBlockDeviceName);
    if (ec2OverrideConfig) {
      logger.debug('EC2 override config parsed from labels', { ec2OverrideConfig });
    }
    <call validation function> with ec2OverrideConfig
```

lambdas/libs/compute-providers/aws/ec2/src/control-plane/runners.ts
```diff
@@ -220,11 +220,24 @@ function generateFleetOverrides(
...
-  const instanceTypesToUse = ec2OverrideConfig?.InstanceType ? [ec2OverrideConfig.InstanceType] : instancesTypes;
   const amiIdToUse = ec2OverrideConfig?.ImageId ?? amiId;
 
+  if (ec2OverrideConfig?.InstanceRequirements) {
+    return subnetsToUse.map((subnetId) => ({
+      SubnetId: subnetId,
+      ImageId: amiIdToUse,
+      ...ec2OverrideConfig,
+    }));
+  }
```
edersonbrilhante: Could we explicitly type the map callback return as FleetLaunchTemplateOverridesRequest, consistent with the item declaration below? The current code is type-safe through the function return annotation, but the explicit annotation makes the request contract clearer.

edersonbrilhante (changes requested):

lambdas/libs/compute-providers/aws/ec2/src/control-plane/runners.ts
slavab89: Addressed in 80fccdbf. I added validateEc2OverrideConfig in dynamic-labels.ts and call it from scale-up.ts immediately after parsing, before logging or runner creation. The scale-up tests now cover both accepted configurations and rejection of InstanceType plus InstanceRequirements before any EC2 runner is created.

lambdas/libs/compute-providers/aws/ec2/src/control-plane/runners.ts
slavab89: Addressed in 80fccdbf. The requirements-only map callback now explicitly returns FleetLaunchTemplateOverridesRequest.

edersonbrilhante:
@slavab89 Once this error is raised, it will pass along to this call `const runnerLabelResolution = await computeProvider.resolveLabelsForRunners(messageLabels);` in `lambdas/functions/control-plane/src/scale-runners/scale-up.ts:205`, then it will pass to the parent caller and it will be retried.

Shouldn't the message be discard?

slavab89:
@edersonbrilhante Good point. Addressed in 1c402763. Invalid runner-label configurations now raise a typed InvalidRunnerLabelsError. The control-plane catches only that error at the per-label-group boundary and skips the group without adding its message IDs to batchItemFailures, so SQS acknowledges and discards those messages. Other label-resolution errors still propagate and retry. I also added a mixed-batch regression test proving an invalid group is discarded while a valid group continues to launch.

edersonbrilhante (approved):
@slavab89 Good work.

PR #5324, opened 2026-09-01
fix(lambda): create log groups before functions

### LINKED ISSUE #5323: fix: prevent Lambda log group creation race during deployment
## Description

Terraform can hit a race condition while deploying the runner Lambdas and their CloudWatch log groups.

For example, in `modules/runners/scale-down.tf`, Terraform may create the scale-down Lambda and its scheduled EventBridge invocation before it creates the explicitly managed CloudWatch log group. If the schedule invokes the Lambda during that window, AWS automatically creates `/aws/lambda/<function-name>`. Terraform then attempts to create the same log group and fails because it already exists.

## Expected behavior

Every Terraform-managed Lambda log group must exist before its Lambda can be created or invoked, preventing AWS from creating the log group outside Terraform. All Lambda modules should use the same dependency pattern for consistent behavior and maintainability.

## Scope

- Standardize all Terraform-managed Lambda log groups in the repository, including these CloudWatch log-group/Lambda pairs:
  - `aws_cloudwatch_log_group.ami_housekeeper` in `modules/ami-housekeeper/main.tf`
  - `aws_cloudwatch_log_group.syncer` in `modules/runner-binaries-syncer/runner-binaries-syncer.tf`
  - `aws_cloudwatch_log_group.scale_down`
  - `aws_cloudwatch_log_group.scale_up`
  - `aws_cloudwatch_log_group.ssm_housekeeper`
  - `aws_cloudwatch_log_group.main`
  - `aws_cloudwatch_log_group.pool`
  - `aws_cloudwatch_log_group.webhook` in `modules/webhook/direct`
  - `aws_cloudwatch_log_group.dispatcher` in `modules/webhook/eventbridge`
  - `aws_cloudwatch_log_group.webhook` in `modules/webhook/eventbridge`
- Derive each log-group name without referencing the Lambda resource, avoiding the current implicit `log group -> Lambda` dependency.
- Make each Lambda explicitly depend on its own log group.
- Preserve the existing function and log-group names, retention settings, encryption, and tags.
- Add or update tests to verify the dependency ordering for all affected Lambdas.

## Acceptance criteria

- A fresh deployment cannot invoke a managed Lambda before Terraform has created its CloudWatch log group.
- Terraform no longer fails with a duplicate/already-existing log-group error when a Lambda is invoked during deployment.
- All Terraform-managed Lambda/log-group pairs follow the same dependency pattern.

edersonbrilhante:
## Description

- Define a shared name for each Terraform-managed Lambda and its CloudWatch log group.
- Make each Lambda explicitly depend on its own log group.
- Prevent AWS from auto-creating a Lambda log group before Terraform creates the managed resource.
- Standardize this dependency pattern across all affected Lambda modules while preserving existing names, retention, encryption, and tags.

## Test Plan

- `terraform fmt -recursive -check=true -write=false`
- `terraform validate`
- `tflint` for the root module and all affected modules
- `terraform -chdir=modules/runners test -test-directory=tests` (1 passed, 0 failed)
- Generated the Terraform plan graph and verified every affected Lambda depends on its CloudWatch log group.

## Related Issues

Fixes #5323

### DIFF
--- a/modules/ami-housekeeper/main.tf
+++ b/modules/ami-housekeeper/main.tf
@@ -1,6 +1,7 @@
 locals {
-  lambda_zip = var.lambda_zip == null ? "${path.module}/../../lambdas/functions/ami-housekeeper/ami-housekeeper.zip" : var.lambda_zip
-  role_path  = var.role_path == null ? "/${var.prefix}/" : var.role_path
+  lambda_name = "${var.prefix}-ami-housekeeper"
+  lambda_zip  = var.lambda_zip == null ? "${path.module}/../../lambdas/functions/ami-housekeeper/ami-housekeeper.zip" : var.lambda_zip
+  role_path   = var.role_path == null ? "/${var.prefix}/" : var.role_path
 }
 
 resource "aws_lambda_function" "ami_housekeeper" {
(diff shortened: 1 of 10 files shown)


Brend-Smits (approved):
LGTM :shipit:!

Tested it on an existing deployment and did not see changes, which is the important part here. Clean deployment still works well as well.

PR #5325, opened 2026-09-01
fix(runners): switch to on-demand instances to address Serverless gitaction verify failing with Error

dgignac-tii:
## Description

To debug Serverless gitaction verify failing with Error: The operation was canceled

re: https://turnitin.atlassian.net/browse/DEVOPS-28350


edersonbrilhante:
@dgignac-tii if you need to use on-demand pass the value to the variable. Spot is the default. We don't plan to switch the default value

PR #5335, opened 2026-09-02
fix(ec2): improve failure observability

edersonbrilhante:
## Description

Improves EC2 failure observability following #5312.

Caught EC2 failures now retain the error message alongside bounded `failureCodes`, so unexpected failures remain diagnosable even when no identifier is accepted. The redundant inner CreateFleet warning was also removed; the outer handler now emits that warning once after converting the swallowed request failure into its result.

## Test Plan

- Focused EC2 Vitest suite: 72 tests passed.
- ESLint passed for the changed TypeScript file.
- Source-only TypeScript check passed for `runners.ts`.
- Pre-commit checks and `git diff --check` passed.

The broader compute-provider TypeScript project check remains blocked by existing Vitest configuration/dependency-version errors unrelated to this change.

## Related Issues

Depends on #5312.


Brend-Smits (approved):
LGTM! Nice!

PR #5340, opened 2026-09-03
feat(multi-runner): consume resolved experimental config

edersonbrilhante:
## Description

- Derive runner-binary targets from the resolved multi-runner configuration.
- Break the binary-discovery configuration cycle by keeping discovery inputs independent from generated resources.
- Add effective-config coverage for the resolved binary artifact and preserve the v1-to-v2 translation boundary.

## Test Plan

- `terraform validate` in `modules/multi-runner`
- `terraform test` in `modules/multi-runner`
- `git diff --check` and rebase `range-diff` verification

## Related Issues

Depends on #5284.

PR #5342, opened 2026-09-03
refactor(multi-runner): consume effective configuration

edersonbrilhante:
## Description

- Consume `local.effective_config` across the multi-runner AMI housekeeper, queues, runners, SSM, webhook, and termination-watcher resources.
- Apply effective global and per-lane tags, artifacts, encryption, lifecycle, and provider settings consistently.
- Add effective-config fixtures covering global artifact precedence, SSM housekeeper settings, queue settings, and resolved runner-binary targets.

## Test Plan

- `terraform fmt -recursive`
- `terraform validate` in `modules/multi-runner`
- `terraform test` in `modules/multi-runner` — 6 passed, 0 failed
- Pre-commit formatting, validation/TFLint, and merge-conflict checks

## Related Issues

Depends on #5340.


modules/multi-runner/webhook.tf
```diff
@@ -1,44 +1,70 @@
+locals {
+  webhook_runner_config = {
+    for k, v in local.effective_config.multi_runner_config : k => v
+    if v.orchestration_provider.webhook != null
```
edersonbrilhante: **[P1] Guard the optional matcher configuration**\n\n`matcherConfig` is optional for the v2 entry shape, but this filter only checks that `webhook` is non-null; lines 13-18 then dereference `v.orchestration_provider.webhook.matcherConfig`. A valid webhook lane without matcher configuration will therefore fail during planning. Filter this map on `matcherConfig != null` (or provide an explicit default) and add a regression fixture for that supported shape.

Brend-Smits (approved):
LGTM :shipit:!

PR #5343, opened 2026-09-03
feat(compute-providers): add AWS EC2 provider modules

edersonbrilhante:
## Description

Adds the internal AWS EC2 compute-provider module and its trust-policy submodule. The provider owns EC2 runner instances, launch templates, networking, AMI selection, logging, bootstrap configuration, instance profiles, runner IAM policy fragments, and compute-provider validation.

The module exposes a typed provider contract for runner-config and orchestration consumers while keeping shared runner composition and orchestration outside the EC2 implementation.

### Compatibility decision

The experimental v2 EC2 provider intentionally remains separate from the stable `modules/runners` implementation. It owns its own Terraform resource definitions and copies of the runner bootstrap templates so the provider boundary can evolve without changing the stable module's public contract, resource graph, state addresses, or deployed behavior.

This creates temporary Terraform and template duplication, which increases review and maintenance effort and requires equivalent fixes to be evaluated in both implementations. Consolidation is deferred until stable v1 is deprecated and a migration can be designed and verified independently. ADR-002 records the decision and alternatives.

The ten copied template bodies match the stable templates. Eight are byte-identical; `cloudwatch_config.json` and `user-data.ps1` differ only by a final newline.

## Test Plan

- Repository Terraform formatting, validation/TFLint, and merge-conflict checks
- EC2 provider and trust-policy Terraform tests are included in the PR
- Compared all ten EC2 provider templates with `modules/runners/templates`; eight are byte-identical and two have only a final-newline difference
- `git diff --check`
- Local provider-plugin handshake prevented a complete test run on macOS arm64

## Related Issues

Depends on #5342.


Brend-Smits (approved):
Looks good to me, I checked if the hash from the duplicates are matching and as advertised 8 out of 10 match.
The ADR is also updated to reflect this duplication. Let's make sure we have a clear strategy to move away from v1 so we don't stick with the dupe for too long.

Good to go from my point of view.

PR #5347, opened 2026-09-03
chore(scale-set): docs and workflows changes

edersonbrilhante:
## Description

Adds the documentation, CI, release, and integration-test support needed to validate and publish the GitHub Actions scale-set service introduced by #5300.

This PR includes:

- Dependabot coverage and grouping for GitHub Actions, Lambda dependencies, the scale-set service container, MiniStack, and documentation dependencies.
- Lambda CI coverage for formatting, linting, tests, distribution builds, multi-architecture scale-set container builds, and an isolated container smoke test. The smoke test runs the image with a read-only filesystem, dropped capabilities, no-new-privileges, and no network access, then validates the controller health response.
- A MiniStack ECS/MockServer integration test that builds and pushes the scale-set image to the local registry, applies the `multi-runner-scale-set` example, validates the rendered SSM configuration and ECS task definition, exercises GitHub App and scale-set protocol requests, verifies runner launch and registration, changes the minimum capacity to zero, verifies scale-down and termination, and cleans up the deployment.
- Pinned MockServer expectations and MiniStack fixtures for installation-token exchange, runner registration, scale-set discovery and label updates, JIT configuration generation, message-session creation, polling, and cleanup.
- Release workflow support for publishing the multi-architecture scale-set service image to GHCR with an SBOM, build provenance, and registry attestation. Release notes now include the immutable image digest and attestation information alongside Lambda artifact attestations.
- Security documentation describing the scale-set image provenance and the `gh attestation verify` command for production image verification.

## Test Plan

- Lambda CI runs the repository format, lint, test, and build checks and builds the scale-set image for `linux/amd64` and `linux/arm64`.
- The container smoke test validates the hardened image independently of AWS or GitHub.
- The MiniStack workflow runs the scale-set ECS/MockServer smoke test with pinned MiniStack and MockServer versions.
- `git diff --check` and the changed-file scope were verified for this follow-up branch.

## Related Issues

- Follow-up to #5300.
- Exercises the Terraform scale-set deployment from #5299.
- Includes the scale-set service smoke-test work from #5375.


tests/ministack/run-scale-set-integration.sh
```diff
@@ -0,0 +1,600 @@
...
+replacement = os.environ["APP_KEY_BASE64"]
+github_config_url = os.environ["GITHUB_CONFIG_URL"]
+with open(source, encoding="utf-8") as source_file:
+    lines = source_file.readlines()
+with open(destination, "w", encoding="utf-8") as destination_file:
+    for line in lines:
+        if line.lstrip().startswith("key_base64"):
+            destination_file.write(f'  key_base64      = "{replacement}"\n')
+        elif line.lstrip().startswith("config_url"):
+            destination_file.write(f'  config_url         = "{github_config_url}"\n')
```
guicaulada, line 255: The fixture now uses `github.url`, so this branch no longer matches and leaves `https://mockserver:1080` unchanged. With the default local hostname or a custom `MINISTACK_GITHUB_MOCK_HOST`/port, the generated configuration differs from the expected endpoint and the assertion below fails. CI still passes because it uses the fixture’s hostname.

We should update both the matched and written field names to `url`:
```suggestion
        elif line.lstrip().startswith("url"):
            destination_file.write(f'  url                = "{github_config_url}"\n')
```

edersonbrilhante: I will skip this fix. I am rewriting the smoke tests right now. It was fine for a POC, but adding more complex test is ugly to read and add new cases. I am converting to a python lib, instead bash with python ingestion code.
Allowing to extends per computer provider and orchestration.

guicaulada: That makes sense, thanks for the context! I’m happy to leave this for the rewrite. We should make sure the new tests cover the default local endpoint and custom hosts/ports too.

PR #5350, opened 2026-09-03
feat(scale-set): wire orchestration through runner config

edersonbrilhante:
## Description

Wire scale-set orchestration through the provider-aware Terraform composition.

- Resolve global and per-lane scale-set settings into the canonical runner configuration.
- Route scale-set lanes through `runner-config` and the scale-set orchestration provider.
- Extend the EC2 provider contract with the scale-set runtime configuration, IAM fragments, and outputs consumed by the controller.
- Gate webhook resources by the selected orchestration provider.
- Add scale-set validation and Terraform coverage across the affected modules.

This PR provides the Terraform wiring used by the scale-set example and the later MiniStack integration PRs.

## Test Plan

- Terraform formatting, validation/TFLint, and merge-conflict checks passed through the repository hooks.
- Scale-set routing and configuration-resolution tests passed.
- Multi-runner and runner-config scale-set Terraform tests passed.
- Native provider tests remain subject to the local macOS arm64 plugin-handshake limitation.

## Related Issues

Depends on #5299.


modules/multi-runner/orchestration-provider.scale-set.tf
```diff
@@ -0,0 +1,62 @@
...
+            kms_key_arn = local.effective_config.ssm.kms_key_id
+          }
+          private_key = {
+            name        = local.primary_app_key_base64.name
+            arn         = local.primary_app_key_base64.arn
+            kms_key_arn = local.effective_config.ssm.kms_key_id
+          }
+          installation_id = {
+            name        = local.primary_app_installation_id.name
+            arn         = local.primary_app_installation_id.arn
```
guicaulada, line 19: Both `installation_id` and `installation_id_ssm` are optional. When neither is supplied, `local.primary_app_installation_id` is null, so these accesses fail planning with `Attempt to get attribute from null value`. I reproduced both errors with a scale-set-only configuration.

We should preserve an omitted installation reference through the scale-set provider schema, manifest, and IAM policy so the controller can discover the installation. If supplying an ID is intended to be mandatory instead, we should make that requirement explicit and validate it before these dereferences.

guicaulada (commented):
Nice work wiring this together! Looks good overall, but we should handle the case where the installation ID is omitted. I’ve left a comment inline.

modules/multi-runner/orchestration-provider.scale-set.tf
edersonbrilhante: It is required only in scaleset. How about a validation if scaleset is enabled?

PR #5359, opened 2026-09-04
refactor(storage): compose runner config storage providers

edersonbrilhante:
## Description

Compose the stable storage capabilities in one factory and add runner-config provider selection. Runner-config storage is selected independently from common GitHub App credentials, environment values are parsed once into immutable adapter configuration, and the old parallel provider factories are removed. The storage-provider package now declares its MIT license metadata.

## Test Plan

- Added provider-resolution and storage-composition tests.
- `git diff --check` passed.
- Runtime Yarn tests could not be run locally because the repository's pinned Yarn launcher is unavailable in this environment; CI should provide the full test result.

## Related Issues

- Related storage-backend work: #5266


lambdas/libs/storage-providers/storage-providers.ts
```diff
@@ -0,0 +1,50 @@
...
+    runnerGroupCache: createAwsSsmRunnerGroupCacheStore({ configPath, parameterStoreTags }),
+    consumer: createAwsSsmRunnerConfigConsumer({ SSM_TOKEN_PATH: tokenPath }, consumerConfig),
+    ...createCommonStorage(environment),
+  };
+}
+
+export function createCommonStorage(environment: Environment = process.env): CommonStorage {
+  return {
+    githubAppCredentials: createAwsSsmGitHubAppCredentialsStore(environment),
+  };
```
guicaulada, line 47: Passing the store into authentication bypasses the existing cache. The same authentication sequence now reads credentials five times instead of once, adding SSM requests and throttling pressure. Please cache the credentials per store. This change would restore reuse within each invocation:

```suggestion
  const store = createAwsSsmGitHubAppCredentialsStore(environment);
  let credentials: ReturnType<typeof store.get> | undefined;

  return {
    githubAppCredentials: {
      get: () => (credentials ??= store.get()),
    },
  };
```

guicaulada (commented):
LGTM! One minor comment about caching GitHub credentials... not sure if not caching them is intentional here, or if they are being cached elsewhere.

guicaulada (dismissed):

guicaulada (approved):

PR #5367, opened 2026-09-04
feat(examples): add multi-runner-v2 example

edersonbrilhante:
## Description

Adds the `examples/multi-runner-v2` example for the unified experimental v2 interface.

- Demonstrates global defaults and per-lane runner, webhook, and compute-provider settings.
- Uses the v2 module inputs without the legacy v1 compatibility arguments.
- Adds example documentation, Terraform verification, provider locks, and navigation entries.

## Test Plan

- `terraform init -backend=false -input=false` for `examples/multi-runner-v2`
- `terraform validate` for the example
- Terraform formatting, validation/TFLint, and merge-conflict checks

## Related Issues

Depends on #5368, which depends on #5346.


Brend-Smits (dismissed):
Looks good, I will approve it with the side note that:
- Migrating an existing deployment between v1 and v2 in place races on identically-named resources (runner IAM roles `${prefix}-<lane>-runner-<md5>`, SSM config params, CloudWatch log groups) because `modules/runners` and runner-config share the naming scheme. 
The first apply fails partway with AlreadyExists (a new resource is created before its identically-named twin finishes destroying); an immediate second apply always succeeds. It's not a data-loss risk and doesn't affect a clean/new deployment (which is what the example targets), but it's something we have to improve on before making the announcement about v2 being available. We need to make sure people can actually migrate an existing stack.

Brend-Smits (approved):

PR #5375, opened 2026-09-04
test(scale-set): run service container smoke test

edersonbrilhante:
## Description

Add the scale-set service-container smoke test and the ECS/MockServer scale-set lifecycle test on top of the scale-set Terraform wiring.

- Build the scale-set service image from `lambdas/services/scale-set/Dockerfile`, push it to a MiniStack ECR repository, and allow the MiniStack account to pull it.
- Run the controller through its real ECS container entrypoint with a read-only filesystem, dropped capabilities, and `no-new-privileges`.
- Configure the scale-set example with the MockServer GitHub App/Actions endpoints and the required SSM-backed GitHub App values.
- Verify the controller resolves the runner group and scale set, updates labels, creates a session, and reaches a converged state.
- Start at `minRunners = 1`, verify MiniStack creates and registers the EC2 runner, then register a new ECS task definition with `minRunners = 0` and verify the controller removes the mocked GitHub runner and terminates the EC2 instance.
- Keep the MockServer protocol fixture deterministic, including dynamic agent-name matching and runner/session cleanup.

The workflow uses MiniStack 1.5.12 and a pinned MockServer image. The scale-set lifecycle integration remains in this PR; PR #5416 provides only the reusable example-fixture support.

## Test Plan

- Full local MiniStack 1.5.12 lifecycle test passed: image build/ECR push, Terraform apply of 272 resources, ECS controller startup, scale-up runner registration, ECS redeployment with `minRunners = 0`, EC2 runner termination, MockServer session cleanup, and Terraform destroy.
- `sh -n tests/ministack/run-scale-set-integration.sh`
- `shellcheck -S warning tests/ministack/run-scale-set-integration.sh`
- `python3 -m json.tool mockserver/initializerJson.json`
- `terraform fmt -check examples/multi-runner-scale-set/main.tf examples/multi-runner-scale-set/variables.tf`
- `git diff --check`

## Related Issues

Depends on #5347. The Terraform wiring is provided by the scale-set stack beginning at #5350; the example is restored in #5405.


examples/multi-runner-scale-set/README.md
```diff
@@ -27,6 +27,7 @@ Configure the GitHub App variables before applying:
 terraform init
 terraform apply \
   -var='github_app={id="123456",key_base64="...",installation_id="123456789"}' \
+  -var='github={config_url="https://github.com",runner_owner="example",registration_level="organization"}' \
```
guicaulada: This `config_url` is passed to `enterprise_server.url`, and any non-null value sets `forceGhes = true`. Following the documented command therefore sends authentication requests to `https://github.com/api/v3/...` instead of `https://api.github.com/...`.

We could omit `config_url` for hosted GitHub so the example uses the public API defaults:

```suggestion
  -var='github={runner_owner="example",registration_level="organization"}' \
```

guicaulada (commented):
Nice work expanding the test coverage! Looks good overall, but we should adjust the GitHub.com example to avoid enabling GHES routing. I’ve added a suggestion inline.

PR #5377, opened 2026-09-04
feat(microvm): add microvm example

edersonbrilhante:
## Description

Adds the regional Lambda MicroVM foundation module, example, and image-build prerequisites.

- Creates the encrypted artifact storage, build and operator IAM roles, Lambda Network Connectors, connector security groups, runtime usage policy, and propagation dependencies.
- Adds the `microvm-foundation` example and documentation for provisioning the regional prerequisites required by MicroVM images.
- Adds the MicroVM Ubuntu image-builder configuration, scripts, service setup, and Packer assets.
- Updates the Terraform, MiniStack, Packer, and documentation workflows for the foundation and image-build paths.

## Test Plan

- Terraform formatting, initialization, and validation for the foundation module and examples.
- Provider and module tests included in the affected Terraform modules.
- Terraform commit hooks, TFLint, conflict checks, and `git diff --check`.
- No live AWS deployment was performed.

## Related Issues

- Stacked on #5413.
- Follow-up runtime provider implementation: #5255.


images/microvm-ubuntu/packer/scripts/microvm/image/ubuntu24.arm64.Dockerfile
```diff
@@ -0,0 +1,174 @@
...
+            --uid 1000 --gid 1000 runner; \
+    fi \
+    && install -d -m 0755 /opt/microvm /etc/services.d/cloudwatch-agent /var/log/microvm \
+    && install -m 0600 /dev/null /var/log/microvm/internal-services.log \
+    && install -m 0600 /dev/null /var/log/microvm/run.log \
+    && chown -R runner:runner /home/runner /opt/actions-runner
+
+COPY --chmod=0555 image-entrypoint.sh /opt/microvm/image-entrypoint.sh
+COPY --chmod=0555 start-services.sh /opt/microvm/start-services.sh
+COPY --chmod=0755 services/cloudwatch-agent.sh /etc/services.d/cloudwatch-agent/run
```
guicaulada, line 143: I think we may be missing the runner supervisor in this image. The hook in #5280 launches `/opt/microvm/entrypoint.sh run`, but I could only find `image-entrypoint.sh` here, which starts the hook server itself. The hook ZIP also seems to contain only `server.js` and `package.json`.

Unless that supervisor comes from somewhere else, I think the launch would fail with ENOENT. Maybe we should include the script that consumes the JIT document on stdin and signals `ready\n` on fd 3, then check that the assembled image can start a runner. Am I overlooking another step that supplies it?

PR #5384, opened 2026-09-07
fix(compute-providers): add provider diagnostics

edersonbrilhante:
## Description

Centralize compute-provider names and defaults, replace EC2 plugin literals with the canonical provider value, and add structured provider/capability diagnostics at pool, scale-up, scale-down, webhook, and registry boundaries.

Diagnostics avoid secrets, JIT configuration, signed webhook data, and full EC2 request payloads. Focused tests also cover provider resolution, registry consistency, capability selection, and sensitive logging behavior.

## Test Plan

- Passed focused compute-provider tests: 298/298
- Passed control-plane tests: 345/345
- Passed webhook tests: 83/83
- Passed lint for compute-providers, control-plane, and webhook
- Passed formatting validation for all changed TypeScript files
- Passed control-plane and webhook typechecks
- Compute-provider typecheck still reports pre-existing errors in `vitest.config.ts` and `vitest.base.config.ts`

## Related Issues

N/A

### DIFF
--- a/lambdas/libs/compute-providers/aws/ec2/logger.ts
+++ b/lambdas/libs/compute-providers/aws/ec2/logger.ts
@@ -1,0 +1,11 @@
+import { createChildLogger } from '@aws-github-runner/aws-powertools-util';
+
+import { computeProvider } from '../../provider-types';
+
+export function createEc2ComputeProviderLogger(module: string) {
+  const logger = createChildLogger(module);
+  logger.appendPersistentKeys({
+    computeProvider: computeProvider.ec2,
+  });
+  return logger;
+}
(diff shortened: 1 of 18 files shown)


lambdas/libs/compute-providers/aws/ec2/logger.ts
```diff
@@ -0,0 +1,11 @@
+import { createChildLogger } from '@aws-github-runner/aws-powertools-util';
+
+import { computeProvider } from '../../provider-types';
+
+export function createEc2ComputeProviderLogger(module: string) {
+  const logger = createChildLogger(module);
+  logger.appendPersistentKeys({
```
guicaulada, line 7: I think we should update the existing control-plane logger mock alongside this change. `functions/control-plane/src/lambda.test.ts` automocks `@aws-github-runner/aws-powertools-util`, which appears to leave `createChildLogger` returning undefined during provider imports. The current CI log points to the `appendPersistentKeys` call here as the reason that suite fails before its tests run.

Maybe we can have that mock return a logger with `appendPersistentKeys` and rerun the full control-plane suite as well as the new logger test.

PR #5395, opened 2026-09-08
chore(lambda): migrate Nx packages to 23.1.3

edersonbrilhante:
## Description

Migrates the Lambda workspace from Nx 22 to Nx 23.1.3.

- Aligns `nx` and the direct `@nx/*` packages on 23.1.3.
- Moves Vitest target inference to the `@nx/vitest` plugin, as required by Nx 23.
- Regenerates `lambdas/yarn.lock`.

Related: #5389

## Test Plan

- `yarn install --immutable`
- `yarn format-check`
- `yarn lint`
- `yarn test` (14 files, 350 tests passed)
- `yarn build`

## Related Issues

Related to #5389


lambdas/yarn.lock
```diff
@@ -10564,31 +10059,10 @@ __metadata:
...
-
-"socks@npm:^2.8.3":
-  version: 2.8.3
-  resolution: "socks@npm:2.8.3"
-  dependencies:
-    ip-address: "npm:^9.0.5"
-    smart-buffer: "npm:^4.2.0"
-  checksum: 10c0/d54a52bf9325165770b674a67241143a3d8b4e4c8884560c4e0e078aace2a728dffc7f70150660f51b85797c4e1a3b82f9b7aa25e0a0ceae1a243365da5c51a7
+"smol-toml@npm:1.6.1":
+  version: 1.6.1
```
guicaulada, line 10063: I think we should resolve this dependency before merging the Nx upgrade. Both Dependency Review and OSV are flagging the new `smol-toml@1.6.1` entry for [GHSA-7w5x-hrqm-74c2](https://github.com/advisories/GHSA-7w5x-hrqm-74c2), and OSV lists 1.7.1 as the fixed version.

Maybe we can update the dependency that brings it in, or use a compatible resolution, then regenerate the lockfile and rerun the migration checks. That should also clear the two failing dependency checks.

PR #5397, opened 2026-09-08
feat(scale-down): opt-in idle confirmation window before terminating not-busy runners

### LINKED ISSUE #5085: Scale-down can terminate a runner that picks up a job between busy check and termination

lochlanbennettodlum-taktile:
## Description

Closes #5085. Supersedes #5228.

This is @jensenbox's idle confirmation window from #5228, rebased onto the new provider layout and with the lint failures fixed. All credit for the approach and the production data is theirs, I just carried it forward since it stopped applying after the plugin refactor.

Quick recap of why: the GitHub `busy` flag isn't trustworthy enough to terminate on a single reading. People on #5085 have CloudTrail evidence of it reading `false` seconds after a job started, and in some cases minutes into a running job. The deregister-then-recheck idea in #5086 can't help because a GET after DELETE returns 404, which we already treat as "not busy".

So instead, opt in with `scale_down_idle_confirmation_seconds`. On a not-busy reading the runner gets tagged and termination is deferred. It's only terminated when a later evaluation still says not-busy and the window has elapsed. A busy reading in between clears the tag and starts over. Default is `0`, which is the current behaviour.

A couple of things I changed while porting:

- `ScaleDownComputeProvider` gains `markIdle` / `unmarkIdle`. They're required, per @edersonbrilhante's note below about DynamoDB becoming the state store. EC2 uses an instance tag for now, same as `ghr:orphan`, so no IAM changes. The template provider stubs both.
- Runners kept warm by `idle_config` also get their tag cleared. They never go through the removal path, so an old tag on one would otherwise let it be terminated on the first not-busy reading once the idle count drops. Added a test for this.
- The variable is threaded through the root module, `modules/runners` and `modules/multi-runner` (including the new translation layers). I left the experimental webhook orchestration module for a follow-up.

## Test Plan

- `yarn format-check`, `yarn lint`, `yarn build`, `yarn test` all green
- New tests for: window start, deferral, termination after the window, busy reset, kept-idle reset, disabled window, EC2 tag payloads
- `terraform validate`, `fmt -check` and `terraform test` pass for `modules/runners` and `modules/multi-runner`
- From the original PR: 14 hours on the author's fleet with a 300s window, zero mid-job terminations vs roughly one burst per hour before

## Related Issues

#5085, #5228, #5086, #5201

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -146,6 +146,18 @@ Cron expressions are parsed by [cron-parser](https://github.com/harrisiirak/cron
 
 For time zones please check [TZ database name column](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) for the supported values.
 
+### Idle confirmation window <!-- omit in toc -->
+
+Before terminating a runner, the scale-down lambda asks GitHub whether the runner is busy. That busy flag can be stale: it can read `false` for a runner that was assigned a job a few seconds earlier, and in rare cases for a runner that has been executing a job for several minutes. When that happens the lambda terminates an instance mid-job and the job fails with "The runner has received a shutdown signal".
+
+Set `scale_down_idle_confirmation_seconds` to require not-busy readings that span at least the given window before a runner is terminated. On the first not-busy reading the lambda tags the instance with `ghr:idle_detected_at` and defers termination. It terminates only when a later evaluation still reads not-busy and the window has elapsed. Any busy reading in between removes the tag and restarts the window. Use at least one scale-down schedule interval, for example `300` for the default five minute schedule, so that two consecutive evaluations must agree. The trade-off is that a genuinely idle runner lives one extra interval before it is removed.
+
+```hcl
+scale_down_idle_confirmation_seconds = 300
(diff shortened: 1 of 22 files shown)


edersonbrilhante:
@lochlanbennettodlum-taktile we plan to add dynamodb as state db. We can make the new functions required.

spcaipers-arm:
@edersonbrilhante can we merge all PRs related to this issue soon? We have problems on a daily basis, and I don't want to fork it just to patch it.

lochlanbennettodlum-taktile:
Done in 0758f90: `markIdle` / `unmarkIdle` are now required on `ScaleDownComputeProvider`, the optional-provider fallback and its test are gone, and the contract mocks and template are updated. Makes sense with DynamoDB coming as the state store. PR description updated to match.

edersonbrilhante (approved):
Good work

PR #5411, opened 2026-09-11
feat(dynamic-labels): add allow-list mode to the AWS dynamic labels policy

### LINKED ISSUE #5412: Dynamic labels policy has no allow-list mode, only deny-list
## Description

`aws_dynamic_labels_policy` (the guardrail for `ghr-ec2-*` dynamic labels) only supports a blocklist: `blocked_keys` rejects specific keys, `restricted_keys` limits the values a specific key can have, and any key not on either list is allowed.

This means: if GitHub adds a new EC2 override key later, or a key is simply left off the list, that key stays allowed until someone notices and blocks it. There is no way to set up a strict "nothing except these specific keys" policy without listing every single key to forbid — an open-ended, easy-to-miss list.

## Impact

An allow-list (deny by default) is the safer choice here, because dynamic labels let workflow authors set arbitrary EC2 launch settings (instance type, AMI, subnet, EBS volume size) — see the security warning already in `docs/configuration.md`. Right now, any new EC2 override key GitHub adds in the future is allowed by default for every deployment using this policy, until someone notices and adds it to `blocked_keys`.

## Proposed fix

Add `allowed_keys` to `AwsDynamicLabelsPolicy`: when set (not empty), only the keys listed there are allowed — everything else is rejected. `blocked_keys`/`restricted_keys` keep working exactly as before when `allowed_keys` is not set. Since `allowed_keys` and `blocked_keys` mean opposite things (allow-only vs. block-only), only one of them should be used at a time in the same policy — checked by Terraform at config time, and also checked at runtime (rejecting everything) if a policy with both somehow still reaches the code.

atsikham:
Fixes #5412

## Summary

`aws_dynamic_labels_policy` (the guardrail for `ghr-ec2-*` dynamic labels) only supports a blocklist: `blocked_keys` rejects specific keys, `restricted_keys` limits the values a specific key can have, and any key not on either list is allowed. This means a new EC2 override key GitHub adds later — or one that was simply left off the list — stays allowed until someone notices and blocks it. There was no way to set up a strict "nothing except these specific keys" policy without listing every single key to forbid.

## Changes

- Adds `allowed_keys?: string[]` to `AwsDynamicLabelsPolicy`. When set (not empty), any key **not** on that list is rejected — deny by default instead of allow by default. When it's unset or empty, nothing changes from today.
- `blocked_keys` and `restricted_keys` keep working exactly as before, on their own.
- `allowed_keys` and `blocked_keys` mean opposite things, so only one should be used per policy at a time:
  - Terraform checks this and fails `plan`/`apply` if both are set — checked everywhere this policy can be configured: the root module's `aws_dynamic_labels_policy`, `modules/webhook`'s `runner_matcher_config` (used when configuring the webhook module directly), and both the old and new shapes of `modules/multi-runner`'s `multi_runner_config`.
  - If a policy with both somehow still reaches the code some other way (for example, a hand-edited SSM parameter), it rejects every dynamic label instead of guessing which list to use.
- Docs (`docs/configuration.md`) updated with an example for the allow-list next to the existing blocklist example, the new check order, and a note on why the allow-list closes a real gap the blocklist has — a blocklist can't protect against a *future* key that doesn't exist yet; an allow-list can.

## Test plan

- [x] `yarn test` — new cases in `dynamic-labels-policy.test.ts`: unset/empty `allowed_keys` changes nothing, a key outside a non-empty `allowed_keys` is rejected, `allowed_keys` works together with `restricted_keys`, and the reject-everything behavior when both `allowed_keys`/`blocked_keys` are set. Full `compute-providers` (308 tests) and `webhook` (83 tests) suites pass.
- [x] `yarn lint` / `yarn format` / `tsc --noEmit`
- [x] `terraform fmt -check` / `terraform validate` on the root module, `modules/webhook` (standalone), and `modules/multi-runner` (standalone)


modules/webhook/variables.tf
edersonbrilhante: I think we can replace any for the proper nested structure

edersonbrilhante:
@atsikham The project requires signed your commits. Please sign so we can merge :)

modules/webhook/variables.tf
atsikham: Fixed

modules/multi-runner/variables.tf
```diff
@@ -679,6 +680,19 @@ variable "multi_runner_config" {
...
+      for config in var.multi_runner_config : !(
+        try(length(config.matcherConfig.awsDynamicLabelsPolicy.allowed_keys), 0) > 0 &&
+        try(length(config.matcherConfig.awsDynamicLabelsPolicy.blocked_keys), 0) > 0
+        ) && !(
+        try(length(config.orchestration_provider.webhook.matcherConfig.awsDynamicLabelsPolicy.allowed_keys), 0) > 0 &&
+        try(length(config.orchestration_provider.webhook.matcherConfig.awsDynamicLabelsPolicy.blocked_keys), 0) > 0
+      )
+    ])
+    error_message = "multi_runner_config: allowed_keys and blocked_keys cannot both be set in awsDynamicLabelsPolicy."
+  }
```
edersonbrilhante, line 695: Can you move this to validations.tf file?

atsikham: Moved into validations.tf's common_validation_errors, matching the existing pattern in that file.

edersonbrilhante (approved):

PR #5413, opened 2026-09-11
feat(microvm): add Terraform provider

edersonbrilhante:
## Description

Adds the AWS Lambda MicroVM compute-provider foundation and wires it into the Terraform runner configuration contracts.

- Adds the MicroVM compute-provider module, including control-plane resources, IAM policies, logging, runner configuration, outputs, validation, and the trust-policy submodule.
- Adds provider-specific Terraform tests and documentation.
- Adds global and per-runner MicroVM configuration and output resolution alongside the existing EC2 provider.
- Keeps the Terraform provider boundary independent from the runtime Lambda implementation, which is delivered by the follow-up provider PR.

This work is intentionally split across four PRs for reviewability. The complete end-to-end smoke coverage is implemented in #5280 and exercises both EC2 and MicroVM against:

- A webhook event without a dynamic label.
- A webhook event with a dynamic label.
- Pool scale-up.
- Scale-down.

## Test Plan

- Terraform tests for the MicroVM provider, trust policy, runner configuration, and multi-runner configuration.
- Terraform formatting, initialization, validation, and TFLint checks for affected modules.
- `git diff --check`
- No complete real-environment MicroVM deployment was performed in this PR because the end-to-end test depends on the complete stacked implementation:
  1. Apply the MicroVM foundation module from #5377.
  2. Build a real MicroVM image with the Packer image configuration.
  3. Build the lifecycle-hook server from #5280 and include it in the MicroVM image. The image requires this internal server to receive and process the MicroVM hooks.
  4. Deploy the updated Terraform code and Lambda artifacts.
  5. Run the unified Python MiniStack smoke test from #5280.

## Related Issues

- First Terraform layer in the four-PR MicroVM stack.
- Foundation module and image prerequisites: #5377.
- Runtime provider implementation: #5255.
- Lifecycle-hook server, Lambda packaging, and EC2 plus MicroVM smoke test: #5280.


modules/compute-providers/aws/microvm/control-plane.tf
```diff
@@ -0,0 +1,116 @@
...
+    MICROVM_EXECUTION_ROLE_ARN         = var.runner.iam.role.arn
+    MICROVM_IMAGE_ARN                  = var.config.image_arn
+    MICROVM_IMAGE_VERSION              = var.config.image_version == null ? "" : var.config.image_version
+    MICROVM_INGRESS_NETWORK_CONNECTORS = length(var.config.ingress_network_connectors) == 0 ? "" : jsonencode(var.config.ingress_network_connectors)
+    MICROVM_LOG_GROUP                  = aws_cloudwatch_log_group.runtime.name
+    MICROVM_METADATA_SSM_PATH          = local.microvm_metadata_ssm_path
+  })
+
+  scale_up_environment_variables   = local.microvm_environment_variables
+  scale_down_environment_variables = local.microvm_environment_variables
```
guicaulada: I think we should include the token path in the scale-down environment before landing this layer. On this head, `terraform test` in `modules/runner-config` fails in `routes_lambda_microvm_provider` because `SSM_TOKEN_PATH` is absent from the scale-down Lambda. The runtime provider in #5255 also requires it in `loadMicrovmProviderConfig`, so this would prevent scale-down from loading its configuration. I see the shared scale-down wiring added later in #5280; maybe we should move that small piece here so this Terraform layer and its test can stand on their own.

PR #5426, opened 2026-09-12
fix(github-runner): record the rate-limit metric even when the call fails

### LINKED ISSUE #5425: GitHubAppRateLimitRemaining metric is not recorded for the exact call that gets rate-limited
## Description

`isJobQueued` and `createJitConfig` (in `github-runner.ts`) both call `metricGitHubAppRateLimit(response.headers, appIndex)` right after a successful GitHub API call, using the rate-limit headers on that response. If the call itself fails because the app is rate-limited, that line is never reached, since it comes after the `await` that just threw.

## Impact

The `GitHubAppRateLimitRemaining` CloudWatch metric (enabled via `ENABLE_METRIC_GITHUB_APP_RATE_LIMIT`) never gets a data point for the exact call that hit the rate limit — the one moment the metric exists to catch. A dashboard built on this metric can show budget trending down over time but will miss the specific blocked calls, making it harder to correlate "which call got blocked and when" with what else was happening.

## Proposed fix

Catch the error in both places, pull the same rate-limit headers off the failed response if present, and record the metric before re-raising or moving on — the headers are already there on a 403/429 response, they're just not being read.

atsikham:
Fixes #5425

## Summary

`isJobQueued` and `createJitConfig` only called `metricGitHubAppRateLimit` after a successful response, reading the rate-limit headers off that response. When the call itself fails because the app is rate-limited, that line is never reached, so the exact call that hit zero remaining never shows up in the `GitHubAppRateLimitRemaining` metric.

## Changes

- Both now catch the error, pull the same rate-limit headers off the failed response if present, and still record the metric — before re-raising in `isJobQueued`, or before moving on to the next runner in `createJitConfig`.

## Test plan

- [x] `yarn test` — new `github-runner.test.ts`: `isJobQueued` records the metric on a successful call (regression guard), records it using the error's response headers when the call is rate-limited, and skips it cleanly when the error has no response headers at all; `createJitConfig` records the metric using the error's response headers when JIT config generation is rate-limited. Full control-plane suite passes (360 tests).
- [x] `tsc --noEmit`


Brend-Smits (approved):
Thank you for this pull request! I've tested this change and it works.

PR #5427, opened 2026-09-12
feat(runners): add network_interfaces for advanced EC2 launch template configuration

atsikham:
## Summary

Adds a new `network_interfaces` variable that exposes the full set of network interface fields supported by the `aws_launch_template` resource (interface type, network card index, IPv4/IPv6 prefix counts, connection tracking, ENA options, and more). Today only `associate_public_ipv4_address` is configurable, which covers a single simple interface and nothing else.

The new variable is wired through every configuration layer the same way `cpu_options` already is: root module, `modules/runners`, `modules/runner-config`, `modules/multi-runner` (both schemas), and the EC2 compute-provider module.

## Behavior

- `network_interfaces` defaults to an empty list. When empty, the launch template still derives a single interface from `associate_public_ipv4_address` exactly as before — no behavior change for existing configurations.
- When `network_interfaces` is set, it fully replaces that single derived interface; each list entry becomes its own `network_interfaces` block on the launch template, so callers set `associate_public_ip_address`/`security_groups` explicitly per entry.
- No Lambda or other runtime code is involved — this is a launch-template-level Terraform setting only, consistent with how `cpu_options` works.

## Test plan

- `terraform fmt -check -diff` and `terraform validate` pass in the root module, `modules/runners`, `modules/compute-providers/aws/ec2`, and `modules/multi-runner`.
- Added two `terraform test` cases to `modules/compute-providers/aws/ec2/tests/provider.tftest.hcl`: leaving `network_interfaces` unset still derives one interface from `associate_public_ipv4_address`; setting `network_interfaces` renders every configured interface with its fields on the launch template.
- Full existing test suites for the touched modules (`terraform test`) pass unchanged.


variables.tf
```diff
@@ -1048,6 +1048,44 @@ variable "associate_public_ipv4_address" {
   default     = false
 }
 
+variable "network_interfaces" {
```
edersonbrilhante, line 1051: The convention(in root folder) is to add prefix `runner_`

atsikham: Renamed to runner_network_interfaces.

edersonbrilhante (approved):
LGMT. Thanks for implementing in v1 and v2

PR #5434, opened 2026-09-16
ci(terraform): batch module and example verification

edersonbrilhante:
### Motivation

As new Terraform/OpenTofu features and modules were added, the validation matrix grew to almost 200 jobs. Each matrix combination starts a separate runner, increasing runner consumption and causing this and other pull requests to wait too long for available runners.

This change batches related validations while preserving coverage, reducing the workflow to 10 jobs.

### Changes

- Run module verification in one job per IaC version.
- Store all module paths, including the root module, in `MODULES`.
- Run Terraform and OpenTofu tests for `runners` and `multi-runner` using `TEST_MODULES`.
- Run example verification in one job per IaC version.
- Test Terraform 1.5.6/latest and OpenTofu 1.11/latest.
- Cache provider plugins separately for each job type.
- Create the plugin cache directory before running Terraform or OpenTofu.
- Use repository-root-relative module paths.
- Add valid ARN mocks to the runners pool test for OpenTofu compatibility.


Brend-Smits (commented):
Nice reduction in matrix size. 
One thing worth double-checking: `TEST_MODULES` only covers runners and multi-runner, but the previous matrix ran terraform test for 8 additional modules that still have .tftest.hcl files in the tree (compute-providers/aws/ec2 + trust-policy, orchestration-providers/webhook + its 3 submodules, runner-config + ssm-housekeeper). 

Was narrowing test execution to just those two intentional, or should `TEST_MODULES` include all modules that have a tests dir (which could also be auto-derived via `find modules -name '*.tftest.hcl'` instead of hardcoded)?

In general I'm not a big fan of having a hardcoded module list in the terraform.yml.
What about having some kind of 'discovery' job before that does something like:
```YAML
jobs:
  discover:
    name: Discover modules
    runs-on: ubuntu-latest
    outputs:
      modules: ${{ steps.discover.outputs.modules }}
    steps:
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
        with:
          persist-credentials: false
      - id: discover
        run: |
          {
            echo "modules<<EOF"
            { echo "."; find modules -type f -name '*.tf' -not -path '*/tests/*' -exec dirname {} \; | sort -u; }
            echo "EOF"
          } >> "$GITHUB_OUTPUT"

  verify_modules:
    needs: discover
    env:
      MODULES: ${{ needs.discover.outputs.modules }}
```

edersonbrilhante:
@Brend-Smits It is intentional I narrowed down for now. I will add later in another PR. I didn't want to fix small bugs in this PR and then create more conflicts in the scaleset and microvm PRs

Brend-Smits (approved):

PR #5445, opened 2026-09-18
fix: sign docs bot commits on fork via SSH key

rpaz-lemon:
## Description

The `update-docs` bot commits on `lemonatio`'s fork were unsigned. The workflow has a GraphQL `createCommitOnBranch` path that GitHub auto-verifies, but it's gated to `github.repository_owner == 'github-aws-runners'`, so it never runs here — the fork path falls back to a plain `terraform-docs/gh-actions` push instead.

This reuses the SSH-signing pattern already in production in `lemonatio/workflows`' `update-readme.yml` (and the `infra-*-module` repos that call it): assume the `github-actions-infra` role via OIDC, pull the SSH signing key from Secrets Manager (`lemoncash_prd_gh_signing_key`), configure `gpg.format ssh` + `commit.gpgsign`, and commit with `-S`.

Only the fork path (`github.repository_owner != 'github-aws-runners'`) changes — the upstream `github-aws-runners` path is untouched.

## Test Plan

- `actionlint` on the modified workflow: no new findings (pre-existing `SC2086` on line 167 is unrelated, in the `deploy-pages` job).
- Will confirm on this PR's own CI run that the `docs` job authenticates via OIDC, retrieves the signing key, and pushes a commit that shows as `Verified`.

## Related Issues

N/A

### DIFF
--- a/.github/workflows/sync-upstream.yml
+++ b/.github/workflows/sync-upstream.yml
@@ -1,0 +1,18 @@
+name: Sync fork with upstream
+
+on:
+  schedule:
+    - cron: "0 6 * * 1" # Every Monday at 06:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Sync fork
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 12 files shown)


rpaz-lemon:
Opened against the wrong repo by mistake (meant to target lemonatio's fork, not upstream). Closing immediately.

